PR #4, opened 2019-09-20
numba jit does not seem to be used...

davidtrem:
numba jit does not seem to be used...


seb5g:
Indeed within pymodaq, it is however used by one of the plugins. I've added numba dependency within pymodaq_plugins repo and removed it from here

PR #76, opened 2022-04-28
Add --input-file optional command-line argument to h5browser.py 

rgeneaux:
Hi,
I wanted to associate the h5browser to .h5 files so that I can open them directly when double clicking on them.
Here's a suggestion to make that work, h5browser.py now accepts an optional command-line argument, either one of:
```
python h5browser.py -i filename.h5
python h5browser.py --input-file=filename.h5
```

Then we can create a simple batch file (here on windows 10):
_H5Opener.bat_
```
@ECHO OFF
"C:\Miniconda\envs\myenv\python.exe" "your\path\to\pymodaq\h5browser.py" -i %1
```

Then do the "Right click/Open with..." on an .h5 file, and set H5Opener as the default program to read .h5 and .hdf5 files!

Possible future improvements:
- modify the h5browser.exe script too, so that one can run "h5browser -i file.h5"
- provide a simple utility that associates h5 files to h5browser automatically. and perhaps while we're at it, set a nice icon to h5 files too :-)

Cheers

### DIFF
--- a/docs/src/usage/modules/Utility_Modules.rst
+++ b/docs/src/usage/modules/Utility_Modules.rst
@@ -305,6 +305,38 @@ data viewer (type :ref:`NDviewer` that can display data dimensionality up to 4)
 
 .. :download:`png <h5browser.png>`
 
+Associating H5Browser with .h5 files
++++++++++++++++++++++++++++++++++++++
+
+By default, the H5Browser always asks the user to select a file. One can instead open a specified .h5 file directly,
+using the --input optional command line argument as follows:
+
+``python h5browser.py --input my_h5_file.h5``.
+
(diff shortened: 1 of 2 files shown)


seb5g:
Hello,
sorry I just saw your PR. I think it is nice but after trying it and reading about getopts I saw the package argparse seems to be more complete and provide command line help functionnalities. 

Do you think you could use this package, and make 
h5browser --help
display the available commands?

seb5g:
Hello, did you see the reply?

rgeneaux:
Done! argparse is nice indeed.
Added the doc to create the .bat script too.

seb5g (approved):

PR #79, opened 2022-07-26
Use Logger extension for saving the acquisition with PID extension (in progress)

quantumm:
As I wanted to use the Logger extension in combination with the PID extension I faced a strange behavior: the amount of ROI values saved (defined in a 1Dviewer in my case) would not be the same as the number of spectra saved. So I digged quite a lot into pymodaq internals to understand this behavior. Finally I understood that the ROI values are calculated by the viewer, and that using a refresh time of the viewer higher than the actual acquisition rate would provoke that the ROI values are not calculated and saved for each acquisition. And this provokes a discrepency between the spectra saved and the ROI values, which is for sure not an expected behavior. An easy solution is just to put the refresh time at 0 and let the visualization running, but this is not fully satisfactory.

For now this PR does NOT solve this problem, but I hope it will come soon. As soon is the time for vacations I wanted to do a first stop and maybe have a first partial review ;)

In the end it is mainly docstrings that I put along my way, I followed the Numpy convention: [https://numpydoc.readthedocs.io/en/latest/format.html#documenting-class-instances](https://numpydoc.readthedocs.io/en/latest/format.html#documenting-class-instances). Hope it will help the reading of the code. Maybe I obsessed a bit with documentation but it is like that ;)

A modification in the Modules manager: asking "Probe detectors’ data" would freeze a running acquisition. Now we separate two cases: if the detector is not grabbing we ask for a grab, otherwise it is not needed and then we do not disturb the running acquisition.

There are also a few bug fixes in the related modules.

There is still some work to do:

* the ROI values saving should not depend on the refresh time of the viewer (explained above)
* include actuators in the logger extension (and not only detectors) so that we can also save their position.
* stopping the logger extension will stop the acquisition. This is not an expected behavior.
* how to save other parameters? For exemple in my case I would like to save the actuator position, but also a phase that is the result of a calculation from the ROI values and other parameters.

Enjoy! :D

### DIFF
--- a/src/pymodaq/daq_logger.py
+++ b/src/pymodaq/daq_logger.py
@@ -420,18 +420,25 @@ class DAQ_Logging(QObject):
 
         Parameters
         ----------
-
-        acquisition_from_viewer : OrderedDict
+        acquisition_from_viewer : OrderedDict{
+                    Ndatas : int
+                    acq_time_s : float
+                    name : str
+                    data0D : OrderedDict{
+                                <channel name> : DataToExport
+                                …
+                                }
+                        The dictionary data0D can contain data from 0D detectors, but also some measurements (e.g. an
+                        ROI integration).
+                    data1D : OrderedDict
+                        Same as data0D but for 1D data.
+                    data2D : OrderedDict
+                        Same for 2D data.
+                    dataND : OrderedDict
+                        Same for ND data.
+                    }
             Dictionary that contains data, metadata and measurements corresponding to an acquisition. Data of different
-            dimensions are classified in different entries of the dictionary. For example:
-
-            acquisition_from_viewer["name"] is the name of the detector
-
-            acquisition_from_viewer["data0D"] would contain the lineouts from defined ROIs
-
-            acquisition_from_viewer["data1D"] would contain the spectrum from a spectrometer
-
-            …
+            dimensions are classified in different keys of the dictionary.
 
         """
         try:
--- a/src/pymodaq/daq_logger.py
+++ b/src/pymodaq/daq_logger.py
@@ -63,8 +63,8 @@ class DAQ_Logger(CustomApp):
         self.wait_time = 1000
 
         self.logger_thread = None
-        self.logger = None  # should be a reference either to self.h5saver or self.dblogger depending the choice of the user
-
+        self.logger = None  # should be a reference either to self.h5saver or self.dblogger depending on the choice of
+        # the user
         self.setup_ui()
 
     def setup_actions(self):
@@ -122,11 +122,11 @@ class DAQ_Logger(CustomApp):
 
     def connect_things(self):
         self.status_signal[str].connect(self.dashboard.add_status)
-        self._actions['quit'].triggered.connect(self.quit_fun)
-        self._actions['start'].triggered.connect(self.start_logging)
-        self._actions['stop'].triggered.connect(self.stop_logging)
-        self._actions['grab_all'].triggered.connect(self.start_all)
-        self._actions['infos'].triggered.connect(self.dashboard.show_log)
+        self._actions['quit'].connect_to(self.quit_fun)
+        self._actions['start'].connect_to(self.start_logging)
+        self._actions['stop'].connect_to(self.stop_logging)
+        self._actions['grab_all'].connect_to(self.start_all)
+        self._actions['infos'].connect_to(self.dashboard.show_log)
 
     def setup_menu(self):
         """
--- a/src/pymodaq/daq_logger.py
+++ b/src/pymodaq/daq_logger.py
@@ -249,8 +249,12 @@ class DAQ_Logger(CustomApp):
         self.settings.child('log_type').setValue(log_type)
 
     def start_logging(self):
-        """
-            Start a logging.
+        """Start saving the detectors data and measurements.
+
+        This method is called from the Logger extension module UI main menu: Play.
+
+        Instantiate a DAQ_Logging object, put it into a thread, and call its own start_logging method.
+
         """
         self.status_widget.setText('Starting logging')
 
@@ -413,11 +417,13 @@ class DAQ_Logging(QObject):
             logger.exception(str(e))
 
     def connect_detectors(self, status=True):
-        """
-        Connect detectors to DAQ_Logging do_save_continuous method
+        """Connect detectors grab_done_signals to DAQ_Logging.do_save_continuous method.
+
         Parameters
         ----------
-        status: (bool) If True make the connection else disconnect
+        status : bool
+            If True make the connection else disconnect.
+
         """
         self.modules_manager.connect_detectors(connect=status, slot=self.do_save_continuous)
 
@@ -440,6 +446,11 @@ class DAQ_Logging(QObject):
         self.data_logger.stop_logger()
 
     def start_logging(self):
+        """Start saving the detectors data and measurements.
+
+        This method is called by the DAQ_Logger.start_logging method.
+
+        """
         try:
             self.connect_detectors()
             self.stop_logging_flag = False
--- a/src/pymodaq/daq_logger.py
+++ b/src/pymodaq/daq_logger.py
@@ -425,6 +425,7 @@ class DAQ_Logging(QObject):
                                         Ndatas : int
                                         acq_time_s : float
                                         name : str
+                                            Name of the detector.
                                         data0D : OrderedDict{
                                                     <channel name> : DataToExport
                                                     …
--- a/src/pymodaq/daq_logger.py
+++ b/src/pymodaq/daq_logger.py
@@ -408,7 +408,7 @@ class DAQ_Logging(QObject):
             self.stop_logging()
 
     @Slot(OrderedDict)
-    def do_save_continuous(self, acquisition_from_viewer):
+    def do_save_continuous(self, acquisition_from_viewer: OrderedDict) -> None:
         """Receive raw data, metadata and measurements from the detectors and call the data_logger to save it.
 
         Triggered by the grab_done_signals of the detectors (DAQ_Viewers modules). It is triggered each time a detector
--- a/src/pymodaq/daq_logger.py
+++ b/src/pymodaq/daq_logger.py
@@ -65,6 +65,8 @@ class DAQ_Logger(CustomApp):
         self.logger_thread = None
         self.logger = None  # should be a reference either to self.h5saver or self.dblogger depending the choice of the user
 
+        self.setup_ui()
+
     def setup_actions(self):
         '''
         subclass method from ActionManager
@@ -120,13 +122,11 @@ class DAQ_Logger(CustomApp):
 
     def connect_things(self):
         self.status_signal[str].connect(self.dashboard.add_status)
-        self._actions['quit'].connect(self.quit_fun)
-
-        self._actions['start'].connect(self.start_logging)
-        self._actions['stop'].connect(self.stop_logging)
-        self._actions['grab_all'].connect(self.start_all)
-
-        self._actions['infos'].connect(self.dashboard.show_log)
+        self._actions['quit'].triggered.connect(self.quit_fun)
+        self._actions['start'].triggered.connect(self.start_logging)
+        self._actions['stop'].triggered.connect(self.stop_logging)
+        self._actions['grab_all'].triggered.connect(self.start_all)
+        self._actions['infos'].triggered.connect(self.dashboard.show_log)
 
     def setup_menu(self):
         """
--- a/src/pymodaq/daq_logger.py
+++ b/src/pymodaq/daq_logger.py
@@ -407,17 +407,40 @@ class DAQ_Logging(QObject):
             self.stop_scan_flag = True
             self.stop_logging()
 
-    def do_save_continuous(self, datas):
-        """
+    def do_save_continuous(self, acquisition_from_viewer):
+        """Receive data, metadata and measurements from the detectors and call the data_logger to save it.
+
+        Triggered by the grab_done_signals of the detectors (DAQ_Viewers modules). It is triggered each time a detector
+        (selected in the configuration of the logger) has finished its acquisition.
+
+        The signals are connected by the DAQ_Logging.connect_detectors method.
+
+        The type of the data_logger depends on the choice of the user. The data can be saved in a h5 file or a database.
+        It will be a H5Logger if the user choose to save the data in a h5 file.
+
+        Parameters
+        ----------
+
+        acquisition_from_viewer : OrderedDict
+            Dictionary that contains data, metadata and measurements corresponding to an acquisition. Data of different
+            dimensions are classified in different entries of the dictionary. For example:
+
+            acquisition_from_viewer["name"] is the name of the detector
+
+            acquisition_from_viewer["data0D"] would contain the lineouts from defined ROIs
+
+            acquisition_from_viewer["data1D"] would contain the spectrum from a spectrometer
+
+            …
 
         """
         try:
-            self.data_logger.add_datas(datas)
+            self.data_logger.add_datas(acquisition_from_viewer)
         except Exception as e:
             logger.exception(str(e))
 
     def connect_detectors(self, status=True):
-        """Connect detectors grab_done_signals to DAQ_Logging.do_save_continuous method.
+        """Connect detectors (DAQ_Viewer modules) grab_done_signals to DAQ_Logging.do_save_continuous method.
 
         Parameters
         ----------
--- a/src/pymodaq/daq_logger.py
+++ b/src/pymodaq/daq_logger.py
@@ -407,8 +407,9 @@ class DAQ_Logging(QObject):
             self.stop_scan_flag = True
             self.stop_logging()
 
+    @Slot(OrderedDict)
     def do_save_continuous(self, acquisition_from_viewer):
-        """Receive data, metadata and measurements from the detectors and call the data_logger to save it.
+        """Receive raw data, metadata and measurements from the detectors and call the data_logger to save it.
 
         Triggered by the grab_done_signals of the detectors (DAQ_Viewers modules). It is triggered each time a detector
         (selected in the configuration of the logger) has finished its acquisition.
@@ -421,22 +422,22 @@ class DAQ_Logging(QObject):
         Parameters
         ----------
         acquisition_from_viewer : OrderedDict{
-                    Ndatas : int
-                    acq_time_s : float
-                    name : str
-                    data0D : OrderedDict{
-                                <channel name> : DataToExport
-                                …
-                                }
-                        The dictionary data0D can contain data from 0D detectors, but also some measurements (e.g. an
-                        ROI integration).
-                    data1D : OrderedDict
-                        Same as data0D but for 1D data.
-                    data2D : OrderedDict
-                        Same for 2D data.
-                    dataND : OrderedDict
-                        Same for ND data.
-                    }
+                                        Ndatas : int
+                                        acq_time_s : float
+                                        name : str
+                                        data0D : OrderedDict{
+                                                    <channel name> : DataToExport
+                                                    …
+                                                    }
+                                            The dictionary data0D can contain data from 0D detectors, but also some
+                                            measurements (e.g. an ROI integration).
+                                        data1D : OrderedDict
+                                            Same as data0D but for 1D data.
+                                        data2D : OrderedDict
+                                            Same for 2D data.
+                                        dataND : OrderedDict
+                                            Same for ND data.
+                                        }
             Dictionary that contains data, metadata and measurements corresponding to an acquisition. Data of different
             dimensions are classified in different keys of the dictionary.
 
--- a/src/pymodaq/daq_utils/daq_utils.py
+++ b/src/pymodaq/daq_utils/daq_utils.py
@@ -464,10 +464,10 @@ def get_data_dimension(data: np.ndarray, scan_type='scan1D', remove_scan_dimensi
 
     """
     if not isinstance(data, np.ndarray):
-        raise Exception("Input parameter should be an ndarray.")
+        raise TypeError(f"Input parameter should be an ndarray: {type(data)} received.")
 
     if data.ndim == 0:
-        raise Exception("The dimension of the array should be superior to zero.")
+        raise ValueError("The dimension of the array should be superior to zero.")
 
     dimension = data.ndim
 
--- a/src/pymodaq/daq_utils/daq_utils.py
+++ b/src/pymodaq/daq_utils/daq_utils.py
@@ -2,9 +2,6 @@ import os
 import sys
 from collections import OrderedDict
 from ctypes import CFUNCTYPE
-
-from pymodaq.daq_utils.config import get_set_config_path, get_set_preset_path, Config
-from pymodaq.daq_utils.messenger import deprecation_msg
 if 'win32' in sys.platform:
     from ctypes import WINFUNCTYPE
 import datetime
@@ -21,21 +18,22 @@ import pkgutil
 import traceback
 import warnings
 import numbers
-
-import numpy as np
-from qtpy import QtCore
-from qtpy.QtCore import QLocale
-from pymodaq.daq_utils.qvariant import QVariant
-
 python_version = f'{str(sys.version_info.major)}.{str(sys.version_info.minor)}'
 if version_mod.parse(python_version) >= version_mod.parse('3.8'):  # from version 3.8 this feature is included in the
     # standard lib
     from importlib import metadata
 else:
     import importlib_metadata as metadata  # pragma: no cover
+from typing import Tuple
 
+import numpy as np
+from qtpy import QtCore
+from qtpy.QtCore import QLocale
+
+from pymodaq.daq_utils.qvariant import QVariant
+from pymodaq.daq_utils.config import get_set_config_path, get_set_preset_path, Config
+from pymodaq.daq_utils.messenger import deprecation_msg
 from pymodaq.daq_utils.exceptions import DataSourceError
-
 
 plot_colors = [(255, 255, 255), (255, 0, 0), (0, 255, 0), (0, 0, 255), (14, 207, 189), (207, 14, 166), (207, 204, 14)]
 config = Config()
@@ -442,7 +440,7 @@ def uncapitalize(string, Nfirst=1):
     return string[:Nfirst].lower() + string[Nfirst:]
 
 
-def get_data_dimension(data: np.ndarray, scan_type='scan1D', remove_scan_dimension=False):
+def get_data_dimension(data: np.ndarray, scan_type='scan1D', remove_scan_dimension=False) -> Tuple[Tuple, str, str]:
     """Return the shape, dimension and size of the input data.
 
     Notice that the dimension returned does not necessarily correspond to the size of the input ndarray. If the shape of
--- a/src/pymodaq/daq_utils/daq_utils.py
+++ b/src/pymodaq/daq_utils/daq_utils.py
@@ -443,6 +443,25 @@ def uncapitalize(string, Nfirst=1):
 
 
 def get_data_dimension(arr, scan_type='scan1D', remove_scan_dimension=False):
+    """Get a numpy ndarray and return its shape, dimension and size.
+
+    Parameters
+    ----------
+    arr : numpy ndarray
+        Typically raw data from an acquisition (a spectrum, an image…).
+    scan_type : str
+        ??
+    remove_scan_dimension : bool
+        ??
+
+    Returns
+    -------
+    tuple of tuple, string, int
+
+    """
+    if not isinstance(arr, np.ndarray):
+        raise Exception
+
     dimension = len(arr.shape)
     if dimension == 1:
         if arr.size == 1:
@@ -456,6 +475,7 @@ def get_data_dimension(arr, scan_type='scan1D', remove_scan_dimension=False):
     else:
         if dimension > 2:
             dimension = 'N'
+
     return arr.shape, f'{dimension}D', arr.size
 
 
@@ -725,15 +745,20 @@ class DataFromPlugins(Data):
 
 class DataToExport(Data):
     def __init__(self, data=None, dim='', source='raw', **kwargs):
-        """
-        Utility class defining a data being exported from pymodaq's viewers, attributes can be accessed as dictionary keys
+        """Utility class defining a data being exported from pymodaq's viewers, attributes can be accessed as dictionary
+        keys.
+
         Parameters
         ----------
-        data: (ndarray or a scalar)
-        dim: (str) data dimensionality (either Data0D, Data1D, Data2D or DataND)
-        source: (str) either 'raw' for raw data or 'roi' for data extracted from a roi
+        data : ndarray or float or int or None
+        dim : str
+            Data dimensionality (either Data0D, Data1D, Data2D or DataND).
+        source : str
+            Either 'raw' for raw data or 'roi' for data extracted from a roi.
+
         """
         super().__init__(source=source, **kwargs)
+
         if data is None or isinstance(data, np.ndarray) or isinstance(data, float) or isinstance(data, int):
             self['data'] = data
         else:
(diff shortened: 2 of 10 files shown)


seb5g (changes requested):
see details below

src/pymodaq/daq_viewer/daq_viewer_main.py
```diff
@@ -1843,31 +1980,58 @@ def process_tcpip_cmds(self, status):
...
+        one could refresh the UI displaying slower than the actual speed of the acquisition. It actually reduces the
+        number of acquisition per second and thus gives a lower limit to the number of acquisition you can save per
+        second (see grab_data method).
+        !!!Not to be confused with DAQ_Viewer.wait_time!!!
+    grab_state : bool
+        True if the viewer is grabbing, False otherwise.
+    single_grab : bool
+        True if the last acquisition order from the user was a snap (single grab), False otherwise.
+    Naverage : int
+        ??
```
seb5g, line 2012: integer to track the total number of acquisition to be averaged before displaying. this averaging is done in the DAQ_Detector if the plugin has not the self.hardware_averaging set to True else it should be done inside the plugin. For instance, it could be smarter to do the averaging inside a camera plugin when one ofen use a buffer and where it is faster to do so than exporting (and copying) all data and then averaging

src/pymodaq/daq_viewer/daq_viewer_main.py
```diff
@@ -1843,31 +1980,58 @@ def process_tcpip_cmds(self, status):
...
+    ind_average : int
+        ??
```
seb5g, line 2014: keep track of the index of data within the averaging process (if local averaging)

src/pymodaq/daq_viewer/daq_viewer_main.py
```diff
@@ -1843,31 +1980,58 @@ def process_tcpip_cmds(self, status):
...
+    average_done : bool
+        ??
```
seb5g, line 2016: if True will stop the while loop used to average data locally

src/pymodaq/daq_viewer/daq_viewer_main.py
```diff
@@ -2122,17 +2288,19 @@ def single(self, Naverage=1, args_as_dict={}):
...
-            =============== =========== ==================
+        This method is called (indirectly) by DAQ_Viewer.grab_data.
+        Call the detector_plugin.grab_data method.
+
+        !!!Not to be confused with DAQ_Viewer.grab_data!!!
+
+        Parameters
+        ----------
+        Naverage : int
+            ??
```
seb5g, line 2300: integer indicating the number of data to average before displaying data

src/pymodaq/daq_viewer/daq_viewer_main.py
```diff
@@ -2122,17 +2288,19 @@ def single(self, Naverage=1, args_as_dict={}):
...
+        live : bool
+            ??
```
seb5g, line 2302: corresponds to a grab (with opposition to a snap)

src/pymodaq/dashboard.py
```diff
@@ -610,42 +647,7 @@ def set_file_preset(self, filename):
...
-                        QtWidgets.QApplication.processEvents()
-                        self.poll_init(actuators_modules[-1])
-                        QtWidgets.QApplication.processEvents()
-
-            except Exception as e:
-                logger.exception(str(e))
-
-            # ################################################################
-            # ##### sort plugins by IDs and within the same IDs by Master and Slave status
+            # Sort plugins by IDs and within the same IDs by Master and Slave status
```
seb5g, line 650: why removing this?

src/pymodaq/dashboard.py
```diff
@@ -785,8 +787,8 @@ def set_file_preset(self, filename):
                 self.load_layout_state(path)
 
             self.mainwindow.setWindowTitle(f'PyMoDAQ Dashboard: {self.title}')
-            if self.pid_module is not None:
-                self.pid_module.set_module_manager(detector_modules, actuators_modules)
+            # if self.pid_module is not None:
+            #     self.pid_module.set_module_manager(detector_modules, actuators_modules)
```
seb5g, line 791: why commenting these?

src/pymodaq/dashboard.py
```diff
@@ -1049,19 +1043,14 @@ def set_preset_mode(self, filename):
...
                 self.actuators_modules = actuators_modules
                 self.detector_modules = detector_modules
 
                 self.modules_manager = ModulesManager(self.detector_modules, self.actuators_modules)
-                #
-                if self.pid_module is not None:
-                    self.pid_module.ini_model_action.click()
-                #
-
-                #####################################################
```
seb5g, line 1064: why removing this?

src/pymodaq/dashboard.py
```diff
@@ -1108,11 +1097,44 @@ def set_preset_mode(self, filename):
...
+                for setp in model_class.setpoints_names:
+                    self.add_move(setp, None, 'PID', [], [], actuators_modules)
+                    actuators_modules[-1].controller = dict(
+                        curr_point=self.pid_module.curr_points_signal,
+                        setpoint=self.pid_module.setpoints_signal,
+                        emit_curr_points=self.pid_module.emit_curr_points_sig)
+                    actuators_modules[-1].ui.IniStage_pb.click()
+                    QtWidgets.QApplication.processEvents()
+                    self.poll_init(actuators_modules[-1])
+                    QtWidgets.QApplication.processEvents()
```
seb5g, line 1131: ok you've put it there, could you tell me why?

src/pymodaq/pid/pid_controller.py
```diff
@@ -49,8 +49,6 @@ class DAQ_PID(QObject):
 
         {'title': 'Main Settings:', 'name': 'main_settings', 'expanded': True, 'type': 'group', 'children': [
             {'title': 'Acquisition Timeout (ms):', 'name': 'timeout', 'type': 'int', 'value': 10000},
-            {'title': 'epsilon', 'name': 'epsilon', 'type': 'float', 'value': 0.01,
-             'tooltip': 'Precision at which move is considered as done'},
```
seb5g, line 53: why removing it?

src/pymodaq/pid/pid_controller.py
```diff
@@ -301,17 +299,19 @@ def stop_moves(self, overshoot):
...
+        Construct the UI depending on the number of setpoints defined in the model.
+        Call the ini_model function of the model.
+        """
         try:
-            if self.model_class is None:
-                self.set_model()
+            model_name = self.settings.child('models', 'model_class').value()
+            self.model_class = find_dict_in_list_from_key_val(self.models, 'name', model_name)['class'](self)
+            self.set_setpoints_buttons()
+            self.model_class.ini_model()
```
seb5g, line 314: if I removed it from there there should be a reason...well maybe I did crap, but that could cause some hidden bug...

src/pymodaq/pid/pid_controller.py
```diff
@@ -592,7 +596,7 @@ def start_PID(self, sync_detectors=True, sync_acts=False):
                     self.outputs = [pid.setpoint for pid in self.pids]
 
                 dt = time.perf_counter() - self.current_time
-                self.outputs_to_actuators = self.model_class.convert_output(self.outputs, dt, stab=True)
+                self.outputs_to_actuators = self.model_class.convert_output(self.outputs)
```
seb5g, line 599: You should leave it, why removing these, there are usefull!

seb5g: see the PIDModelBoiler example

seb5g (changes requested):

src/pymodaq/daq_logger.py
```diff
@@ -63,7 +63,9 @@ def __init__(self, dockarea=None, dashboard=None):
         self.wait_time = 1000
 
         self.logger_thread = None
-        self.logger = None  # should be a reference either to self.h5saver or self.dblogger depending the choice of the user
+        self.logger = None  # should be a reference either to self.h5saver or self.dblogger depending on the choice of
+        # the user
+        self.setup_ui()
```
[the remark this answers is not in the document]
quantumm: This is just a matter of respecting the lign length maximum of 120 characters.

src/pymodaq/daq_utils/h5modules.py
```diff
@@ -1897,18 +1923,25 @@ def add_datas(self, datas):
...
                     data_group = self.h5saver.get_node(det_group, utils.capitalize(data_type))
+
                 for ind_channel, channel in enumerate(datas[data_type]):
                     channel_group = self.h5saver.get_group_by_title(data_group, channel)
                     if channel_group is None:
                         channel_group = self.h5saver.add_CH_group(data_group, title=channel)
+                        # This condition should be added for the 0D case because H5Saver.add_data requires that the key
+                        # data_dict["data"] should be an ndarray, and not a float.
+                        if data_type == "data0D":
+                            datas[data_type][channel]["data"] = np.array(datas[data_type][channel]["data"])
```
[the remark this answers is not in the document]
quantumm: The last [commits](https://github.com/CEMES-CNRS/PyMoDAQ/pull/79/commits/addec23e326823ba48709a20dd5e26f048e0a561) should solve those issues.
Now the type of the DataToExport["data"] key is checked at the begining of the last for loop and changed to a ndarray of shape (1,) if it is a scalar.
At appending into the EARRAY, it is no more needed to distinguish the "0D" case.

Don’t you think it would be nice to restrict the DataToExport["data"] key to be a ndarray with dimension > 0? (I don’t know if it is usefull somewhere else in the code that it can accept float, int or None). This way the DataToExport objects would for sure be "clean" to be appended within pytables arrays, and one would not have to deal with the particular case of scalar data at this stage.

src/pymodaq/daq_utils/managers/modules_manager.py
```diff
@@ -3,13 +3,46 @@
 from qtpy import QtWidgets
 import time
 from pymodaq.daq_utils import daq_utils as utils
+from pymodaq.daq_viewer.daq_viewer_main import DAQ_Viewer
```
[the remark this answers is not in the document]
quantumm: It is actually used in the grab_one_detector method.

Could it be detrimental for performances?

src/pymodaq/daq_utils/managers/modules_manager.py
```diff
@@ -3,13 +3,46 @@
...
 
 logger = utils.set_logger(utils.get_module_name(__file__))
 
 
 class ModulesManager(QObject):
+    """
+
+    Attributes
+    ----------
+    detectors : list of DAQ_Viewer
```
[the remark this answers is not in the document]
quantumm: Done in 576279b27e0920f493800ad987c5cf16abcd7f1f.

src/pymodaq/daq_utils/managers/modules_manager.py
```diff
@@ -192,24 +225,43 @@ def parameter_tree_changed(self, param, changes):
...
+            detector_acquisition = detector.data_to_save_export
+            det_name = detector.detector_name
+
+            try:
+                # If the detector is not grabbing (detector_acquisition is None), we send a grab signal. But if it is
+                # grabbing, we do not need to send the grab signal, which would freeze the acquisition.
+                if detector_acquisition is None:
+                    detector.grab_done_signal.connect(self.det_done)
+                    detector_acquisition = self.grab_one_detector(detector)
+                    detector.grab_done_signal.disconnect(self.det_done)
```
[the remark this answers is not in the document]
quantumm: > this method will not work generally. It will only work when no grab has been done yet on the detector. However if you did a first grab, then stops, then look at this attribute, its value will not be none...

Yes I think that is the behavior we need.

If we are in a state of a dashboard with two detectors, with the first one acquiring and the second one has never acquired (buffer = None). If we launch a snap to every detectors, then we will freeze the acquisition of the first one, which is not what we want. The intention here is to send a grab signal only when needed, so only to detectors with empty buffer, which should not be of much consequence since it means they are just "off". That is why the grab_one_detector has been added.

I try to explained it better in the docstrings but changed nothing in the logic: 8149a13607dc38f31a814fbe92f5af2987c5bee6

I don’t think I understand your last question. Maybe there is a better way to get this information without grabbing?

src/pymodaq/daq_utils/plotting/data_viewers/viewer1D.py
```diff
@@ -10,6 +10,7 @@
 import pyqtgraph as pg
```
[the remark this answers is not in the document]
quantumm: Ok ;)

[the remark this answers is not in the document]
quantumm: That was an easy one :P

src/pymodaq/daq_viewer/daq_viewer_main.py
```diff
@@ -60,12 +60,76 @@
...
+        DAQ_Viewer.wait_time) will not directly affect the actual speed of the acquisition. If for example the
+        acquisition loop runs at 50 Hz, there is no point to display on the UI at such a speed. In that case the user
+        can specify a refresh time longer than 20 ms. Notice that the longer the refresh time, the better the
+        performances, since the displaying is quite costly in terms of resources. A very short refresh time can slow
+        down the program. For most applications, a refreshing time higher than 200 ms is enough. If the user wants to
+        optimize the performance of the program, he can untick the "Show data and process" parameter on the UI. If the
+        user chooses a refreshing time shorter than the delay between two acquisitions, then it just means that the
+        viewer will display all the acquisitions. Notice also that the ROI lineouts are calculated each time the
+        viewer is refreshed. It means that (up to now) the user cannot perform ROI lineouts calculation and saving
+        faster than the refresh time of the viewer. This is probably not an expected behavior.
```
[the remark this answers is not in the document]
quantumm: You are welcome! :) I needed to write it down :P

What I meant for the unexpected behavior is that for example if I want to save some ROI lineouts from a viewer with the Logger extension, I cannot turn off the visualization of the data. The behavior of the saving depends on the visualization on the UI. Maybe this can be easily corrected by modifying the action when the user uncheck "Show data and process".

src/pymodaq/daq_viewer/daq_viewer_main.py
```diff
@@ -60,12 +60,76 @@
...
+                                    Same as data0D but for 1D data.
+                                data2D : OrderedDict
+                                    Same for 2D data.
+                                dataND : OrderedDict
+                                    Same for ND data.
+                                }
+        Buffer to save the data and the measurements from the current acquisition of the detector.
+    wait_time : int
+        Time in milliseconds a message displayed in the UI status bar will stay before disappearing.
+        !!!Not to be confused with DAQ_Detector.wait_time!!!
```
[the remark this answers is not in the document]
quantumm: Yes maybe we can question if this status bar is still usefull…

src/pymodaq/daq_viewer/daq_viewer_main.py
```diff
@@ -926,23 +1046,26 @@ def get_data_from_viewer(self, datas):
...
 
-    def set_datas_to_viewers(self, datas, temp=False):
+    def set_datas_to_viewers(self, datas: List[utils.DataFromPlugins], temp=False):
+        """Transfer the metadata and the acquired raw data from the detector (DAQ_Detector) to the UI viewer.
+
+        Parameters
+        ----------
+        datas : list of DataFromPlugins
+            The objects in this list contain the acquired raw data and the associated metadata. This list can contain
+            several DataFromPlugins objects because the DAQ_Viewer can manage several detectors???
```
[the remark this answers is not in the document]
quantumm: I am confused with that. In what context will you need a DAQ_Viewer to manage several viewers? Is not that the role of the dashboard?

src/pymodaq/daq_viewer/daq_viewer_main.py
```diff
@@ -1039,7 +1162,16 @@ def process_data(self, datas, container):
...
+        """Transfert the acquired data from the detector (DAQ_Detector) to the UI viewer.
+
+        Triggered by DAQ_Detector.data_detector_sig.
+        Emit the grab_done_signal.
+
+        Parameters
+        ----------
+        datas : list of DataFromPlugins
+            The objects in this list contain the acquired raw data and the associated metadada. This list can contain
+            several DataFromPlugins objects because the DAQ_Viewer can manage several detectors???
```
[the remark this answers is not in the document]
quantumm: I guess it is the same question.

src/pymodaq/daq_viewer/daq_viewer_main.py
quantumm: I tried to improve the DAQ_Detector docstrings here: 5809744b1a7b07727435769b031bdd579baaca7f

but I think the software averaging should be tested because the behavior is quite strange. I did the following:
TEST of DAQ_Viewer with the 1D Mock_spectro:

* if I change Naverage nothing happens
* if then I click Show averaging nothing happens
* if I stop/restart I can see an effect of Show averaging but it is quite strange. I would expect to see the average value of my signal with 10 time less refreshing of the viewer. But instead we see with the same refreshing time the signal that is more and more average at each iteration, up to go back to the non-averaged signal, and it will loop like this.
* the "live averaging" button seems nice.

It is not the purpose of this PR so I let this for further work.

src/pymodaq/dashboard.py
quantumm: Yes it is because in the case you declare to use the PID extension in your preset, the actuators and detectors should be loaded before loading the pid extension.

quantumm: It corresponds to this commit 39d826e88588cd752700edfa96b47799377a9bdd.

src/pymodaq/pid/pid_controller.py
quantumm: I put them back!

src/pymodaq/pid/pid_controller.py
quantumm: Yes actually I do not know so i put it back as it was.

src/pymodaq/pid/pid_controller.py
quantumm: I put it back.

quantumm:
This is almost finished but there is still one or two remarks that I did not answered.
I will be on vacations for the next three weeks so i’ll come back to it with a fresh brain ;)

Ciao!

src/pymodaq/pid/pid_controller.py
quantumm: I put it back as it was.

src/pymodaq/daq_logger.py
```diff
@@ -403,21 +407,55 @@ def queue_command(self, command):
             self.stop_scan_flag = True
             self.stop_logging()
 
-    def do_save_continuous(self, datas):
-        """
+    @Slot(OrderedDict)
+    def do_save_continuous(self, acquisition_from_viewer):
```
Einenlum, line 411: Maybe a typehint could be added to `acquisition_from_viewer`

src/pymodaq/daq_utils/daq_utils.py
```diff
@@ -442,11 +442,37 @@ def uncapitalize(string, Nfirst=1):
...
+    remove_scan_dimension : bool
+        ??
+
+    Returns
+    -------
+    tuple of tuple, string, int
+
+    """
+    if not isinstance(data, np.ndarray):
+        raise Exception("Input parameter should be an ndarray.")
```
Einenlum, line 467: You can use a more specific `ValueError` here.

Einenlum: Maybe adding `f"Input parameter should be an ndarray: {type(np.ndarray)} received."` could help the user.

src/pymodaq/daq_utils/daq_utils.py
```diff
@@ -442,11 +442,37 @@ def uncapitalize(string, Nfirst=1):
...
+
+    if data.ndim == 0:
+        raise Exception("The dimension of the array should be superior to zero.")
```
Einenlum, line 470: Same, `ValueError` could be used here instead.

src/pymodaq/daq_utils/h5modules.py
```diff
@@ -1881,37 +1897,84 @@ def add_detector(self, det_name, settings):
                                      title='Time axis',
                                      metadata=dict(label='Time axis', units='s', nav_index=0))
 
-    def add_datas(self, datas):
-        det_name = datas['name']
+    def add_datas(self, acquisition):
```
Einenlum, line 1900: No typehint here for `acquisition`?

src/pymodaq/daq_utils/managers/modules_manager.py
```diff
@@ -38,15 +99,19 @@ class ModulesManager(QObject):
         ]},
     ]
 
-    def __init__(self, detectors=[], actuators=[], selected_detectors=[], selected_actuators=[], timeout=10000):
+    def __init__(self, detectors=[], actuators=[], selected_detectors=[], selected_actuators=[], timeout_s=None):
```
Einenlum, line 102: Mutable objects as default parameters should be avoided (see [here](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/)).

The recommended alternative is:

```python
def __init__(
    self,
    detectors=None,
    actuators=None,
    selected_detectors=None,
    selected_actuators=None,
    timeout_s=None,
):
    if detectors is None:
        detectors = []
    if actuators is None:
        actuators = []
    if selected_detectors is None:
        selected_detectors = []
    if selected_actuators is None:
        selected_actuators = []
```

Or:

```python
detectors = detectors or []
actuators = actuators or []
# ...
```

src/pymodaq/daq_utils/daq_utils.py
```diff
@@ -442,11 +442,37 @@ def uncapitalize(string, Nfirst=1):
     return string[:Nfirst].lower() + string[Nfirst:]
 
 
-def get_data_dimension(arr, scan_type='scan1D', remove_scan_dimension=False):
-    dimension = len(arr.shape)
-    if dimension == 1:
-        if arr.size == 1:
-            dimension = 0
+def get_data_dimension(data: np.ndarray, scan_type='scan1D', remove_scan_dimension=False):
```
Einenlum, line 445: You can add typehints for each parameter and a returntype:

```python
from typing import Tuple

def get_data_dimension(...) -> Tuple[Tuple, str, int]:
    ...
```

src/pymodaq/daq_viewer/daq_viewer_main.py
```diff
@@ -489,21 +557,28 @@ def snapshot(self, pathname=None, dosave=False, send_to_tcpip=False):
         except Exception as e:
             self.logger.exception(str(e))
 
-    def grab_data(self, grab_state=False, send_to_tcpip=False):
-        """
-            Do a grab session using 2 profile :
-                * if grab pb checked do  a continous save and send an "update_channels" thread command and a "grab" too.
-                * if not send a "stop_grab" thread command with settings "main settings-naverage" node value as an attribute.
+    def grab_data(self, continuous_grab=False, send_to_tcpip=False):
```
Einenlum, line 560: `-> None:`

src/pymodaq/daq_viewer/daq_viewer_main.py
```diff
@@ -926,23 +1046,26 @@ def get_data_from_viewer(self, datas):
                 self.grab_done = True
                 self.grab_done_signal.emit(self.data_to_save_export)
 
-    def set_datas_to_viewers(self, datas, temp=False):
+    def set_datas_to_viewers(self, datas: List[utils.DataFromPlugins], temp=False):
```
Einenlum, line 1049: `-> None:`

src/pymodaq/daq_viewer/daq_viewer_main.py
```diff
@@ -1039,7 +1162,16 @@ def process_data(self, datas, container):
 
     @Slot(list)
     def show_data(self, datas: List[utils.DataFromPlugins]):
-        """
+        """Transfert the acquired data from the detector (DAQ_Detector) to the UI viewer.
```
Einenlum, line 1165: Transfer

src/pymodaq/dashboard.py
quantumm: These comments are now removed.
The pid module loading is done in Dashboard.set_preset_mode.

src/pymodaq/daq_logger.py
quantumm: Done!

src/pymodaq/daq_utils/daq_utils.py
quantumm: I put TypeError here instead because it is the type that is not good.

src/pymodaq/daq_utils/daq_utils.py
quantumm: Done.

src/pymodaq/daq_utils/h5modules.py
quantumm: Done.

Ok but renaming method will be done maybe latter sounds a bit risky ;)

src/pymodaq/daq_utils/managers/modules_manager.py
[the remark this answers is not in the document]
seb5g: yes I would rather not import such a big object where it"s not needed and we may risk circular imports...

src/pymodaq/daq_viewer/daq_viewer_main.py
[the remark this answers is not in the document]
seb5g: no the dashboard displays several instruments. But one instrment can export many kind of data that would be better rendered in different viewers. For instance the output of a lockin amplifier gives you an amplitude in volt and a phase in degree. Plotting both on the same viewer is misleading and make one of the signal invisible (the amplitude) so you can use this to automatically plot each on separate viewers.

src/pymodaq/daq_utils/managers/modules_manager.py
seb5g: ok you can do it

src/pymodaq/daq_utils/managers/modules_manager.py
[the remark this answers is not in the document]
seb5g: Moreover one could add another ROI between the last grab and the next, so the buffer does not reflects the real data content

src/pymodaq/daq_viewer/daq_viewer_main.py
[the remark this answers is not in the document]
seb5g: if you want efficient you don't want to display everything and moreover you may want to save only raw data. So I guess if you want to log ROI you can stick with "show data"

src/pymodaq/daq_utils/h5modules.py
seb5g: data is already plural, you're right. But it is kind of an early mistake I propagated everywhere in the code...

src/pymodaq/daq_utils/managers/modules_manager.py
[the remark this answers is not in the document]
seb5g: actually in the mods I'm doing at the moment, I made both DAQ_Viewer and DAQ_Move object inherits from a common base class:

`
class ControlModule:
    """Base class common to both DAQ_Move and DAQ_Viewer control modules"""

    @property
    def module_type(self):
        return type(self).__name__

    @property
    def initialized_state(self):
        return self._initialized_state

    @property
    def title(self):
        return self._title`

The property `module_type` is what you could use in the grab_one_detector such as 
`
if detector.module_type != 'DAQ_Viewer':
...
`

I'll use that in the DAQ_Logger to assert whether one of the selected module in the ui of a detector or an actuator. You could also use it.

src/pymodaq/daq_viewer/daq_viewer_main.py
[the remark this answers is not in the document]
seb5g: yes so same answer ;-)

src/pymodaq/daq_logger.py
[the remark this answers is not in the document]
seb5g: ok

src/pymodaq/daq_utils/daq_utils.py
seb5g: ok do it

src/pymodaq/daq_utils/managers/modules_manager.py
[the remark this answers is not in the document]
seb5g: So still not convinced about doing the modification

src/pymodaq/daq_viewer/daq_viewer_main.py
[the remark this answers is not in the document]
seb5g: leave it for the moment, I'll refactore DAQ_Viewer in the next pymodaq4beta step

src/pymodaq/daq_viewer/daq_viewer_main.py
seb5g: that's always the default so not needed

src/pymodaq/daq_viewer/daq_viewer_main.py
seb5g: not needed

src/pymodaq/daq_viewer/daq_viewer_main.py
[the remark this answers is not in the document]
seb5g: just call me if this is still not clear

src/pymodaq/daq_viewer/daq_viewer_main.py
seb5g: ok

seb5g (commented):
still some pending stuff

src/pymodaq/daq_utils/h5modules.py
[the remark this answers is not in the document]
quantumm: Done in 9d18261f7016bdb4283c19a9800176bd8905b251.

src/pymodaq/daq_viewer/daq_viewer_main.py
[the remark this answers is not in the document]
quantumm: Ok I think it is more clear now ;)

src/pymodaq/daq_logger.py
quantumm: Here cd169f35d12eae3e0ae9ed0f6d162853d90264d7

src/pymodaq/daq_utils/daq_utils.py
quantumm: Done in 7295cf508c38e393feeafbf1fa16008e7fa711f6.

src/pymodaq/daq_utils/daq_utils.py
quantumm: Here: cd169f35d12eae3e0ae9ed0f6d162853d90264d7.

src/pymodaq/daq_utils/daq_utils.py
quantumm: Here: cd169f35d12eae3e0ae9ed0f6d162853d90264d7.

src/pymodaq/daq_viewer/daq_viewer_main.py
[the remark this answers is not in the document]
quantumm: Done here: 27d444b8e14f4c298cebd7825e8355cdfcd01104

src/pymodaq/daq_utils/managers/modules_manager.py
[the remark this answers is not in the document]
quantumm: I got back to the original get_det_data_list and remove the grab_one_detector method.
But I am quite convinced we will come back to them at some point ;)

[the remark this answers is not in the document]
quantumm: Here: ffda23ed17300df9b83a2ce517eb3470368c0550.

seb5g:
I'll close this as so many things changed in the meantime (we even learned how to do proper PR ;-) )!

quantumm:
Yes I agree. Probably I will come back to those modifications when we will update the PID module for the v4.

PR #92, opened 2023-01-10
Add documentation for the PID module.

quantumm:


seb5g:
Hi
To properly test the documentation, you should built it locally. Normally everything is present in the repo to do so. You just have to install some more packages as stated in the /docs/requirements.txt:

- releases
- sphinx-csv-filter
- pillow
- toml
- pyqt5
- docutils==0.17.1
- sphinx_rtd_theme

then in you command line `cd` to the path /pymodaq/docs where you have a `make.bat` file and execute:

`make html`

that will process your files and produce local html files holding the doc. You'll have to run it a few times at the beginning so that it remove all/almost all the warnings. You can see them in  the newly create *_build* directory. Then open /docs/_build/html/index.html in your browser and you can check how your doc is looking.

Maybe one should add this explanation into the developer documentation

seb5g:
after compiling your changes I got many errors with the new layout of pymodaq_v4 (expecially it tries to load source files in daq_utils) while it is now in utils. I think I changed this but maybe in the data_management branch. So don't worry about it just focus on your particular changes that I won't touch in the mean time. 
I also don't have figures appearing:
image

so to make them appear, you should actually choose another rst directive such as :
```
   .. _myfig_ref:

.. figure:: /image/PID_Module/interferometer_scheme.png
   :alt: interferometer figure

   A caption for your figure

```

The top is creating a reference (a label) to your figure so you'll be able to quote it with ```:numref:`myfig_ref` ``` see the underscore in the label but not when quoting it in the numref.
then the directive figure is creating the actual figure with the image pointing to it. In you case that would be /image/PID_Module/interferometer_scheme.png

then you can add a caption

seb5g:
you'll end with that kind of figure:
image

that you can make smaller using :

```
   .. _myfig_ref:

.. figure:: /image/PID_Module/PIDModelMock.png
   :alt: interferometer figure
   :figwidth: 300px

   A caption for your figure

```

seb5g:
I let you do all these modifications :-)

quantumm:
It should be fine now :)
I forgot a "/" for the paths of the images.
I used "figure" and put some captions.
Let me know if there are more changes to be done ;)

seb5g:
Hi,
I read it all and it is quite good and thorough. I would however like to see something about the pid_model maybe the code of the beam steering to see that it is actually really easy to configure one. 

As for where and how to declare models, it is now possible to add them in any pymodaq_plugins package as shown on the one you used:
image

You have to create a model folder next to the hardware folder, then python will be able to probe into those as they have been configured as entry-points during installation of the package:

image

then in pymodaq there is a method called get_models that will check all the installed entry points of this type, and make the models available in the UI:
image

is that clear?

Something you could also add is that the actuators and detectors used in a given model should be present in the preset file loaded in the dashboard. In the model, they are referred to using their names defined in the preset.

Could add these 3 things?

quantumm:
Hi!
I added a new section to include the changes you requested.
It is actually more complete now!
Let me know if you see other things to add ;)
David

seb5g:
Good job dude, that's a nice doc chapter!

quantumm:
Cool ;)

Issue #91, opened 2023-01-09
Tests of h5modules


LMSC-NTappy:
Hello @seb5g,

I started working on an improved export feature with new files formats. I figured out verifying tests would be a good starting point. I noticed that most lines in the test file are commented out: 

https://github.com/PyMoDAQ/PyMoDAQ/blob/90cae4a78691cb6b9c05116d295ec001b80d823c/tests/daq_utils_test/h5modules_test.py#L127-L139

Any reason why that is? I get passing tests for the H5BrowserUtils

seb5g:
Hello Nico,
Well it's because those tests went a bit messy and because I'm in the process of heavy change to the data saving mechanisms, only the low level tests should be kept. I'll have to do a go around this when I'm getting confident with the new version. Don't worry about your exporting as it is on the higher level objects and if you follow what we said about object factory applied to exporting, that should be easily portable! Maybe you should built those out of the current modules, a bit like a test module...

LMSC-NTappy:
Hey Seb!

fine with me, thanks for the explanations. I reactivated what I need to work with, we'll see how this goes as I progress with the implementation of new exports.

Cheers

Nico

Issue #90, opened 2022-12-21
Exporting hyperspectral data from h5browser


LMSC-SMarinoni:
Hello!

At EPFL-LMSC we're acquiring hyperspectral maps with PyMoDAQ, and we run into a problem during the export from .h5 to .txt using the h5browser. It looks like only 1D and 2D data can be exported, as the h5browser calls the numpy.savetxt() function.

For the moment, we solved the issue by applying numpy.ndarray.flatten() on the data (saving the data shape in the file header to reconstruct the hyperspectral map later on), but there might be a better solution that could be implemented for the entire community.

Here are the logbook lines:
```
2022-12-21 12:19:07,634 - pymodaq.h5modules - ERROR - Expected 1D or 2D array, got 3D array instead
Traceback (most recent call last):
  File "C:\Users\m\Miniconda3\envs\pmodaq\lib\site-packages\pymodaq\daq_utils\h5modules.py", line 2286, in export_data
    self.h5utils.export_data(self.current_node_path, str(file))
  File "C:\Users\m\Miniconda3\envs\pmodaq\lib\site-packages\pymodaq\daq_utils\h5modules.py", line 1976, in export_data
    np.savetxt(filesavename,
  File "<__array_function__ internals>", line 180, in savetxt
  File "C:\Users\m\Miniconda3\envs\pmodaq\lib\site-packages\numpy\lib\npyio.py", line 1512, in savetxt
    raise ValueError(
ValueError: Expected 1D or 2D array, got 3D array instead
```

Cheers! :)

seb5g:
Hi,
Well it doesn't make any sense to export heavy load data as txt while you have them as packed binary data in the h5 file. I would suggest to directly use the data as is from software able to deal with hdf5: python or others. Export to txt for 1D or 2D is just a helper for people who are used to this kind of data management, but it is really a bad practice...

What do you want to do with the txt files?

LMSC-SMarinoni:
Hello,

Thanks for the reply!
Some of us are just used to this kind of data format and work with scripts based on it.
We will surely try to change our habits towards better practices. But then, I would suggest to disable/hide the possibility of heavy (1D, 2D) data export in .txt format. Otherwise, other users will try this possibility and encounter the same problem.

LMSC-NTappy:
Solved with #93

PR #93, opened 2023-01-10
Export improvements.

LMSC-NTappy:
This PR is a refactoring of the `pymodaq.daq_utils.h5modules`

It improves on export capabilities from the `h5browser` by providing a standard interface for customization. The goal is to make support for new file formats easier. 

The way this is done is a transposition from the following dynamic registry factory pattern onto the `H5BrowserUtil` class (which is transformed into a factory). 

https://medium.com/@geoffreykoh/implementing-the-factory-pattern-via-dynamic-registry-and-python-decorators-479fc1537bbe

Currently, the PR passes tests (export tests have been reactivated, see  #91) but no support for new export formats is provided. This is planned as well, but I want to validate the approach first before spending more time working on this. PR is thus WIP

To register new exports, developers must create a new exporter class obeying the following rules:

0. It must implement the abstract base class `H5Exporter`
1. It must be decorated with the `@H5BrowserUtil.register_exporter()` decorator
2. It must implement the `FORMAT_DESCRIPTION` and `FORMAT_EXTENSION` constants used to identify the format
3. It must implement the `_export_data(self,node,filename) -> None:` method that takes care of saving the data for a given input h5 Node and filename

Example with the refactored ".txt" exporter
https://github.com/LMSC-NTappy/PyMoDAQ/blob/45e379c54d598ad61a9f30d59161073339409571/src/pymodaq/daq_utils/h5node_exporters.py#L58-L65

edit:
Adding the factory into `H5BrowserUtil `didn't work in the end, it created weird runtime errors because the registry was not created at import of the module. I re-factored it to put it as a real factory and separated `H5Backend`. Overall I like this much better because the `h5modules.py` was getting very long anyways. Also took the chance to implement export as a binary numpy (.npy) format which supports data of any dimensionality (Solves #90)


seb5g:
Hi 
I just cloned your code and looked at it: I love it!! that's perfectly what I was imagining!!! The only issue is that you pushed that code onto the 3.x.y version of pymodaq. I'm really into the 4.x.y at the moment with a better splitting of objects and functionnalities into specific folders/modules. I think it is ok to merge your pull into the 3.x.y branch as it will still be around for quite some time but do you think you could look at the data_management branch that has this structure:
image

and see how it could fit there (I think with the way you coded it, it should be fairly easy to do so)

src/pymodaq/daq_utils/h5exporters.py
```diff
@@ -0,0 +1,227 @@
...
+class H5h5Exporter(H5Exporter):
+    """ Exporter object for saving nodes as single h5 files"""
+
+    FORMAT_DESCRIPTION = "Single node h5 file"
+    FORMAT_EXTENSION = "h5"
+
+    def _export_data(self, node, filename) -> None:
+        """Export an h5 node"""
+        #This should allow to get the base file object
+        basefile = node._v_file
```
seb5g, line 112: there is an issue here. For h5 files there are multiple possible backends (hence the h5backend code) and I created a wrapper around those (that I could refactor with a factory by the way...). Because of that the variable *node* here is not a pytables node but a Node object (or child Array classes) from the h5backend module. If you want to access the pytables node you'll have to call node.node (that will return the pytables version of the node if you are using the pytables backend) but would return the h5py version of the node if you use the h5py backend.

using the _v_file property will therefore work only with the pytables backend (not with the h5py or h5pyd). So to make it work quick and dirty and only for pytables, you have to write:
``` 
basefile = node.node._v_file
```
but the best approach would be to add a file method to the Node object (in h5backend) and write a specific call for each backend. thnks ;-)

seb5g (changes requested):
Thanks to apply the modification highlighted fro the h5exporter and a big thanks for the work!!

src/pymodaq/daq_utils/h5exporters.py
LMSC-NTappy: Thanks for the heads up!

Indeed I didn't understand very well this whole back-end deal then. I'll refactor this now

LMSC-NTappy:
Thanks for the feedback and nice comments :)

Sure, I will look into adapting this for the v4 as well. I will open another pull request

src/pymodaq/daq_utils/h5exporters.py
seb5g: I remember also now. This is the only way I found (not too troublesome) to get a node and copy it into a new file. Another solution would be to save an array into a new file and scan and copy all its attributes. That would work for all backend transparently!

LMSC-NTappy: In the end I managed to deal with it. Not in the most elegant way as it involves direct h5pyd/pytables invocation in the code but that will do the trick. In the end I like better a paradigm where backend is entirely abstracted out of the `_export_data` methods even if it involves one or two node tricks.

LMSC-NTappy:
Ok, the way it is now all tests pass (so #91 is done), and two new file formats are supported for exports: numpy `.npy` and hyperspy `.hspy`. This solves #90 as both support arbitrary dimensions, additionally the hyperspy format packages axes data automatically in the hyperspy object. I reactivated all h5browser tests so this solves #91 and changed the opening mode of h5 formats to solve #94

I consider the PR ready for review.

src/pymodaq/daq_utils/h5exporters.py
```diff
@@ -210,18 +217,160 @@ def _export_data(self, node, filename) -> None:
...
             data = node.read()
             if not isinstance(data, np.ndarray):
                 data = np.array(data)
 
-            np.save(filename, data)
\ No newline at end of file
+            np.save(filename, data)
+
+
+# This Exporter depends on the availability of the hyperspy package.
```
seb5g, line 246: I would place this code into a specific folder. Then modify the factory code to import (and then register) any file into that folder. It means we will easily be able to expand functionalities while not modifying core exporters (text, h5 and npy). The other exporters could also (and later) be installed as pymodaq plugin functionnalities from an external package (as plugins should work). I'm saying this as yours is a bit dedicated and the dependency to hyperspy could in fine be included into the plugin package.  but for the time being placing it into a dedicated folder will do.

src/pymodaq/daq_utils/h5utils.py
```diff
@@ -0,0 +1,218 @@
+# Standard imports
```
seb5g, line 1: the methods in this new module have been moved into the browsing module (of data_management branch). I would rather not modify too much this version compared to the modification I did in the data_management branch. But maybe I will regroup this kind of methods into a h5modules/h5utils module also in data_management. So overall it is a good idea, I will transpose it into the v4 branch! 

Did you check that imports modification of these methods has been properly done in modules requiring them? I'm not sure the existing tests extend to this kind of testing...

src/pymodaq/daq_utils/h5exporters.py
```diff
@@ -114,8 +114,15 @@ class H5h5Exporter(H5Exporter):
     def _export_data(self, node: Node, filename: str) -> None:
         """Export an h5 node"""
         # This should allow to get the base file object
-        basefile = node._v_file
-        basefile.copy_file(dstfilename=str(filename), overwrite=False)
+        if node.backend == 'tables':
+            basefile = node.node._v_file
```
seb5g, line 118: well if you found that solution and if it works, you should put the code into the Node object. Something like:
```
def get_file(self):
    pass
```

and even a copy_file method into the Node object

you're removing the backend specifics into the backend module, which is better

src/pymodaq/daq_utils/h5utils.py
LMSC-NTappy: Yes, I ran a whole project check on this modification and it worked fine,

No problem at run time neither.

src/pymodaq/daq_utils/h5exporters.py
LMSC-NTappy: If you want but to be honest I don't like a method that returns ones of these complicated File objects from entirely different libraries.

src/pymodaq/daq_utils/h5exporters.py
LMSC-NTappy: It can be done, however you need to know that the register Exporter Factory relies on classes being imported otherwise the registration is not performed.  So it's not as simple as moving the code in an other file or folder, but it must also be imported properly somewhere. 

I don't mind but it just adds complexity for something that will be changed afterwards anyway if you want to make export formats a real thing (and honestly, at this point in time I am not aware of any effort to support other file formats. You were thinking of something specific?)

src/pymodaq/daq_utils/h5exporters.py
seb5g: Then would it prove useful to create a File object wrapper. When I started this thing I moved up the hierarchy with always reasons to wrap higher objects with a common interface. In that case, it would be the File object. We don't need many methods so should be done fairly easily.

src/pymodaq/daq_utils/h5exporters.py
seb5g: Well my goal is to write a developer documentation containing every point where they can add easily functionalities. Such exporters could be it because maybe some users will need text exporters adding some more info into it as the basic version...If we think about it ahead of the need without making it too complex that will help.
As for the importing , I understand. But then when importing your exporter  factory the __init__ file of the module will be loaded. There you can place a mechanism to import all classes with a given signature (inheriting from your base class for instance) in a given custom_exporters (or equivalent name) module.

LMSC-NTappy:
> But if you prefer to wait for when I merge this branch into pymodaq_dev that is fine (and maybe a better idea)!

Clearly I prefer, otherwise I'll get lost in the refactoring. In any case I guess this can wait a bit, at LMSC we deploy it from the fork  and it will also benefit for in-field testing :)

seb5g (approved):

src/pymodaq/daq_utils/h5backend.py
```diff
@@ -0,0 +1,892 @@
...
+        -------
+        str : full path of the node
+        """
+        if self.backend == 'tables':
+            return self._node._v_pathname
+        else:
+            return self._node.name
+
+    def get_file(self):
+        """ Return node file. If this is a
```
seb5g, line 127: create a File Wrapper that would be returned from this method. In the File wrapper, at least the method copy should be included for all backends.

seb5g (changes requested):

src/pymodaq/daq_utils/h5backend.py
LMSC-NTappy: Clearly it would be the responsibility from the h5backend object, which mostly encapsulates the file object, no?

seb5g: I thought about this and looked int the h5 backend. I think it would be good to remove the few things there that are backend specifics and move them into their respective object. For instance: in the H5Backend object there is a method called `get_children`, this one is typically belonging to a GROUP object and indeed when looking into it i saw I created the children method that has exactly the same content... So in the backend the content of the method could be as simply as:

```
    def get_children(self, where):
        """Get a dict containing all children node hanging from where with their name as keys and types among Node,
        """
        where = self.get_node(where)  # return a node object in case where is a string
        if isinstance(where, GROUP):
             return where.children()
        else:
             return dict([])
``` 
I don't know why I missed that. But checking all backend specifics from H5Backend and moving them into their dedicated objects (Node, GROUP, File (to be created)) would make the code clearer and easier to maintain. Then for our initial purpose, the node group would need a get_file returning a File object itself having a copy method. So it could be called from any node.

I think we should end here the review and accept the merge and it would be great, because you dived into this backend stuff, to apply this in another pull request (if you're willing) (or at least to open an issue, copying this comment content as a starting point)!
With pymodaq you know where you start but you don't know where it will lead you ;-)

seb5g (approved):

src/pymodaq/daq_utils/h5backend.py
seb5g: Let me know if you're ok with merging now?

src/pymodaq/daq_utils/h5exporters.py
LMSC-NTappy: For now I just moved it in a different file and the h5modules.py does the import lifting. In v3 I guess there simply won't be any other file formats added so I guess the module init can wait

src/pymodaq/daq_utils/h5exporters.py
LMSC-NTappy: If you think this is a good idea sure.

src/pymodaq/daq_utils/h5backend.py
LMSC-NTappy: Definitely if you can handle these things from the Node sub classes directly it's better. I would go as far as saying that some methods could be moved outside classes altogether.

LMSC-NTappy:
I am okay with merging now. Thanks for the extensive reviewing, I think we managed to meet half way there :)

seb5g (approved):

seb5g:
I 'll let you work on modifying the h5backend as discussed?

LMSC-NTappy:
Fine with me. I'll try to work on it directly on the v4 branch

Issue #96, opened 2023-02-03
Epsilon parameter of daq_moves is always set to 1 regardless of preset value


rgeneaux:
Hi guys,

I noticed an issue with daq_move_base: if you create a dashboard preset with an actuator, upon loading epsilon will always be set to 1, regardless of preset value.

It seems like this happens here:
https://github.com/PyMoDAQ/PyMoDAQ/blob/ee72fa5947fc628771b9ad99501c5a3bb54d6268/src/pymodaq/daq_move/utility_classes.py#L158

https://github.com/PyMoDAQ/PyMoDAQ/blob/ee72fa5947fc628771b9ad99501c5a3bb54d6268/src/pymodaq/daq_move/utility_classes.py#L183

I commented line 183 and it seems like it works... but maybe this is meant for something, I'll let you actually fix it instead of breaking something!

seb5g:
Hi guys,
in fact _epsilon should be redefined in all of our plugins as class attributes especially because the one developing the plugin should be the one that knows what value is adequate (depends on the controller). So if someone doesn't know much he won't have to play with _epsilon.

I just added this in the template:
```
# _epsilon is the initial default value for the epsilon parameter allowing pymodaq to know if the controller reached
# the target value. It is the developer responsibility to put here a meaningful value
_epsilon = 0.1  # TODO replace this by a value that is correct depending on your controller
```
that said, if epsilon is set in the preset to a different value compared to the default one (_epsilon), one should probably apply the preset one. I'll propose a pull request in that direction

PR #97, opened 2023-02-03
take into account epsilon in preset

seb5g:
This should solve issue #96

### DIFF
--- a/src/pymodaq/daq_move/utility_classes.py
+++ b/src/pymodaq/daq_move/utility_classes.py
@@ -180,9 +180,8 @@ class DAQ_Move_base(QObject):
         self.controller_units = self._controller_units
         self.controller = None
 
-        self.settings.child('epsilon').setValue(self._epsilon)
-
-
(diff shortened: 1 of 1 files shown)


seb5g:
I actually made two commits that solve the issue. But the second is better. With this, the preset should now propose as default the _epsilon defined in the plugin. But if a user change it then this chnage is applied

rgeneaux (changes requested):
Two things:

- line 511 
https://github.com/PyMoDAQ/PyMoDAQ/blob/b340eff05f773c5904e25c192e350e97d5e4c430/src/pymodaq/daq_move/utility_classes.py#L511

should now be `params = comon_parameters() + tcp_parameters`

- In the plugin template there is one call to modify as well, see https://github.com/PyMoDAQ/pymodaq_plugins_template/commit/392743b8d3046bc40dbc1d9c949cee56d46bf261#r99506065

Didn't find any more convenient way to suggest changes haha.
But after that, it works great, thanks!

seb5g:
Hi I pushed the modifications here and update the template package as well

PR #98, opened 2023-02-16
Added a button to sync setpoints with current values

MatthieuGuer:
I added a button in the PID interface which sets the current value returned by the model as the setpoint.
For models with several setpoints, there is one button for each.


seb5g:
Could you motivate the need to add these? Could we not just add one to set all setpoints?

seb5g (approved):
Ok tested it, it works well and is indeed usefull

PR #100, opened 2023-03-02
implemented exporters from pull request #93

seb5g:
This pull request applies the development about exporting data from the h5 ecosystem to various file format written by N Tappy on v3 version of pymodaq. Because of tremendous change between v3 and v4 I reimplemented the code here using new layouts and data management.
Could you please test them and eventually discuss changes?


seb5g:
Hi Reviewer Nico or others!,
I cleaned a bt the source code and made sure it is working for some exporters (didn't do hyperspy...). I'm however thinking that just one simple file extension may be restrictive for the export. For instance the ascii exporter is in fact a txt where the data are put accordingly to specifications from the flimj library. But we could find other programs needing txt file in another layout. Do you think we could identify the exporters with a bit more details? For instance the full file filter!
Cheers

Attolight-NTappy:
Hi Seb!

Looks good, I will try to review it this week :)

PR #106, opened 2023-03-20
bugfix on pause_PID

rgeneaux:
indenting error caused pause button to behave weirdly


quantumm:
This should probably make things better ;)

seb5g (approved):

PR #139, opened 2023-07-21
Fix dashboard.check_version()

rgeneaux:
get_pypi_pymodaq('pymodaq') returns only one version (as string), not a list of versions.


seb5g (changes requested):
Hello,
could you systematically do the PR with the pymodaq_dev branch as the target? Could you do it again?

rgeneaux:
there you go

Issue #141, opened 2023-08-22
Error with the scan extension 


aurore-finco:
Hi!
I have been trying to use the DAQ Scan extension with Pymodaq 4.0.6 and I got the following error:

![image](https://github.com/PyMoDAQ/PyMoDAQ/assets/78025408/c5de3d13-2145-49a2-8c87-7384612de0ca)

This is completely blocking the extension, the viewers cannot even be prepared. I repeated it using only mock instruments, so it does not come from my plugins which are under development. I am using python 3.8.17.

On the indicated line 1210 in daq_viewer.py, I changed `self.single(*command.attribute)` for `self.single(command.attribute[0])` and it seems to solve the issue, but I am not sure that it does not break something that I did not notice somewhere else...

quantumm:
If you have started to develop your plugins under version 3, I would recommand that you stick to it.

aurore-finco:
Hi!
thanks for your answer. Yes, I noticed that plugins from version 3 are not compatible. 
This issue is not related to my own plugins, I get exactly the same error with the Mock instruments. Are these compatible with version 4? I do not have other devices that I could use to test.

quantumm:
Ok you are right, I omitted that you tried with the mocks. They should work for sure. So I guess this is a bug in the DAQ Scan module or the mocks in version 4. But I cannot go much further right now.
If you do not absolutely need the features of version 4 and you need a fast solution, I would recommend that you fall back to version 3.

quantumm:
Good luck !

aurore-finco:
Ok, thanks!

seb5g:
I'm back in Monday at work and will look into it. That's a simple issue of
signature of the functions.


Le mar. 22 août 2023 à 14:40, Aurore Finco ***@***.***> a
écrit :

seb5g:
#142 is the pull request trying to solve this

seb5g:
Hi Aurore, thanks to test it and close the issue if this is fine

aurore-finco:
Yes I tested it, it works fine.
Thanks, I am closing the issue.

PR #142, opened 2023-08-28
patched two issues

seb5g:
one related to viewer_docks and viewers_docks spelling the other with the structure of the attribute "attribute" of the ThreadCommand use with the single method


seb5g:
@aurore-finco this pull request patch the issue #141 you've seen recently. Could you install and try this code?

aurore-finco:
Hi!
Thanks! It looks like what I did quickly to make it work in the lab, so it should be fine. I am away this week, I'll install and test it next week as soon as I am back in the lab.

seb5g:
It's fine then I already tested it here, if it still have issues  keep going in #141 otherwise you can close it as well

Issue #151, opened 2023-09-13
Daq_Scan ignores viewer's "Naverage" parameter


rgeneaux:
Hi there,
I noticed that the parameter Naverage that every viewer has (in its Main Settings) is not taken into account during a daq_scan acquisition.
I guess this is because in the ModulesManager class, grab_datas calls 
`sig.emit(utils.ThreadCommand("single", kwargs))`
which then calls the DAQ_Detector method with Naverage=1:
`def single(self, Naverage=1, args_as_dict={}):`

(noticed it in pymo3.6 but the code is the same in 4.0 so I assume the issue remains).

thanks!

rgeneaux:
Fixed in 462b4a7cd81979316dcec2104dc8c7d0b3a909d2
thanks!

PR #152, opened 2023-09-14
make sure the averaging within daq_viewer is applied when using daq_scan

seb5g:
Attemps to solve #151

### DIFF
--- a/src/pymodaq/control_modules/daq_viewer.py
+++ b/src/pymodaq/control_modules/daq_viewer.py
@@ -313,6 +313,15 @@ class DAQ_Viewer(ParameterManager, ControlModule):
             self.ui.detector = det
         self._set_setting_tree()
 
+    @property
+    def Naverage(self):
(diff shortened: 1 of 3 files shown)


rgeneaux (changes requested):
Averaging works in daq_scan, thanks!

I caught another bug in the process:
here: src/pymodaq/utils/config.py       line 329
https://github.com/PyMoDAQ/PyMoDAQ/blob/08dac95658c6924e78a67b2de949e676618f21a7/src/pymodaq/utils/config.py#L329C46-L329C46

`create_toml_from_dict(toml.dumps(config), toml_base_path)`
should be
`create_toml_from_dict(config, toml_base_path)`

otherwise you get an error when your configuration file doesn't match template

seb5g:
I resolved this bug also in the last commit! Cheers. Would you like to merge the PR if ok and close the related issue?

PR #155, opened 2023-10-04
Scan close

seb5g:


seb5g:
The quit button was raising errors: 

        2023-10-04 15:53:36,943 - pymodaq.daq_scan - ERROR - 'DAQScanUI' object has no attribute 'average_dock'
          File "C:\Miniconda3\envs\v4\lib\site-packages\pymodaq\extensions\daq_scan.py", line 341, in quit_fun
            self.ui.average_dock.close()
        AttributeError: 'DAQScanUI' object has no attribute 'average_dock'

There is no more average dock neither scan_dock, but only a parent window and a ViewerDispatcher.

Issue #159, opened 2023-10-16
loss of detectors in daq_scan


francoisBmallet:
Hi, 
  I am having an issue with data acquisition (and plotting) of multiple detectors. Let's take just 2 0D mocks and one actuator. In daq-scan, It seems like the detector names are not correctly passed when grabbing the data and that the next detector overwrites the previous one. Anyway, impossible to plot both mock detector at the same time.

seb5g:
The new PR #160 solves this issue. Could you test it and close the issue if that's fine? Thanks

francoisBmallet:
Hey thanks, just tested and all good now !

PR #160, opened 2023-10-16
replacing the dwa "origin" by the control module's title

seb5g:
so that extensions grabing data from same plugin but duplicated would not produce the same dwa full name (name+origin) and will not be overwritten when appended into the control module dte


seb5g:
solves this issue #159 because they were having the same name when appended into the data_to_export, so one was overwritting the other!

PR #174, opened 2023-10-23
Fix 168 threading problems

### LINKED ISSUE #168: Daq Viewer "initialize detector" crashes viewer / dashboard
Clicking on "initialize detector" of the 2d or 0d viewer crashes the viewer (and dashboard).

There is no error message in the log, but in the console is written something like "QThread: Destroyed while thread is still running". 

Viewer are the mock viewers or an application of the generic camera viewer.
Pymodaq version is 4.0.8
Pyqtgraph is the most recent version
Python 3.11
Windows 10

How to reproduce:
1. Start the dashboard in cli
2. Create a preset with a mockdetector
3. Load preset
4. Click on initialize detector

bmoneke:
The first commit (2e1ae5b) closes #168
The culprit was to call "init_hardware" instead of "init_hardware_ui" in dashboard

The other commits harmonize viewer and move in "send_init" etc.

I found another bug which stops the dashboard from loading. Catching that error (just a try except clause added) allows the dashboard to load. See #175

### DIFF
--- a/src/pymodaq/control_modules/utils.py
+++ b/src/pymodaq/control_modules/utils.py
@@ -374,6 +374,6 @@ class ControlModuleUI(CustomApp):
         """
         raise NotImplementedError
 
-    def send_init(self):
-        """Should be implemented to send to the main app the fact that someone pressed init."""
+    def send_init(self, checked: bool):
+        """Should be implemented to send to the main app the fact that someone (un)checked init."""
         raise NotImplementedError
--- a/src/pymodaq/control_modules/utils.py
+++ b/src/pymodaq/control_modules/utils.py
@@ -102,6 +102,7 @@ class ControlModule(QObject):
     _update_settings_signal = Signal(edict)
     status_sig = Signal(str)
     custom_sig = Signal(ThreadCommand)
+    ui = None
 
     def __init__(self):
         super().__init__()
@@ -281,7 +282,8 @@ class ControlModule(QObject):
         -----
         This method should be preferred to :meth:`init_hardware`
         """
-        raise NotImplementedError
+        if self.ui is not None:
+            self.ui.do_init(do_init)
 
     def show_log(self):
         """Open the log file in the default text editor"""
@@ -373,5 +375,5 @@ class ControlModuleUI(CustomApp):
         raise NotImplementedError
 
     def send_init(self):
-        """Shoudl be implemented to send to the main app the fact that someone pressed init"""
+        """Should be implemented to send to the main app the fact that someone pressed init."""
         raise NotImplementedError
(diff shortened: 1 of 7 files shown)


bmoneke:
Opening a viewer plugin file uses the same error (`init_hardware` instead of `init_hardware_ui`), therefore there you could not choose a different detector either.
Commit 915fefb fixes this.

src/pymodaq/control_modules/utils.py
```diff
@@ -373,5 +375,5 @@ def do_init(self, do_init=True):
         raise NotImplementedError
 
     def send_init(self):
```
seb5g, line 377: should also change this send_init signature here with a checked boolean argument

src/pymodaq/dashboard.py
```diff
@@ -755,24 +755,27 @@ def set_file_preset(self, filename):
...
+                                    if plugin['status'] != "Slave":
+                                        logger.error('error in the master/slave type for plugin {}'.format(plug_name))
+                                    if plug_init:
+                                        actuators_modules[-1].controller = master_controller
+                                        actuators_modules[-1].init_hardware_ui()
+                                        QtWidgets.QApplication.processEvents()
+                                        self.poll_init(actuators_modules[-1])
+                                        QtWidgets.QApplication.processEvents()
+                            except Exception as e:
+                                logger.exception(str(e))
```
seb5g, line 778: but if there is an error here one should actually show a message to the user, maybe a popup stating that no master has been configured yet (or initiliazed) and that the preset should be reviewed...

Maybe log this specific error and raise again an ActuatorError that would be catched by the next try statement, closing the splashscreen and quitting the application

seb5g (commented):
Hi, thanks for these changes, and the work on symetrizing the use of send_init and init_hardware stuff. I raised a question on the dashboard modification that would need some more discussion...

src/pymodaq/dashboard.py
bmoneke: For proper handling I opened issue #175, as I did not know, what your intention is.

Therefore this simple solution.

If it is fine with you, I'll raise that error, which is handled better, and in #175 a proper solution is discussed and implemented.

src/pymodaq/control_modules/utils.py
bmoneke: I'll do it

src/pymodaq/dashboard.py
seb5g: ok let's do that

src/pymodaq/control_modules/utils.py
bmoneke: Done

src/pymodaq/dashboard.py
seb5g: that's fine with me. We'll do another PR from issue #175

Issue #188, opened 2023-10-30
How to handle multiaxes controllers with a priori unknown number of axes


rgeneaux:
Hi,
There is a question I've ran into when writing several actuators plugins, and i'd like to know if someone has an opinion. 
For a multiaxis controller, we now define something like (cf [doc](http://pymodaq.cnrs.fr/en/pymodaq-dev/developer_folder/instrument_plugins.html#actuator-plugin-having-multiple-axis))
`_axis_names = {'Xaxis': 0, 'Yaxis': 1, 'Zaxis': 2}`

However, my issue is that some controllers can handle an arbitrary number of axes - some just daisy chain each other and could potentially have dozens of axes, but we can't know that before initializing the self.controller.

The initialization procedure goes something like
- when importing the plugin, run some API method to search for the connected controllers, generate a list and put that into the parameter tree
- the user chooses which controller to use in the list, then clicks initialize
- the controller uses an API method that lists all available axes. We can then populate `_axis_names` with those axes and the user can swap.
This is fine 

The issue arises when using a preset: ideally we'd like a preset to launch and directly initialize the controller with a chosen axis. However there is no way to define a preset in this way, since the `_axis_names` has not been populated yet when we create the preset (presets are defined before initializing the controller).

We've hacked our way around this in two ways for different plugins, 

1. you can choose that this type of controller will have a fixed number of axes and write `_axis_names = {'Axis 1': 0, 'Axis 2': 1, 'Axis 3': 2}` at the top of the plugin, then update the names once the controller knows them (this is SmaractMCS2). 

2. Another thing I did was to redefine self.settings['multiaxes', 'axis'], instead of a dictionnary it becomes an integer and the user can just use whatever index they want in the preset, and the controller will try to initialize this index (this is Zaber). It writes the names of the axes into another field of the tree once it is available. 

Neither solutions are user friendly: when defining the preset, we have to remember which axes are what (they don't have their names yet). Second solution at least has the advantage of handling arbitrary number of axes.

So I guess we need to find some standardized way of doing that! Thoughts?

seb5g:
To add some thoughts (as requested ;-) ), have you noticed in the preset the little "**plus**"" button on the right part of the combo widget? Well if you press it it will allow you to add entries into the combo:

![image](https://github.com/PyMoDAQ/PyMoDAQ/assets/6609955/afd8f769-86b8-46d2-a659-cdb8719cb146)

[...]

PR #189, opened 2023-10-30
now one can add entries to a combo widget in the settings with a dict under the hood

seb5g:
first specify a str entry to be added in the list then an integer value to be associated with this key!

image

### DIFF
--- a/src/pymodaq/utils/parameter/pymodaq_ptypes/list.py
+++ b/src/pymodaq/utils/parameter/pymodaq_ptypes/list.py
@@ -86,13 +86,24 @@ class ListParameterItem(ListParameterItem):
             --------
             limitsChanged,
         """
-        if type(self.param.opts['limits']) == list:
+        if isinstance(self.param.opts['limits'], list):
             text, ok = QtWidgets.QInputDialog.getText(None, "Enter a value to add to the parameter",
                                                       "String value:", QtWidgets.QLineEdit.Normal)
             if ok and not (text == ""):
                 self.param.opts['limits'].append(text)
                 self.limitsChanged(self.param, self.param.opts['limits'])
                 self.param.setValue(text)
+        elif isinstance(self.param.opts['limits'], dict):
+            text, ok = QtWidgets.QInputDialog.getText(None, "Enter a text to add to the parameter",
+                                                      "String value:", QtWidgets.QLineEdit.Normal)
+            if ok and not (text == ""):
+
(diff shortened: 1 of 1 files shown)


seb5g:
related to #188 and PR #171

seb5g:
@rgeneaux you're in charge for the review. When done and approved, I'll merge it!

rgeneaux:
You're too fast :-) Haven't had time to try it out yet

seb5g:
sorry I was doing a bit of branch cleaning...

rgeneaux:
I was testing on a MockNamesAxes (lab computer is busy right now) and there is an issue when loading the preset

at /pymodaq/utils/parameter/utils.py:164 inside `def set_param_from_param(param_old, param_new):`

```
  elif isinstance(child_old.opts['limits'], dict):
      if child_new.value() not in child_old.opts['limits'].values():
          child_old.opts['limits'].update(dict(str(child_new.value()), child_new.value()))
```
the last line bugs:      dict should take a list of tuples. 
Also the new dictionary item to add to child_old.opts['limits']  should be something like {'Xaxis', 1}, not {str(1), 1}.

i'm unsure if other types of params use dictionary so I'll let you check it instead of breaking other stuff!

seb5g:
I produced a battery of tests to make sure all kinds of Parameters can be encoded to xml and back
The method set_param_from_param is also fine

Could you check this using code from commit: https://github.com/PyMoDAQ/PyMoDAQ/commit/f0b6454865e04ec162bf30587926532f45176e79

seb5g:
Actually this is the one:
https://github.com/PyMoDAQ/PyMoDAQ/commit/0ce46669ef6912e77761a60642ed1ce00a148ee0

rgeneaux:
Ok so now (on pymodaq-dev branch + the changes to set_param_from_param), it works well to add a dict in the preset!
However, for some reason, the value that we added does not make it to the final daq_move. See test with one axis hard-coded in the plugin, the other added in the preset:
axes_test

What's strange is after initializing the daq_move and putting a debug point somewhere,
```
self.settings.child('multiaxes', 'axis').opts['limits']
Out[4]: {'DefinedInsidePlugin (value=1)': 1, 'DefinedInPreset (value=2)': 2}
```
the second value is there! No idea why it doesn't actually appear ?!

seb5g:
that's because the Parameter doesn't notify the widget item that it's limits have changed. Settings the limits using `opts['limits']` will not notify. One should use the `child.setLimits(new_limits)` method. Do you want to try it?

PR #196, opened 2023-11-03
Pymodaq dev

Ashwolaa:
- Adding base class for 1D scan to separate linear/non linear
- Adding some 1D scan possibilities:

1. Introducing the staggered type for 1D scan(0,1,2,3,4)==)0,2,4,3,1 
2. Introducing the parsed type for 1D scan from matlab syntax, easy setup for scans: 
"start:step:stop" such as:
0:0.2:1 ==) [0,0.2,0.4,0.6,0.8,1]
0:0.2:1, 5 ==) [0,0.2,0.4,0.6,0.8,1,5]
0:0.2:1, 5:2:9 ==) [0,0.2,0.4,0.6,0.8,1,5,7,9]
comma can be exchanged for linebreak

### DIFF
--- a/docs/src/index.rst
+++ b/docs/src/index.rst
@@ -132,6 +132,7 @@ Please see :doc:`the changelog </changelog>`.
    glossary
    api/api_doc
    PyMoDAQ Femto <https://pymodaq-femto.readthedocs.io/en/latest/>
+   PyMoDAQ Plugins DAQmx <https://pymodaq.github.io/pymodaq_plugins_daqmx/>
 
 ..   usage/classDiagram
 


seb5g:
Those are funny scans!

You fully understood how to do this. Now I'm thinking:

- Whether we should add them as regular/built-in scans or in a plugin (no mechanism to do that yet but I was planning to do it)
- How to best explain to users how the scan is working: We should probably have a tooltip appearing with the content of the class docstring. Then we should think of a nice way and standardized way to write this docstring (even for the existing ones). Do you want to go and try this (I mean proposing a standardized docstring layout)? Maybe in a separated discussion? 

as for the question on the ordering bit, I don't understand it...

Finally, you should create a new branch with meaningful name everytime you work on something new. Like this you can still use the dev one being synchronized with mine for last features and testing your own functionnalities without breaking it.

Something like: 
pymodaq_dev@remote ->new_1d_scans branch ->PR->pymodaq_dev@origin

seb5g:
by the way you also need to produce some tests to check that your classes are doing what you expect them to do. If you need some input on that, we can discuss it

Ashwolaa:
Thanks for the feedback, that's a good point for the branch part. I'll make it on specific ones from now on.

Concerning the ordering, I differentiate the values that will be explored during the scans to the order in which they are explored.
In fact, linear, random and staggered all go through the same points. Thus, in that case, a more general frame would be a 1d linear evenly space scan with some additional choices on how to go through it. 
So a factory would be dedicated to the generation of the position array while an other factory would be related on how the position array gets shuffled.

I think a plugin would be a good idea overall for many features. The base pymodaq could just have linear scans while some plugins would allow Randon, backwards, staggered or some more fancy way to make the position array.

I'll look into the tooltip, I wanted to add them thinking that you could directly add it to a parameter item (just as an additional opt) but I could not find this option unfortunately in pyqtgraph. Did you make it? Otherwise we could subclass parameteritem to add it as it is a very useful feature.


I have no real clue on the testing part so if you have some built-in code/procédure to dot it, I'll gladly take it.

seb5g:
I worked on these things. now pymodaq can load scanners from plugins. I used your sparsed one (citing your work ;-) ) as an example of how to. I've put it into the pymodaq_plugins_mockexamples where I will store various examples of plugin features. In there there is a directory called `scanners` where I've put your Sparsed class. 

If this plugin is installed with its plugin features being correct: 
image

then your new 1D scan will appear. I also corrected a few typos/bugs I noticed (like evaluate_steps), update the setting title...

I also added the tooltip on the scan settings that display the content of the docstring (but these should be better written as we discussed)...
image

and finally in pymodaq/tests/utils_test/scanner_test/... I've started to write a test layout. One should act pytest things to assert our class is doing what  we expect or not expect (by raising errors). In your case it could be if some characters are wrong then it should do something, ...

Could you have a look to all this? So install pymodaq from the custom_scanners branch, then install the pymodaq_plugins_mockexamples plugin (all this as developer) and let me know if ok. Once we are ok, you will do a new PR to include the Sparsed scanner into pymodaq as I think it is very nice!! OK?

seb5g:
Hello,
there is an issue here. You pushed all your new code into this same branch it means both the scan stuff and the settings stuff are mixed together... It is then impossible to review test your modifications.
You have to create a branch for each  and then create 2 separated PR to PyMoDAQ/pymodaq@pymodaq_dev

Could you do it again please?

seb5g:
as for the icons, i'll show you how to add them once you do the new PR

seb5g:
so my advice: 

- start a new branch called sparsed_scan deriving from my uptodate pymodaq_dev branch. check it out. Copy the files/code modification concerning the sparsed scan and do a PR
- start a new branch called save_load_settings deriving from my uptodate pymodaq_dev branch. check it out. Copy the files/code modification concerning the settings and do a PR

Ashwolaa:
Ah it's so confusing, I messed it up trying to separate everything in different branches. I look into that

Ashwolaa:
Ok I tested 

> I worked on these things. now pymodaq can load scanners from plugins. I used your sparsed one (citing your work ;-) ) as an example of how to. I've put it into the pymodaq_plugins_mockexamples where I will store various examples of plugin features. In there there is a directory called `scanners` where I've put your Sparsed class.
> 
> If this plugin is installed with its plugin features being correct: image
> 
> then your new 1D scan will appear. I also corrected a few typos/bugs I noticed (like evaluate_steps), update the setting title...
> 
> I also added the tooltip on the scan settings that display the content of the docstring (but these should be better written as we discussed)... image
> 
> and finally in pymodaq/tests/utils_test/scanner_test/... I've started to write a test layout. One should act pytest things to assert our class is doing what we expect or not expect (by raising errors). In your case it could be if some characters are wrong then it should do something, ...
> 
> Could you have a look to all this? So install pymodaq from the custom_scanners branch, then install the pymodaq_plugins_mockexamples plugin (all this as developer) and let me know if ok. Once we are ok, you will do a new PR to include the Sparsed scanner into pymodaq as I think it is very nice!! OK?

Ok I tested it. Installed the pymodaq from custom_scanners, install pymodaq_plugin_mockexamples and it seems to work very nicely.

 I think it is better to use the text type parameter item (instead of string), I replaced it in the Mock and the scanner updated properly. So this seems all good for me. I'll add the sparse type to pymodaq now, so directly in the **_1d_scanners.py** if I understood correctly, is that right?

seb5g:
> I think it is better to use the text type parameter item (instead of string), I replaced it in the Mock and the scanner updated properly. So this seems all good for me. I'll add the sparse type to pymodaq now, so directly in the **_1d_scanners.py** if I understood correctly, is that right?

Ok for the text parameter. So now you can add the sparse class in the _1d_scanners.py of a dedicated branch created from uptodate pymodaq_dev and I'll merge it!

PR #199, opened 2023-11-05
Dashboard preset error

seb5g:
Better handling of the possible error when loading a preset in terms of:
* error from the instrument plugin class
* error if not right configuration for slave/Master
* error if not init for slave/master config


seb5g:
solve the issue #168

PR #200, opened 2023-11-06
ParameterTree settings in daq_viewer:

Ashwolaa:
- Loading/saving possibilities in daq_viewer
TODO:
- differentiate main settings from detector settings?
- extend it to daq_move ( should be a more general feature associated to each parameter settings)
- add some safety check to avoid breakup when detector already initialized
- fix the icons
- make custom name/id to avoid changing plugins
- Rewrite ioxml.parameter_to_xml_file and ioxml.XML_file_to_parameter to prefer differentiation by children instead of type group


seb5g:
- differentiate main settings from detector settings?
I guess saving loading main settings is also interesting

- extend it to daq_move ( should be a more general feature associated to each parameter settings)
even more general, it should be associated with the ParameterManager object (so that one could save/load everywhere)

- add some safety check to avoid breakup when detector already initialized
could be tricky indeed

- fix the icons
see next comment
- make custom name/id to avoid changing plugins
what do you mean?

- Rewrite ioxml.parameter_to_xml_file and ioxml.XML_file_to_parameter to prefer differentiation by children instead of type group

What do you mean?

seb5g:
For the icons, once you have the new png files, you have to first add them in the xml file (suffixed *.qrc) :
image

Then process these to produce a python file (containing bytes. these are the content of all the png files) usable by qt... To do it, it's a bit more tricky. You have to install a python package called `pyqt5-tools`. Once installed you'll have a script called `pyrcc5` that will transform the qrc file into the python one:

in the command line cd to the location of your pymodaq resources:
`cd ~\pymodaq\resources\QtDesigner_Ressources`
then:
```
pyrcc5 QtDesigner_ressources.qrc > QtDesigner_ressources_rc.py
```

then your icons should appear

Ashwolaa:
- Rewrite ioxml.parameter_to_xml_file and ioxml.XML_file_to_parameter to prefer differentiation by children instead of type group
Currently, the functions are looking for the group type to check if they have children. However, any parameter tree item can have its own children so the group type should not be the main criteria to differentiate that.


- make custom name/id to avoid changing plugins
I think each file should be attributed to a specific plugin, this would avoid the loading of settings from a different plugin to the current used plugin which is likely to create some bugs. A identifier to only allow "load settings" associated to the currently initialized or selected plugin would avoid that.

I'll include the icons, thanks for the procedure

seb5g:
I finished checking the impact of the new ParameterManager it should be fine now. I'll let you work on the Parameter comparison

Ashwolaa:
Shall we close this PR and move everything to settings for now?

seb5g:
Yes

PR #205, opened 2023-11-08
Settings

Ashwolaa:
Adding three functions to compare parameters
compareParameters : return True if all opts are the same between two parameters
compareStructureParameter  : return True if structures are the same between two parameters
compareValuesParameter : return True if values are the same between two parameters

These functions use respectively the three functions getOpts, getStruct and getValues that are inspired by the Parameter.getValues function which returns an ordered dict.

TODO: 
If the struct is the same, should we make an order dic containing bool information to find mismatch in values or in opts?


Ashwolaa:
The syntax could also be simplified to only compare selected opts but it just requires a bit of work to avoid buggy features. I may look into that later to have a single function to make the OrderedDict.

src/pymodaq/control_modules/daq_move.py
seb5g, line 1: this modification should not be in this PR. I just rebased the settings branch from the pymodaq_dev modification that included this patch. So you should just **fetch the settings branch from upstream** to update your branch and these should be applied automatically

src/pymodaq/control_modules/daq_viewer.py
seb5g, line 1: this modification should not be in this PR. I just rebased the settings branch from the pymodaq_dev modification that included this patch. So you should just **fetch the settings branch from upstream** to update your branch and these should be applied automatically

src/pymodaq/extensions/daq_scan.py
seb5g, line 1: this modification should not be in this PR. I just rebased the settings branch from the pymodaq_dev modification that included this patch. So you should just **fetch the settings branch from upstream** to update your branch and these should be applied automatically

src/pymodaq/utils/parameter/utils.py
seb5g, line 1: this seems fine but I would like you to write the tests to make sure your methods are doing what is expected of them. I'm sure you tested them on a few examples, so it should be fairly straightforward. If you don't know how/where to start, send me the things you tested so far and I'll include them in a test framework

src/pymodaq/utils/parameter/utils.py
```diff
@@ -33,6 +34,38 @@ def get_param_path(param):
         param = param.parent()
     return path[::-1]
 
+def getOpts(param:Parameter,):
+    """Return a tree of all opts that are children of this parameter"""
```
seb5g, line 38: include doscstring following numpy style

src/pymodaq/utils/parameter/utils.py
```diff
@@ -33,6 +34,38 @@ def get_param_path(param):
         param = param.parent()
     return path[::-1]
 
+def getOpts(param:Parameter,):
```
seb5g, line 37: you can also add the expected return type

src/pymodaq/utils/parameter/utils.py
seb5g: actually this is the same remark for all your methods!

seb5g (changes requested):
Good job but require some polishing and tests!

src/pymodaq/control_modules/daq_move.py
Ashwolaa: I am not sure I got that point, I tried with the current procedure but it did not do much:
I added the upstream
**git remote add upstream https://github.com/PyMoDAQ/PyMoDAQ.git**
Then fetching
**git fetch**
But nothing seems to happens in the PR 

If instead I want to sync the fork I either need to discard the 12 commits or merge it on my own.

Ashwolaa:
I noted a maybe not so unexpected behavior:
- if one would load settings in a daq_viewer from a different detector, the settings tree would correctly update
- if one would then press init detector, the detector would be initialized from the settings tree and not from the daq_type, detector combo box.
So when we load settings, should these two values be updated directly? Or should they be discarded to only be included in the settings tree (with the push button + LED, etc...) ? Or should this button not be accessible for daq_viewer/move?

seb5g:
as you say it's not really unexpected, it's just that normally the settings update/loading is only done from the selection of the detector/actuator type. I believe the load settings button should not be appearing in that particular case (as you suggested) but only the save/update.  Could you also work on the documentation to explain your new feature related to the parameter manager? I've added a (empty) section in the developer documentation about this that would need to be completed. Could you describe your part of the thing?

Ashwolaa (commented):
test and numpy style have been implemented

src/pymodaq/utils/parameter/utils.py
seb5g: I don't like the fact to modify the widget depending on what is in your action_list. I would leave the ParameterWidget as it is with all actions. Then depending the action list in the ParameterManager, We would show (or not) the corresponding action (button) using a setVisible method. It means also the connection will always be there between the button and the corresponding method but is the button is not visible, the user cannot trigger the action so we're fine. Like this the object stay always the same, it is just we are allowing or not some of its features. And by default, I would allow all actions  (in the __init__)

src/pymodaq/utils/parameter/utils.py
```diff
@@ -34,6 +35,86 @@ def get_param_path(param):
...
+        Parameters
+        ----------
+        param1: Parameter
+        param2: Parameter   
+        
+        Returns
+        -------
+        Bool    
+    """    
+    return getStruct(param1)==getStruct(param2)
```
seb5g, line 104: add a space around the ==

src/pymodaq/utils/parameter/utils.py
```diff
@@ -34,6 +35,86 @@ def get_param_path(param):
...
+        Parameters
+        ----------
+        param1: Parameter
+        param2: Parameter   
+        
+        Returns
+        -------
+        Bool    
+    """    
+    return getValues(param1)==getValues(param2)
```
seb5g, line 117: add a space around the ==

src/pymodaq/utils/parameter/utils.py
```diff
@@ -52,7 +133,8 @@ def iter_children(param, childlist=[]):
     """
     for child in param.children():
         childlist.append(child.name())
-        if 'group' in child.type():
+        if child.hasChildren():
```
seb5g, line 136: ok

seb5g (changes requested):

src/pymodaq/utils/parameter/utils.py
Ashwolaa: _I don't like the fact to modify the widget depending on what is in your action_list. I would leave the ParameterWidget as it is with all actions. Then depending the action list in the ParameterManager, We would show (or not) the corresponding action (button) using a setVisible method. It means also the connection will always be there between the button and the corresponding method but is the button is not visible, the user cannot trigger the action so we're fine. Like this the object stay always the same, it is just we are allowing or not some of its features. And by default, I would allow all actions (in the init)_

You are right, I did not make full use of the add_action function which allows to define if it should be visible or not.
I added it in the setup_actions, the button is only showed if its associated action is given in the action list

src/pymodaq/utils/managers/parameter_manager.py
```diff
@@ -3,22 +3,19 @@
...
-from pymodaq.utils.parameter import Parameter, ParameterTree, ioxml
+from pymodaq.utils.parameter import Parameter, ParameterTree, ioxml, utils
 from pymodaq.utils.gui_utils.file_io import select_file
 from pymodaq.utils.config import get_set_config_dir
 
 
 class ParameterTreeWidget(ActionManager):
 
-    def __init__(self):
+    def __init__(self,action_list=[]):
```
seb5g, line 13: add a space after the comma

src/pymodaq/utils/managers/parameter_manager.py
```diff
@@ -29,22 +26,30 @@ def __init__(self):
 
         self.tree.setMinimumWidth(150)
         self.tree.setMinimumHeight(300)
+        
+        self.setup_actions(action_list) # Making the button
+        if len(action_list) != 0: # If no action allowed, no need for splitter
```
seb5g, line 31: I don't like having a widget in some cases and not in other cases because if one calls it from an external program it could lead to an error... Its no big deal to leave the splitter especially because it is closed by default...

src/pymodaq/utils/managers/parameter_manager.py
```diff
@@ -29,22 +26,30 @@ def __init__(self):
...
+            self.widget.layout().addWidget(self.tree)
 
-        self.splitter.addWidget(toolbar)
-        self.splitter.addWidget(self.tree)
-
-        self.splitter.setSizes([0, 300])
-        self.setup_actions()
-
-    def setup_actions(self):
+    def setup_actions(self,action_list:list = []):
```
seb5g, line 41: add space after comma

src/pymodaq/utils/managers/parameter_manager.py
```diff
@@ -29,22 +26,30 @@ def __init__(self):
...
-
         See Also
         --------
         ActionManager.add_action
         """
-        self.add_action('save_settings', 'Save Settings', 'saveTree', "Save Settings")
-        self.add_action('load_settings', 'Load Settings', 'openTree', "Load Settings")
+        self.add_action('save_settings', 'Save Settings', 'saveTree', "Save current settings in an xml file",
+                        visible='save' in action_list)
+        self.add_action('update_settings', 'Update Settings', 'updateTree', "Update the settings from an xml file, the settings structure loaded must be identical to the current one",
```
seb5g, line 49: I think the line is a bit long and should exceed the authorized length just before "Update the settings ... enter a new line

src/pymodaq/utils/managers/parameter_manager.py
```diff
@@ -29,22 +26,30 @@ def __init__(self):
...
+                        visible='update' in action_list)                
+        self.add_action('load_settings', 'Load Settings', 'openTree', "Load current settings from an xml file, the current settings structure is erased and is replaced by the new one",
```
seb5g, line 51: same as above

seb5g (changes requested):

seb5g (approved):

PR #206, opened 2023-11-09
fix typo in move_utility_classes.py

rgeneaux:
let me know if this is the wrong branch for bugfixes - i'm a bit confused


seb5g:
yes it is confusing, that's why I did a bit of cleaning, the branch to use for testing the prerelease and patching it is:
https://github.com/PyMoDAQ/PyMoDAQ:4.0.x_dev)](https://github.com/PyMoDAQ/PyMoDAQ/tree/4.0.x_dev

PR #209, opened 2023-11-10
Use setLimits in set_param_from_param

rgeneaux:
took me a while to figure out that .copy() was necessary :-)


seb5g (approved):

seb5g:
Could you add a test to make sure the widget items are properly updated when using set_param_from_param? You can have a look to the one I wrote in Param_utils_test to get the widget from a tree.

rgeneaux:
sorry I can't find where you do the test on the widget! I checked tests/utils/parameter_test/param_utils_test.py in branch settings or pymodaq-dev. is it somewhere else?

PR #213, opened 2023-11-15
Adding checkbox for listitem in parameter types

Ashwolaa:
Order of click is not preserved but no real need for now


Ashwolaa:
checkbox is accessible by using the type itemselect with the opts['checkbox']=True

seb5g:
could you explain a bit more what's your PR is about in terms of functionnality, final goal, show maybe some pictures... because I have to dig too much into it and it takes me too much time to review!
thx

Ashwolaa:
Sorry, let me recap.

**Summary:** I edited the itemselect parameter types to allow the selection and non-selection with checkboxes instead of clicking.
**Functionality:** Give a direct vision and more resilient selection process in the itemselect parameter type, the selected elements are the one with the checkbox ticked.
**Final goal:** It is just a parameter type so it just gives more possibility for the user. However, I would like to use it in the scanner afterwards to select the actuators as I think it is more appropriate than the currently used item_select.
**Use:** The parameter type is still item select, the new functionality is called by using the additional opts['checkbox'] with the value True.

I added it in the parameter_ex.py such that it goes in the same category as the Selectable items
checkable

src/pymodaq/utils/parameter/pymodaq_ptypes/itemselect.py
```diff
@@ -68,7 +68,6 @@ def set_value(self, values):
         if allitems != values['all_items']:
             self.clear()
             for value in values['all_items']:
-                # if value not in allitems:
                 item = QtWidgets.QListWidgetItem(value)
                 item.setFlags(item.flags() | QtCore.Qt.ItemIsUserCheckable)
```
seb5g, line 72: is it not missing here the option checking to see if one should add the usercheckable item?
something like:
```
if 'check' in opts and opts['check'] is True:
    item.setFlags(item.flags() | QtCore.Qt.ItemIsUserCheckable)
else:
    item.setFlags(item.flags())
```

seb5g (changes requested):
I like this possibility indeed!
But please look at my comments!

src/pymodaq/utils/parameter/pymodaq_ptypes/itemselect.py
Ashwolaa: I do not think that it is missing. The way I did it was by adding a new class ItemCheck which is very similar to ItemSelect. 
I chose this option to immediately differ from the start if the checkbox should be used or just standard item selection.  The reason for it is that they use different functions overall:
- the signal connection: item.changed vs item.SelectionChanged
- the selection: self.selectedItems vs self.all_items() if item.checkState()!=0
- the initialization: self.addItem vs item, item.setFlag, item.setCheckState, self.addItem

However, I am ok to use a single class giving the opts checkbox in the init if you prefer.

Ashwolaa:
By the way, currently we can add with the pushbutton new entry with the same text as existing one, are we ok with that as we cannot edit them currently?
image

Otherwise, I will add a dummy test in the buttonClicked function

  if text in self.param.value()['all_items']:
      print('Entry already exists, please use a different name.')
      return

seb5g:
yes you can add this because otherwise it would not be possible to differentiate between 2 entries.

seb5g:
your itemcheck is redondant with the itemselect. You can either: inherit from itemselect and redefine only the method that is different, or modify itemselect to take into account the display or not of the checkboxes. I would prefer the second option because the change is minimal and easily optional!

src/pymodaq/examples/parameter_ex.py
```diff
@@ -67,8 +67,14 @@ class ParameterEx:
...
              'value': dict(all_items=['item1', 'item2', 'item3'], selected=['item2']),
-             'tip': 'If show_pb is True, user can add items to the list', 'show_pb': True},
-        ]},
+             'tip': 'If show_pb is True, user can add items to the list', 'show_pb': True,},
+            {'title': 'Checkable items', 'name': 'itemsbisbis', 'type': 'itemselect',
+             'value': dict(all_items=['item1', 'item2', 'item3'], selected=['item2']),
+             'tip': 'If checkbox is True, user can select item by checking/unchecking items', 'show_pb': True,'checkbox':True,},                        
+            {'title': 'Dragable items', 'name': 'itemsbisbisbis', 'type': 'itemselect',
+             'value': dict(all_items=['item1', 'item2', 'item3'], selected=['item2']),
+             'tip': 'If dragdrop is True, user can drag or drop items inside the list', 'show_pb': True,'checkbox':True,'dragdrop':True},
```
seb5g, line 76: is there a use in allowing drag/drop?

src/pymodaq/utils/parameter/pymodaq_ptypes/itemselect.py
```diff
@@ -5,29 +5,34 @@
 
 
 class ItemSelect_pb(QtWidgets.QWidget):
-    def __init__(self):
+    def __init__(self,checkbox=False,):
```
seb5g, line 8: add space after comma

src/pymodaq/utils/parameter/pymodaq_ptypes/itemselect.py
```diff
@@ -5,29 +5,34 @@
...
         self.hor_layout.addWidget(self.add_pb)
         self.hor_layout.setSpacing(0)
-
         self.setLayout(self.hor_layout)
 
 
+
 class ItemSelect(QtWidgets.QListWidget):
-    def __init__(self):
+    def __init__(self,hasCheckbox=False):
```
seb5g, line 33: add space after comma

src/pymodaq/utils/parameter/pymodaq_ptypes/itemselect.py
```diff
@@ -36,9 +41,13 @@ def get_value(self):
             Returns
             -------
             dictionnary
-                The dictionnary of all_items compared to the slelectedItems.
+                The dictionnary of all_items compared to the selectedItems.                                
         """
-        selitems = [item.text() for item in self.selectedItems()]
+        if self.hasCheckbox:            
+            selitems = [item.text() for item in self.all_items() if item.checkState()!=0]
```
seb5g, line 47: What happens when you click on the item (as we did previously)? Does the check box becomes checked? Or is it possible to be selected with a checkbox not checked?

seb5g: When you check the box, is the item put in the selected list?

seb5g: I'd like you to write tests to verify this behaviour please

src/pymodaq/utils/parameter/pymodaq_ptypes/itemselect.py
```diff
@@ -84,7 +98,17 @@ def makeWidget(self):
         self.asSubItem = True
         self.hideWidget = False
         opts = self.param.opts
-        w = ItemSelect_pb()
+        
+        if 'checkbox' in opts and opts['checkbox']:      
+            w = ItemSelect_pb(checkbox = opts['checkbox'])
+            w.sigChanged = w.itemselect.itemChanged
```
seb5g, line 104: I guess this is taking care of my question above, but still need to be put in tests to ensure the behaviour is what we expect

seb5g (changes requested):
some typos :-|
tests are missing :-(
but it looks better now ;-)

src/pymodaq/examples/parameter_ex.py
Ashwolaa: I would again argue that it makes the item more general.
drag/drop cannot really hurt as long as it is internal and is used in a simple qlistwidget

Ashwolaa: With the checkable items, it allows you to change the order of actuators that you desire

src/pymodaq/utils/parameter/pymodaq_ptypes/itemselect.py
Ashwolaa: It does nothing when you click, the selection is decided with the item.checkState() which is just looking for the checkbox value for each item
0 == unchecked
1 == somewhat check (used with tristate)
2 == check

This function is as simple as **self.all_items()**  and is built similarly as **allitems = [item.text() for item in self.all_items()]** 
 A test looks a bit overkilled honestly, the parameter_ex.py is already testing all of what you are asking.

Where would you like to put it?

Ashwolaa: > When you check the box, is the item put in the selected list?

Yes the ordering depends on the initial order as it goes through all the list from up to down

Ashwolaa: > What happens when you click on the item (as we did previously)? Does the check box becomes checked? Or is it possible to be selected with a checkbox not checked?

- It does nothing, the signal is related to the itemChanged now which will only be modified if the text or the checkbox is changed 
- It is not possible to select without checking the checkbox, only ticking the checkbox will emit a signal

seb5g: > A test looks a bit overkilled honestly, the parameter_ex.py is already testing all of what you are asking.

Well except no one is going to run the example when commiting changes of behaviour (such as what you do right now). This is why programmatic tests are needed, I would put them here:
image

And we have to make sure that both objects behave exactly the same (in terms of selection and ordering) otherwise the behaviour would change is we go from using check boxes to not using them

seb5g (changes requested):

src/pymodaq/examples/parameter_ex.py
seb5g: with the checkable items, the order of the items where you clicked is not setting the order?

src/pymodaq/utils/parameter/pymodaq_ptypes/itemselect.py
```diff
@@ -71,28 +82,35 @@ def set_value(self, values):
...
-            self.clear()
-            for value in values['all_items']:
-                item = QtWidgets.QListWidgetItem(value)
-                if self.hasCheckbox:
+        allitems = []
+        # Check existing items
+        for item in self.all_items():     
+            if item.text() not in values['all_items']: # Remove items from list if text not in values
+                item = self.takeItem(self.row(item))
+                del item # Qt recommand to delete the item when removed
```
seb5g, line 90: we are in the python world so the unused things are garbage collected (no use to delete them manually)

src/pymodaq/utils/parameter/pymodaq_ptypes/itemselect.py
```diff
@@ -122,14 +140,21 @@ def makeWidget(self):
             w.add_pb.setVisible(opts['show_pb'])
         else:
             w.add_pb.setVisible(False)
+            
+        if 'show_mb' in opts:
+            w.remove_pb.setVisible(opts['show_mb'])
```
seb5g, line 145: you can do shorter using:
```
w.remove_pb.setVisible(opts.get('show_mb', False))
```

seb5g (changes requested):
If this Parameter get more and more features, it is of paramount importance to test its proper functionnalities.

By writing tests for you PR on settings (save/load/update) I noticed quite a few bugs related to some particular items. So I made sure I tested your methods for comparison and for loading/saving... for each type of parameter in the parameter_ex. So you'll have to do this also to test your new functionnalities here  in this Parameter

Ashwolaa:
Merged pymodaq-dev branch to get settings

Ashwolaa:
> If this Parameter get more and more features, it is of paramount importance to test its proper functionnalities.
> 
> By writing tests for you PR on settings (save/load/update) I noticed quite a few bugs related to some particular items. So I made sure I tested your methods for comparison and for loading/saving... for each type of parameter in the parameter_ex. So you'll have to do this also to test your new functionnalities here in this Parameter

I created a new file for the testing which is based on the Parameter_ex.py

I guess the test should essentially assert that the setValue and the getValue correctly give the correct output onto the GUI or from the GUI respectively.
Would you know how you can access the QListwidget from the ParameterManager?

Ashwolaa:
I added the order of click with the checkbox to match the initial behavior. I just do not really find a good approach on how to build a test of this feature.

I add a video of the feature with the scanner function 

https://github.com/PyMoDAQ/PyMoDAQ/assets/51032965/29af7acc-11cc-49f6-aae4-68e24888d073

seb5g:
You can access the widget using something like:

```
tree = ParameterTree()
tree.setParameters(settings_old, showTop=False)

dict_item, _ = find_objects_in_list_from_attr_name_val(tree.listAllItems(), 'param',
                                                     settings_old.child('main_settings', 'axis'))
dict_widget: QtWidgets.QComboBox = dict_item.widget.combo
```

which is taken from a test in :
image

PS: your video is corrupt and cannot be watched

quantumm:
The video is working for me.

seb5g:
it's actually working on chromium. If you could translate your actions in the video in a set of tests that would be perfect!

seb5g:
You'll let me know when this is ready for review?

Ashwolaa:
Should be good hopefully, by the way I added the selection with double click when using checkbox. The selection is particularly easy with it.

seb5g (approved):

PR #215, opened 2023-11-16
Update write_documentation.rst

jerlfan:
Just adding a preicsion about the make html procedure with windows.


seb5g (approved):
Indeed this is required with powershell. I didn't know people were using it ;-)

PR #220, opened 2023-11-19
Save scanner settings

seb5g:
this PR allows to save the scanner settings set by a user for any kind of combination between:

* choice of actuators
* choice of scan type
* choice of scan subtype

these are saved in a local toml file : scanner_settings.toml in the user's .pymodaq config folder. Initial value of the scanner settings can also be modified there or in the system wide scanner_settings.toml file


seb5g:
this saving mechanism is automatic. The user feeling is that the last settings he applied on a particular configuration (actuator+scan_type+scan_sub_type) will be recalled if he selects again the same (actuator+scan_type+scan_sub_type)

seb5g:
@rgeneaux and @Ashwolaa  could you test these functionalities from that branch and let me know if it is satisfactory? No need to review the detail of the code, but only that the behaviour is nice and as expected!
cheers
Seb

Ashwolaa:
I have not tested everything but it feels weird that it can record more than one/two actuators for 1D/2D scans in the scanner_settings.toml

image

Similar story here, it creates entry with 1 actuator for 2D scans
image

I guess it is related to the way signals are emitted and files get selected. This usually happen when selecting more or less than 2 actuators for 2D scans and similarly for 1D scan.


The 2D spiral do not get properly saved according to the actuators selected.

seb5g:
Yes it is weird but just an extra unused saving as the scan won't start if the actuators number is not correct. Its just that its updating the file everytime one single change is applied. Maybe a better way would be to save the settings only if the user started a scan with them? What do you think?

Ashwolaa:
> Yes it is weird but just an extra unused saving as the scan won't start if the actuators number is not correct. Its just that its updating the file everytime one single change is applied. Maybe a better way would be to save the settings only if the user started a scan with them? What do you think?

Yes I agree, you would like to reuse your last scans so only need to save it when launching the scan or by connecting it to the set_scan/calculate_position button

seb5g:
Alright the config quadruplet(acturators/scan type/subtype/settings) is now only saved when the user press set_scan (either on the Scanner object or on the DAQ_Scan). It automatically load the right config if it exists in the config file. I suggest you guys to test these from running the pymodaq.utils.scanner.scanner module that shows the Scanner object and a simplified UI compared to the DAQScan.

PR #225, opened 2023-11-24
Aspect ratio on Viewer2D

Ashwolaa:
The aspect ratio on Viewer2D is initially set to be 1:1 but the button is not correctly toggled.
This patch adds the toggling of the button such that the interface is consistent with the initial definition of the imageWidget (see setupUI in ImageWidget.py)

Idea: Add an extra keyword to the add_action function such as  "isChecked/isToggled"?


seb5g:
> Idea: Add an extra keyword to the add_action function such as "isChecked/isToggled"?

Yes why not, I would prefer the `checked` keyword, do you want to implement it?

Ashwolaa:
sure, I'll do that when I have a bit of spare time
Envoyé : mardi 28 novembre 2023 09:34
Cc : Ashwolaa ***@***.***>; Author ***@***.***>
Objet : Re: [PyMoDAQ/PyMoDAQ] Aspect ratio on Viewer2D (PR #225)


Idea: Add an extra keyword to the add_action function such as "isChecked/isToggled"?

Yes why not, I would prefer the checked keyword, do you want to implement it?

PR #226, opened 2023-11-28
ROI 

Ashwolaa:
I detected a few errors in the roi_manager.

- The center definition was not properly working when giving an angle to the ROI.
This is now fixed by using a new function rotate2D and a redefinition of the center function
- The setAngle was not defined to rotate the ROI with respect to its center. I think it is much more intuitive for the user if this is the case.
This is now fixed by adding the argument center = [0.5,0.5] in setAngle

Ideas: Many functions are shared between ellipse/rect roi. It could be worth to add a baseclass for it.

### DIFF
--- a/src/pymodaq/utils/managers/roi_manager.py
+++ b/src/pymodaq/utils/managers/roi_manager.py
@@ -441,8 +441,6 @@ class ROIManager(QObject):
                     else:
                         newroi = EllipseROI(index=newindex, pos=pos,
                                             size=[width, height])
-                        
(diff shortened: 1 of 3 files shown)


Ashwolaa (commented):
I also changed the initial size that takes the ROI when created as it currently covers the full screen in the 1D case which is not user friendly as it becomes just hard to grab the edges. I put 1/3-2/3 of the screen as a guess.

seb5g (approved):

Issue #224, opened 2023-11-22
Indexing of scan while taking negative steps


Khodeo:
Hi

I don't know if this is of concern, but I noticed that in pymodaq4 while taking the 1D and 2D scans the indexing given by

self.scanner.get_indexes_from_scan_index(self.ind_scan)

in the daq_scan go from 0 to N if the step of the actuator is positive, while go backward from N to 0 if the step of the actuator is negative. In the final h5 file the nav_axis is instead saved always in the correct ordering, as is defined by the user. 

I think this is because in the scan_factory the function get_info_from_positions is using a numpy.unique(ax) that is also sorting ax. If the convention on the indexing going from 0 to N is to be mantained both for positive and negative steps, I think this could be substituted with something like

ax_indexes = np.unique(ax, return_index=True)[1]
ax=[ax[index] for index in sorted(ax_indexes)]
axes_unique.append(np.array(ax)) 

in order to mantain the ordering of the user defined axis while doing the unique.

seb5g:
Indeed in some cases the indexing with respect to the navigation axis is reversed, I'll look at PR230 for this. I guess we need some testing to make sure the indexing is always correct with respect to the saved axes (the nav one especially because set by the scanner)

PR #230, opened 2023-12-06
Pull request for issue  #224

Ashwolaa:
From scan_factory.py
self.axes_indexes was built with the unique function in  get_info_from_positions and led to a sorting of the value.
This made it not reliable when doing backwards scan as the indexes given did not match the position array which was unsorted)

This fixes the issue #224 by keeping the iniital sorting given by the user

### DIFF
--- a/src/pymodaq/utils/scanner/scan_factory.py
+++ b/src/pymodaq/utils/scanner/scan_factory.py
@@ -127,7 +127,7 @@ class ScannerBase(ScanParameterManager, metaclass=ABCMeta):
                 positions = np.expand_dims(positions, 1)
             axes_unique = []
             for ax in positions.T:
-                axes_unique.append(np.unique(ax))
+                axes_unique.append(ax[np.sort(np.unique(ax, return_index=True)[1])])                
             axes_indexes = np.zeros_like(positions, dtype=int)
             for ind in range(positions.shape[0]):
                 for ind_pos, pos in enumerate(positions[ind]):
--- a/src/pymodaq/utils/scanner/scan_factory.py
+++ b/src/pymodaq/utils/scanner/scan_factory.py
@@ -127,7 +127,7 @@ class ScannerBase(ScanParameterManager, metaclass=ABCMeta):
                 positions = np.expand_dims(positions, 1)
             axes_unique = []
             for ax in positions.T:
-                ax = np.sort(np.unique(ax, return_index=True)[1]) # Getting unique values in ax without any sorting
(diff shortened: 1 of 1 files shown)


seb5g:
image

Ashwolaa:
Sorry, I am not sure to totally follow what approach you prefer and what tests are asked

The scanner gives:
**Before:**
position: initial definition from user (keep ordering)
axes_unique: unique + sorted version of position, the function get_nav_axes() outputs an Axis with this variable as data
axes_indexes: indexes relating the positions to the axes_unique array, these indexes are then used when storing data

**Within this pull request:**
position: initial definition from user (keep ordering) #_UNCHANGED_
axes_unique: unique + unsorted version of position, the function get_nav_axes() still outputs an Axis with this variable as data # _CHANGED: It is now unsorted in the same order as position_ 
axes_indexes: indexes relating the position to the axes_unique array, these indexes are then used when storing data # _CHANGED: This becomes a  "useless" variable when we do not repeat the same points as the indexes just follows the evolution of the position array so for a simple 2D scan [0,0 ; 0,1 ; 0,2 ; 1,0 ; 1,1 ; 1,2 ]_

In fact, I believe the best way to handle things is to not sort anything as the order in which the data are recorded can be of particular importance when having to deal with instabilities in an experiment. The user can then sort the data when he needs too and this option can be instead be added in the h5browser if needed

seb5g:
Sorting may have its importance for proper plotting of the data... But I think the data were never sorted into the h5 file except for the 2D uniform grid cases (linear, back and forth and spiral) because then sorting them and placing them into a regular grid allows to plot the data as an image, otherwise if would be done using the triangulation method which for uniform grid data is ugly. 

So beware that because of the overall generality of PyMoDAq, what you think make sense could be more tricky than at first glance!

And indeed by testing your PR and trying to do a spiral scan, I got errors for the plotting. axes_indexes is the index in the sorted version of the axes (especially because of what I wrote above). 

Anyhow, I didn't get how it was supposed to create an error at the beginning of the issue #224 , could you expose more precisely how to reproduce the error you're seeing?

Ashwolaa:
Minimal example with daq_scan + h5browser

- Launch daq_scan
- pick Det 2D
- pick XAxis as actuator
- With 1D Linear scan
Xaxis start: 50 / Xaxis stop: -50 / Xaxis step: -50
- Launch scan
- Open output file with h5browser
By scrolling at Data00, we can see that 50 is positionned at -50 and vice versa so indexes have been flipped somehow.

This example is to see it with the h5browser but this was first observed on experimental data directly

PR #236, opened 2024-01-10
small bugfixes

rgeneaux:


seb5g:
can you request a review and a reviewer when you're done?

rgeneaux:
unfortunately I don't think anyone but you has the rights to do that!
no option to do so here:

image

seb5g (approved):

PR #237, opened 2024-01-11
Display patch for 2D linear scan

Ashwolaa:
- "Better" display for 2D linear (stages are separated with children for start/stop/step)
- Same convention as 1D linear start/stop/step (before start/step/stop)
image


seb5g:
yes it's nicer, could you do the same grouping for the spiral one also?
Cheers

PR #239, opened 2024-01-11
Fix "use PID as actuator" in preset

rgeneaux:
Hi,
In version 4, the option to check "use pid as actuator" in a preset was not working yet. Loading the PID module directly with the preset bugged, and so did the initialization of the "fake" associated actuator.

This is fixed by moving the PID module initialization after the dashboard has a properly defined self.modules_manager (which is accessible to the DAQ_PID afterwards). Then the fake actuator is created, after which we update the dashboard's self.modules_manager to add it. The utility plugin (daq_move_PID) also is now added to the list of available plugins (done in get_instrument_plugins()). 
 
Tested using mock detectors and actuators together with one of our PID models https://github.com/Attolab/pymodaq_plugins_PIDStabFringes/blob/main/src/pymodaq_plugins_pid/models/PIDSpatialInterferometrySE1bis.py

### DIFF
--- a/src/pymodaq/dashboard.py
+++ b/src/pymodaq/dashboard.py
@@ -673,41 +673,6 @@ class DashBoard(QObject):
             det_docks_viewer = []
             move_forms = []
 
-            # # set PID if checked in managers
-            try:
-                if self.preset_manager.preset_params.child('use_pid').value():
-                    self.load_pid_module()
-
-                    self.pid_module.settings.child('models', 'model_class').setValue(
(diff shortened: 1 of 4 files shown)


seb5g:
could you test is also with the beeamsteering model which is the benchmarking so far? 

You'll find the MockCamera actuators and detectors in the pymodaq_plugins_mockexamples plugin that also defines the related PID model

and then I'll review it also (you have to notify someone, see top right)

seb5g:
I did test it with the beam steering and it works fine!!

PR #240, opened 2024-01-12
Bugfix/spread data nav axis

### LINKED ISSUE #224: Indexing of scan while taking negative steps

seb5g:
Provide a better comprehension of the axis indexing provided with the scanners definition 

Should Close #224

by properly defining the content of the subclassed method `get_indexes_from_scan_index`

See below for instance how redefining as such for 1D scan make the indexing coherent with the scan definition. Only 2Dscan: back and forth and spiral are defined differently to display the scan in a viewer2D and not in a spread ViewerND

```
class Scan1DBase(ScannerBase):
    params = []
    n_axes = 1
    distribution = DataDistribution['uniform']

    def __init__(self, actuators: List = None, **_ignored):
        super().__init__(actuators=actuators)

    def get_nav_axes(self) -> List[Axis]:
        return [Axis(label=f'{self.actuators[0].title}',
                     units=f'{self.actuators[0].units}',
                     data=np.squeeze(self.positions))]

    def get_scan_shape(self) -> Tuple[int]:
        return len(self.positions),

    def get_indexes_from_scan_index(self, scan_index: int) -> Tuple[int]:
        """To be reimplemented. Calculations of indexes within the scan"""
        return (scan_index,)
```

As for the sorting. I agree that this should be done separately, for instance on plotting. This is why this PR introduces a new action button on the Viewer1D. This feature can be easily tested from the new mockexamples instrument plugin: daq_1Dviewer_MockRandom, which generate a noisy gaussian with an axis being randomly generated in the gaussian range. Tis gives that:

image

and by clicking on the sort button, this gives that:

image

The correlation is that I removed the sort 1D option in the general DAQ_Scan settings.

Thanks for testing this


Ashwolaa (approved):
- Test for linear case, 1D and 2D : Looks good :+1: 
- Test with Tabular with custom entries (see below): Nax axes spit [0 0 0] for both axes. It seems unrelated to the current change though I'll raise it as an issue
image

Ashwolaa:
Test MockRandom with sorting:
Overlay does not act on the sorted values but on the unsorted ones.

PR #261, opened 2024-02-06
Bug/input arguments h5browser

### LINKED ISSUE #260: file name / path as command line parameter for h5browser?
When launching the h5browser, a file-open dialog pops up pointing to the root directory. Depending on how many files one wants to inspect and how deep the tree is, it becomes quickly rather tedious to brachiate each time through half the forest. Therefore I added a few lines to browsing.py and h5browser.py so that the command python -m pymodaq.extensions.h5browser my-data-file.h5 will open that file directly if it exists or come up with the file dialog pointing to the corresponding path if the latter exists, otherwise the longest valid parent path. Questions:

a) Would that be of more general interest? Should I launch a merge request to the dev branch after having done some testing?

b) If yes, "Open if existing or else ask what to open" with sensible default path options sounds like a rather generic task. Could be useful in various other places. Should that rather go to some utils module?

c) Related to working on that: after having forked PyMoDAQ, cloned it locally and checked out 4.2.x_dev, resources/VERSION contains still 4.1. is that intentional?

d) When starting the dashboard in the virtual environment where the 4.2.x_dev branch is installed as editable, a message pops up that a new version of PyMoDAQ would be available. When accepting to upgrade, the editable version 4.2.x_dev gets replaced by 4.1.1 (the git repository still being there untouched but dismissed for python in favour of the newly installed 4.1.1 in the site-packages).

e) Should I raise the latter two items elsewhere?

seb5g:
This branch recreate a previous commit from v3 that didn't go into v4 from @rgeneaux 

Allows adding input argument when callling h5browser over the command line. Doc also highlight how to specify the h5browser to open the h5 file by default!

### DIFF
--- a/docs/src/extensions_folder/h5browser.rst
+++ b/docs/src/extensions_folder/h5browser.rst
@@ -4,7 +4,10 @@
 H5Browser
 =========
 
-The h5 browser is an object that helps browsing of datas and metadatas. It asks you to select a h5 file
+Exploring data
+++++++++++++++
+
+The h5 browser is an object that helps browsing of data and metadata. It asks you to select a h5 file
 and then display a window such as :numref:`figure_h5browser`. Depending the element of the file you are
 selecting in the h5 file tree, various metadata can be displayed, such as *scan settings* or
 *module settings* at the time of saving. When double clicking on data type entries in the tree, the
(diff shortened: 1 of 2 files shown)


docs/src/extensions_folder/h5browser.rst
```diff
@@ -39,3 +42,35 @@ Some options are available when right clicking on a node, see :numref:`figure_h5
...
+how to do it on Windows. Let us assume that you have a conda environment named *my_env*, in which PyMoDAQ is installed.
+
+In Windows, the path to your conda executable will be something like:
+
+``C:\Miniconda3\condabin\conda.bat``
+
+and the path to the H5Browser will be (note that we are going into the *my_env* folder):
+
+``C:\Miniconda3\envs\my_env\Lib\site-packages\pymodaq\h5browser.py``
+
```
rgeneaux: Not necessary anymore (since you use h5browser script directly)

docs/src/extensions_folder/h5browser.rst
```diff
@@ -39,3 +42,35 @@ Some options are available when right clicking on a node, see :numref:`figure_h5
...
+Now that you have written down these two paths, open your favorite text editing tool (e.g. notepad) and create a file
+called *H5Opener.bat* (for instance) with the following contents:
+
+  .. code-block:: python
+
+    @ECHO OFF
+    call C:\Miniconda\condabin\conda.bat activate my_env
```
rgeneaux: missing the 3 in "Miniconda3" compared to above (sneaky!!)
perhaps add a note saying to precisely check the path

rgeneaux (changes requested):
code works well! see comments about documentation

docs/src/extensions_folder/h5browser.rst
seb5g: done

docs/src/extensions_folder/h5browser.rst
seb5g: done

rgeneaux (approved):

PR #262, opened 2024-02-06
Hotfix/tabular scanner

### LINKED ISSUE #241: bug/scanner/tabular:  Positions are not being updated
Testing #240 with Tabular, I realized that the navAxes was not correctly updating
As a test example, with this input and pressing Calculate positions,
image

The output in the terminal gives:
image

seb5g:
patch issue #241 

The issue was that the model was updated too much. Therefore the tableview was linked to a model (updating its values) while the calculate positions was linked to another model with values always equal to init ones that is 0.0


Ashwolaa (approved):
It now works as intended!

PR #264, opened 2024-02-07
Leco test

BenediktBurger:
This PR is only there, to see, what changes are implemented, even in other LECO files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,6 +104,9 @@ venv.bak/
 .spyderproject
 .spyproject
 
+# VS code project settings
+.vscode/*
+
 # Rope project settings
 .ropeproject
 
--- a/src/pymodaq/control_modules/daq_move.py
+++ b/src/pymodaq/control_modules/daq_move.py
@@ -37,10 +37,8 @@ from pymodaq.utils.h5modules import module_saving
 from pymodaq.utils.data import DataRaw, DataToExport, DataFromPlugins, DataActuator
 from pymodaq.utils.h5modules.backends import Node
 
-try:
-    from pymodaq_plugins_thg.hardware.leco_client import PymodaqListener
-except ModuleNotFoundError:
-    PymodaqListener = False
+from pymodaq.utils.leco.leco_client import PymodaqListener
+
 
 local_path = config_mod.get_set_local_dir()
 sys.path.append(local_path)
@@ -612,7 +610,7 @@ class DAQ_Move(ParameterManager, ControlModule):
             self._tcpclient_thread.start()
 
     def connect_leco(self):
-        if self.settings.child("main_settings", "leco", "connect_leco_server").value() and PymodaqListener is not False:
+        if self.settings.child("main_settings", "leco", "connect_leco_server").value():
             self._leco_client = PymodaqListener(name=self.settings.child('main_settings', 'leco', 'name').value(),
                                                 server_name=self.settings.child('main_settings', 'leco', 'server_name').value(),
                                                 )
--- a/src/pymodaq/control_modules/daq_move.py
+++ b/src/pymodaq/control_modules/daq_move.py
@@ -611,8 +611,10 @@ class DAQ_Move(ParameterManager, ControlModule):
 
     def connect_leco(self):
         if self.settings.child("main_settings", "leco", "connect_leco_server").value():
-            self._leco_client = PymodaqListener(name=self.settings.child('main_settings', 'leco', 'name').value(),
-                                                server_name=self.settings.child('main_settings', 'leco', 'server_name').value(),
+            name = self.settings.child("main_settings", "module_name").value()
+            director_name = self.settings.child('main_settings', 'leco', 'director_name').value()
+            self._leco_client = PymodaqListener(name=name,
+                                                remote_name=director_name,
                                                 )
             self._leco_client.cmd_signal.connect(self.process_tcpip_cmds)
             self._command_tcpip[ThreadCommand].connect(self._leco_client.queue_command)
--- a/src/pymodaq/control_modules/daq_move.py
+++ b/src/pymodaq/control_modules/daq_move.py
@@ -37,7 +37,7 @@ from pymodaq.utils.h5modules import module_saving
 from pymodaq.utils.data import DataRaw, DataToExport, DataFromPlugins, DataActuator
 from pymodaq.utils.h5modules.backends import Node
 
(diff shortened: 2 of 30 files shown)


BenediktBurger:
## Content explanation

- The daq_move, daq_viewer (and their utility classes) have been modified to offer the option to enable leco alongside TCP
  Whenever you activate that leco function, they will use the `PymodaqListener` from `leco_client.py` to listen for incoming messages. (for details see below)
- The `PymodaqListener` listens for incoming messages (calling methods), which will emit signals (with ThreadCommand), such that the Module will behave accordingly.
  It has also a method, which can be connected to a tcp Signal, which will react to changes in the Module (e.g. settings changed, position reached, data ready etc.). These changes will cause the Listener to call a method on the dashboard module (you have to tell the Listener the name of that module via the `set_remote_name` method).
- `director_utils` contains pyleco Directors (small utilities for facilitating calling remote methods), which correspond to the PymodaqListener:
  Each method of the PymodaqListener, which can be called via LECO, is also present in these Directors. For example `Director.set_remote_name` will call `PymodaqListener.set_remote_name`.
  That makes it simpler to program, as only the Director and the Listener have to match each other. Everyone else uses the Director and does not have to worry about str names of methods.
- daq_move_LECODirector.py and leco_viewerLECOServer.py (I had to choose these names, as the modules did not want to start otherwise) correspond to the TCPServer modules.
  These are pymodaq modules which can control some other module.
  They use the directors from `director_utils`.
  These could be in their own plugins package, but for easier development (as they match the ActorModules), they are included here.

## Goal

### Final vision

1. You start a Coordinator (executing `pyleco.coordinators.coordinator.py`).
2. You start any Move/Viewer module (we call it here ActorModule)
   1. with some user given name `"my_actor_module"`
   2. You activate the `connect_leco` option in the main settings
3. You start the corresponding LECOServer / LECODirector module
   1. You set `actor_name="my_actor_module"` in this server module
   2. You connect / initialize this module (how do you do it?)
4. Changes to the DirectorModule will reflect in the ActorModule and vice versa.
5. The DirectorModule shows data from the ActorModule

### Next steps

Make it work

## Current state

- The mock move module works as expected (first test) using the pymodaq listener
- I was not able to start the DirectorModules (LECOServer) connecting to LECO, probably I miss something.
- Naming of files, classes, options in the menu are open for discussion.

BenediktBurger:
I made the move_LECODirector to get the values of the mockactor, but the control fields are deactivated, I have to look into it.

You have to move `move_LECODirector` in some move plugins folder.

BenediktBurger:
Note to myself:
Another message type could indicate one json frame and one bytes frame.
In this case the MessageHandler will store the message in a temporary variable before calling "rpc" thing
Afterwards the temporary variable is deleted.

The method can then transmit that information to the module.

BenediktBurger:
Ideas:

- [ ] Add an option to the main_settings/leco/ options to specify the name in the leco network (if it should differ from the module name)?
- [ ] Transfer the objects via base64 instead of in an additional frame?
- [ ] Change settings remotely: Does not work yet (did it work with TCP?)

BenediktBurger:
I switched from special message type to pure JSON messages (which makes the handling simpler).
In order to send the pymodaq objects via JSON, I had to use base64 bytest-to-text encoding (quite simple to add to the Serializer).

seb5g:
Hi,
I started to look at your work but there is a missing object in pymodaq.utils/leco/pymodaq_listener  : PymodaqCommunicator It is imported by the leco_director.py module but I could not find where it could be??

Concerning the points below

> ## Goal
> ### Final vision
> 
>     1. You start a Coordinator (executing `pyleco.coordinators.coordinator.py`).
that's fine
> 
>     2. You start any Move/Viewer module (we call it here ActorModule)
>        
>        1. with some user given name `"my_actor_module"`
>        2. You activate the `connect_leco` option in the main settings
works just fine
> 
>     3. You start the corresponding LECOServer / LECODirector module
>        
>        1. You set `actor_name="my_actor_module"` in this server module
>        2. You connect / initialize this module (how do you do it?)
We'll have to move these into the pymodaq_plugins_mock plugin (The TCP version of these are stored there) and because this plugin is always installed by default, those leco directors will also be listed and you will be able to initialize them (once you've set the correct associated actor name)
> 
>     4. Changes to the DirectorModule will reflect in the ActorModule and vice versa.
could not test this because of what I said in the above
> 
>     5. The DirectorModule shows data from the ActorModule
could not test 

Thanks for letting me know where is the PymodaqCommunicator

seb5g:
> I switched from special message type to pure JSON messages (which makes the handling simpler). In order to send the pymodaq objects via JSON, I had to use base64 bytest-to-text encoding (quite simple to add to the Serializer).

Do you think it make the "data" heavier? I especially think of  heavy data such as 2D or ND with multiple axes?

BenediktBurger:
> Do you think it make the "data" heavier? I especially think of heavy data such as 2D or ND with multiple axes?

Yes, the data increases by 33% and conversion will require computation time. I kept a branch with that code.
However, it makes the code more simple.

> PymodaqCommunicator

That was needed for the special handling.
It seems, that I forgot to remove it from some imports, when I switched to base64.
I'm sorry, I'll fix it.

seb5g:
I tested it just now, it works just fine!!

I modified a few things in the fork I created from your own fork... I patched the LECODirector class for its naming as 'main_settings' is not available in the instrument plugin side. I managed to init 2 directors controlling one real daq_viewer (Mock) :-D

About these modifications, how do we proceed (because fork of fork is maybe not the best scenario...)

BenediktBurger:
> About these modifications, how do we proceed (because fork of fork is maybe not the best scenario...)

You could
- create a new development branch in the PyMoDAQ repo
- We do development with PRs versus this "leco"-branch
- Finally, the "leco"-branch will be merged into the main development branch.

Alternatively, we just do it in the main development branch.

BenediktBurger:
Content is merged in PR #267 into the development branch [feature/pyleco](https://github.com/PyMoDAQ/PyMoDAQ/tree/feature/pyleco).
There development will happen.

seb5g:
Alright I created the PyMoDAQ/ feature/pyleco branch based on 4.2.x_dev and this current PR

seb5g:
the directors plugins are now on : https://github.com/PyMoDAQ/pymodaq_plugins_mock/tree/feature/pyleco-directors

PR #266, opened 2024-02-15
Features/data ploting interactive

seb5g:
This PR adds a plot method to data object.

This allows for a very quick way to plot data either using the matplotlib backend or the qt one (with PyMoDAQ  DataViewers). 

Other backends can be added from the use of a Factory Pattern

To illustrate those and some other convenience methods introduced along the way, this PR also propose a new tutorial on data loading, manipulation and plotting!


seb5g:
@Ashwolaa or @rgeneaux or @quantumm  could you review this PR and check both the code and the tutorial?

PR #268, opened 2024-02-21
Leco name and rework

BenediktBurger:
I added the option to give a name to the ActorModules (just called `name`).
If that name is not given, it will take the name of the module.

You can change that name at any time (even while connected to the leco network: it will sign out, change the name and sign in again).

I also created a superclass for both Modules, such that some logic can be shared and does not have to be written twice.

As an example, how to use enums for ThreadCommands, I added a few ones for LECO


seb5g:
thanks for the refactorization

PR #272, opened 2024-02-22
remove pyleco dependency

seb5g:
pyleco is only compatible with python >=3.9 so the python3.8 CI tests failed because of the added dependency...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,6 @@ dependencies = [
     "toml",
     "qtconsole",
     "tables<3.9",  # issue with some version of required package blosc2>=2.2.8
-    "pyleco",
 ]
 
 [project.scripts]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,6 +51,7 @@ dependencies = [
     "toml",
     "qtconsole",
     "tables<3.9",  # issue with some version of required package blosc2>=2.2.8
+    "pyleco; python_version>\"3.8\"",
 ]
 
 [project.scripts]


seb5g:
@BenediktBurger just removing the dependency is not enough has then all CI fails when trying to import it in the tests. Do you have an idea on how to not use pyleco with python<3.9 but still make our CI work?

seb5g:
I added a condition on the python version but still the CI fails for python ==3.8

seb5g:
and I also have this error:

```
 from pymodaq.utils.leco.utils import PymodaqMessage, PYMODAQ_MESSAGE_TYPE
E   ImportError: cannot import name 'PymodaqMessage' from 'pymodaq.utils.leco.utils' (/home/runner/work/PyMoDAQ/PyMoDAQ/src/pymodaq/utils/leco/utils.py)
=============================== warnings summary ===============================
```

BenediktBurger:
> and I also have this error:

Interesting, I cannot find "PymodaqMessage" via grep in the source code. Maybe a caching issue?

EDIT: I found it in the tests (I was looking only in the src)

BenediktBurger:
#273 removes the culprit

I forgot to adjust the test.

What about running the tests on each push/pull request opened?

seb5g:
> What about running the tests on each push/pull request opened?

I just saw it is not the case for you! I'll check why because when I do a PR or a push, it runs them

BenediktBurger:
maybe pushes/PRs to the main branch?

That's what we have:
~~~
on:
  push:
    branches:
      - main
  pull_request:
~~~

You have just `on push`

seb5g:
I just did a PR to add the pull_request trigger (there was only push before)

PR #276, opened 2024-02-23
Fix typing in serialize_object

BenediktBurger:
It seems, that `isinstance(obj, Union)` does not work in Python<3.10.

`typing.get_args` gets the simplified arguments and should work as an argument for `isinstance`.

### DIFF
--- a/src/pymodaq/utils/leco/utils.py
+++ b/src/pymodaq/utils/leco/utils.py
@@ -1,6 +1,6 @@
 import subprocess
 import sys
-from typing import Any, Union
+from typing import Any, Union, get_args
 
 # import also the DeSerializer for easier imports in dependents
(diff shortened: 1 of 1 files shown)


BenediktBurger:
Python 3.8 fails, as pyleco is not available.

seb5g (approved):

seb5g:
is pyleco not working with python<3.9 only related to typing issues?

BenediktBurger:
> is pyleco not working with python<3.9 only related to typing issues?

This PR fixed the test for Python 3.9.

The depency we use to create the JSON-RPC messages etc. is not available vor python<3.9, as that dependency uses typings heavily.
I'm working on a compatible version, which circumvents that dependency.

seb5g:
👍

PR #299, opened 2024-04-19
bugfix load_ROI: 'filter' argument is a str, not list of str.

rgeneaux:
"Load ROI" in a Viewer wasn't working because 'filter' option of file_io.select_file needs to be a str, not list of str. 
[QtWidgets.QFileDialog.getOpenFileName](https://doc.qt.io/qtforpython-6/PySide6/QtWidgets/QFileDialog.html#PySide6.QtWidgets.QFileDialog.getOpenFileName)

Corrected and changed docstring of file_io.select_file. 
Didn't find any other call of file_io.select_file with list of str anywhere else so this should be ok!


seb5g (approved):

seb5g:
I'll cherry pick your commit to apply it to 4.1 also

PR #323, opened 2024-06-13
Fix object deserialization

### LINKED ISSUE #320: DeSerializer.object_deserialization does not work
In #318 I noticed, that `Deserializer().object_deserialization()` does fail, while `Deserializer().dwa_deserialization()` works.

I expected `object_deserialization` to be a one-for-all method to deserialize anything (similar to `to_bytes`).

I figured out, why it fails:
https://github.com/PyMoDAQ/PyMoDAQ/blob/9244938e4292b5d06aa02e26b07947fcba6a7ccb/src/pymodaq/utils/tcp_ip/serializer.py#L620-L622

The second line (L621) reads the type, but reading already consumes the bytes from the strings.
Later on the called `deserialization` methods try to read the type again, which fails, as the type has been read and the corresponding bytes are not accessible anymore.

In #318 I propose a fix:
~~~python
    def object_deserialization(self, manual=False):
        """Deserialize an object.

        If you call it manually, you have to set manual to True.
        """
        obj_type = self.string_deserialization()
        hand_over = obj_type if manual else None
        elt = None
        if obj_type == 'scalar':
            elt = self.scalar_deserialization()
        elif obj_type == 'string':
            elt = self.string_deserialization()
        elif obj_type == 'array':
            elt = self.ndarray_deserialization()
        elif obj_type in ('dwa', "DataActuator"):
            elt = self.dwa_deserialization(obj_type=hand_over)
        elif obj_type == "DataToExport":
            elt = self.dte_deserialization(obj_type=hand_over)

    def dte_deserialization(self, obj_type: Optional[str] = None) -> DataToExport:
        """Convert bytes into a DataToExport object

        Convert the first bytes into a DataToExport reading first information about the underlying data

        Returns
        -------
        DataToExport: the decoded DataToExport
        """
        class_name = self.string_deserialization() if obj_type is None else obj_type
~~~

Here, the read obj_type is given to the deserialization method, which does not read the type again.
However, in order to not break code, this possibility is optional and disabled per default.

BenediktBurger:
Closes #320

### DIFF
--- a/src/pymodaq/utils/tcp_ip/serializer.py
+++ b/src/pymodaq/utils/tcp_ip/serializer.py
@@ -6,7 +6,7 @@ Created the 20/10/2023
 """
 from base64 import b64encode, b64decode
 import numbers
-from typing import Tuple, List, Union, TYPE_CHECKING, Iterable
+from typing import Optional, Tuple, List, Union, TYPE_CHECKING
 
 
 import numpy as np
@@ -563,7 +563,7 @@ class DeSerializer:
         str_obj = self._bytes_string.get_first_nbytes(string_len).decode()
         return str_obj
 
-    def scalar_deserialization(self) -> numbers.Number:
+    def scalar_deserialization(self, data_type: Optional[str] = None) -> numbers.Number:
         """Convert bytes into a numbers.Number object
 
         Get first the data type from a string deserialization, then the data length and finally convert this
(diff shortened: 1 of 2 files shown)


seb5g:
for regular types, it is not necessary to specify the type (in fact the subtype). The issue is only for dwa deserialization because this DataWithAxes type exists in various flavor (in particular the DataActuator that is important for the DAQ_Move)

seb5g:
Ok I think I misled you. Our programs should be self aware of what type of object is to be expected. For instance a data_ready signal will convey a dte hence you should use a dte_deserialization method. The same applies for other objects. the method object_deserialization is only meant to be used within dte_deserialization od dwa_deserailization

Does that solves your issues? If yes I will add some docstrings and tests for that (and maybe make the object_deserialization a private method)

seb5g:
to be even more precise:

the method _object_deserialization_ is to be used only where _type_and_object_serialization_ has been used in Serializer, that is in:

- list_serialization: because we don't know what could be in the list
- dwa_serialization: because the field extra_attributes could contains anything

PR #324, opened 2024-06-13
better explanation about usage and test suite for the object_deserialization

### LINKED ISSUE #320: DeSerializer.object_deserialization does not work

seb5g:
Related to #323 and closes #320 

Focus on explaining not to use directly the _object_deserialization_ method but only the real object deserialization (dwa, list, scalar, dte, ...)

### DIFF
--- a/src/pymodaq/utils/parameter/utils.py
+++ b/src/pymodaq/utils/parameter/utils.py
@@ -27,11 +27,11 @@ class ParameterWithPath:
         self._path = path
 
     @property
-    def parameter(self):
+    def parameter(self) -> Parameter:
         return self._parameter
 
     @property
-    def path(self):
+    def path(self) -> List[str]:
         return self._path
 
 
--- a/src/pymodaq/utils/parameter/utils.py
+++ b/src/pymodaq/utils/parameter/utils.py
@@ -1,12 +1,38 @@
 from __future__ import annotations
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, List
 import numpy as np
 from collections import OrderedDict
-
+from dataclasses import dataclass
 from pymodaq.utils.daq_utils import find_keys_from_val
 
 if TYPE_CHECKING:
     from pymodaq.utils.parameter import Parameter
+
+
+class ParameterWithPath:
+    """ holds together a Parameter object and its full path
+
+    To be used when communicating between TCPIP to reconstruct properly the Parameter
+
+    Parameters
+    ----------
+    parameter: Parameter
+        a Parameter object
+    path: full path of the parameter, if None it is constructed from the parameter parents
+    """
+    def __init__(self, parameter: Parameter, path: List[str] = None):
+        self._parameter = parameter
+        if path is None:
+            path = get_param_path(parameter)
+        self._path = path
+
+    @property
+    def parameter(self):
+        return self._parameter
+
+    @property
+    def path(self):
+        return self._path
 
 
 def get_widget_from_tree(parameter_tree, widget_instance):
@@ -18,22 +44,25 @@ def get_widget_from_tree(parameter_tree, widget_instance):
     return widgets
 
 
-def get_param_path(param):
-    """
+def get_param_path(param: Parameter) -> List[str]:
+    """ Get the parameter path from its highest parent down to the given parameter including its
+    identifier (name)
 
     Parameters
     ----------
-    param
+    param: Parameter
+        The parameter object
 
     Returns
     -------
-
+    List[str]: the path as a list of parameter identifiers
     """
     path = [param.name()]
     while param.parent() is not None:
         path.append(param.parent().name())
         param = param.parent()
     return path[::-1]
+
 
 def getOpts(param:Parameter,) -> OrderedDict:
     """Return an OrderedDict with tree structures of all opts for all children of this parameter
--- a/src/pymodaq/utils/tcp_ip/serializer.py
+++ b/src/pymodaq/utils/tcp_ip/serializer.py
@@ -270,7 +270,8 @@ class Serializer:
         return bytes_string
 
     def object_type_serialization(self, obj: Union[Axis, DataToExport, DataWithAxes]) -> bytes:
-        """ Convert an object type into a bytes message as a string together with the info to convert it back
+        """ Convert an object type into a bytes message as a string together with the info to
+        convert it back
 
         Applies to Data object from the pymodaq.utils.data module
         """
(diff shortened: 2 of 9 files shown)


seb5g:
I renamed the methods to make them symmetric

I added bytes, dte and ParameterWithPath objects in the type_and_object_serialization/deserialization

The test suite illustrates this symmetry. Is that enough for pyleco?

src/pymodaq/utils/parameter/utils.py
```diff
@@ -1,14 +1,40 @@
...
+            path = get_param_path(parameter)
+        self._path = path
+
+    @property
+    def parameter(self):
+        return self._parameter
+
+    @property
+    def path(self):
+        return self._path
```
BenediktBurger: If you use type hints for input, why not also for output?
```suggestion
    @property
    def parameter(self) -> Parameter:
        return self._parameter

    @property
    def path(self) -> List[str]:
        return self._path
```

src/pymodaq/utils/tcp_ip/serializer.py
```diff
@@ -617,23 +640,36 @@ def ndarray_deserialization(self) -> np.ndarray:
         ndarray = np.atleast_1d(ndarray)  # remove singleton dimensions
         return ndarray
 
-    def object_deserialization(self):
+    def type_and_object_deserialization(self):
+        """ Deserialize specific objects from their binary representation.
```
BenediktBurger: I think it is good to these methods are invers functions, for example:

```suggestion
        """ Deserialize specific objects from their binary representation (inverse of `Serializer.type_and_object_serialization`).
```

Obviously a similar statement could be above and a note in the documentation, that these functions are the entry/exit points of serialization.

tests/utils/tcp_ip/serializer_test.py
```diff
@@ -231,3 +233,95 @@ def test_base_64_de_serialization(get_data: DataToExport):
...
+        serialized = b'\x00\x00\x00\x06string\x00\x00\x00\x0bhello world'
+
+        assert Serializer().type_and_object_serialization(obj) == serialized
+        assert DeSerializer(serialized).type_and_object_deserialization() == obj
+
+    def test_bytes(self):
+        obj = b'hello binary world'
+        serialized = b'\x00\x00\x00\x05bytes\x00\x00\x00\x12hello binary world'
+        assert Serializer().type_and_object_serialization(obj) == serialized
+        assert DeSerializer(serialized).type_and_object_deserialization() == obj
```
BenediktBurger: Why don't you parametrize the tests, as they are always the same, only the parameters change?
```suggestion
    @pytest.mark.parametrize("obj, serialized", (
        (True, b'\x00\x00\x00\x06scalar\x00\x00\x00\x03|b1\x00\x00\x00\x01\x01'),
        (10.45, b'\x00\x00\x00\x06scalar\x00\x00\x00\x03<f8\x00\x00\x00\x08fffff\xe6$@'),
        ('hello world', b'\x00\x00\x00\x06string\x00\x00\x00\x0bhello world'),
        (b'hello binary world', b'\x00\x00\x00\x05bytes\x00\x00\x00\x12hello binary world'),
    ))
    def test_serialization(self, obj, serialized):
        assert Serializer().type_and_object_serialization(obj) == serialized
        assert DeSerializer(serialized).type_and_object_deserialization() == obj
```

If you like to add ids (that you know, which one failed), you could add an `ids` list to the parametrize decorator.

src/pymodaq/utils/tcp_ip/serializer.py
BenediktBurger, line 1: You can add now `Parameter` to the Serializable list (line 20).

Actually, you could use that list as a type hint for the `type_and_object_serialization` method.

tests/utils/tcp_ip/serializer_test.py
```diff
@@ -231,3 +233,95 @@ def test_base_64_de_serialization(get_data: DataToExport):
...
+
+    def test_array(self):
+        obj = np.array([[0.1, 0.5], [5, 7], [8, 9]])
+        serialized = (b'\x00\x00\x00\x05array\x00\x00\x00\x03<f8\x00\x00\x000\x00\x00\x00\x02\x00'
+                      b'\x00\x00\x03\x00\x00\x00\x02\x9a\x99\x99\x99\x99\x99\xb9?\x00\x00\x00\x00'
+                      b'\x00\x00\xe0?\x00\x00\x00\x00\x00\x00\x14@\x00\x00\x00\x00\x00\x00\x1c@\x00'
+                      b'\x00\x00\x00\x00\x00 @\x00\x00\x00\x00\x00\x00"@')
+
+        assert Serializer().type_and_object_serialization(obj) == serialized
+        assert np.allclose(DeSerializer(serialized).type_and_object_deserialization(), obj)
```
BenediktBurger, line 258: You could include this in the upper, parametrized test as well.

BenediktBurger (approved):
Looks good.

Thanks for tackling that issue.

I left a few suggestions.

BenediktBurger:
The changes are useful for the leco communication.

Thanks

seb5g:
I merged it in the 4.3.x_dev branch, so you can rebase  yours on it!

PR #328, opened 2024-06-13
Sliders now accept suffix and SIprefix if given in opts

Ashwolaa:
**Description**
Slider parameters can now display units with SIprefix
The syntax follows the NumericParameterItem format from pyqtgraph.

**Usage**
params =[ {'title': 'Linear Slide with suffix', 'name': 'linearslidewithsuffixandsiPrefix', 'type': 'slide', 'value': 50, 'default': 50, 'min': 0, 'max': 1e6, 'subtype': 'linear','suffix':'V','siPrefix':True}, ]

**Example**
Exemple of usage given in parameter_ex.py

https://github.com/PyMoDAQ/PyMoDAQ/assets/51032965/f94884d3-21e3-40e4-9528-8b29a769d52c

### DIFF
--- a/src/pymodaq/examples/parameter_ex.py
+++ b/src/pymodaq/examples/parameter_ex.py
@@ -25,12 +25,15 @@ class ParameterEx(ParameterManager):
             {'title': 'Linear Slide float', 'name': 'linearslidefloat', 'type': 'slide', 'value': 50, 'default': 50,
              'min': 0,
              'max': 123, 'subtype': 'linear'},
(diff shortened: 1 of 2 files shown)


seb5g (approved):

seb5g:
could you add some details on how to set this in the documentation?
and thanks , that looks nice!

Ashwolaa:
Sure, where should I put it?

In this part?
[(https://github.com/PyMoDAQ/PyMoDAQ/blob/4.3.x_dev/docs/src/api/utility_api/parameter/new_items.rst?)]

seb5g:
yes that sounds like a good location

PR #334, opened 2024-06-19
Added confirmation overwrite if FileExists

loicguilmard:
L'implémentation est dans le PresetManager mais il n'y pas de méthode dédiée, c'est ajouté dans le if res == dialog.Accepted:

Modifications que ioxml.py :
La sauvegarde étant directement gérer dans ioxml, j'ai rajouté Overwrite=True dans les paramètres de ioxml.parameter_to_xml_file pour garder une compatibilité si il y a des appels sans spécifier celui-ci.
J'ai aussi rajouté l'import de os.path pour la methode isfile.

Modification de PresetManager:
J'utilise un try: ioxml.parameter_to_xml_file(..., overwrite=False) except pour gérer un retour d'information sur l'existence du fichier.
En cas de FileExistsError, on fait une boite de dialogue, on récupère le choix et on écrase ou log que le user veut pas puis on continue l'exécution.

A noter que sauvegarder un nouveau preset ou modifier un preset arrive dans la même méthode, c'est seulement un prechargement du preset.
Donc on a l'overwrite warning qu'on ait fait new ou modif.

### DIFF
--- a/src/pymodaq/utils/managers/preset_manager.py
+++ b/src/pymodaq/utils/managers/preset_manager.py
@@ -176,6 +176,7 @@ class PresetManager:
                                             os.path.join(path, filename_without_extension),
                                             overwrite=False)
             except FileExistsError as currenterror:
+                logger.warning(str(currenterror)+"File " + filename_without_extension + ".xml exists")
                 confirmation = QtWidgets.QMessageBox()
                 confirmation.setWindowTitle('Overwrite confirmation')
                 confirmation.setText("File exist do you want to overwrite it")
@@ -185,6 +186,7 @@ class PresetManager:
                 if result == QtWidgets.QMessageBox.Yes:
                     ioxml.parameter_to_xml_file(self.preset_params,
                                                 os.path.join(path, filename_without_extension))
+                    logger.warning("File " + filename_without_extension + ".xml overwriten at user request")
                 else:
                     logger.warning("File "+filename_without_extension+".xml wasn't saved at user request")
                     # emit status signal to dashboard to write : did not saved ?
--- a/src/pymodaq/utils/managers/preset_manager.py
+++ b/src/pymodaq/utils/managers/preset_manager.py
@@ -7,6 +7,7 @@ import sys
 import os
 from pymodaq.utils.parameter import ParameterTree, Parameter
 from pymodaq.utils.parameter import ioxml
+from pymodaq.utils.messenger import dialog as dialogbox
 from pymodaq.utils import daq_utils as utils
 from pathlib import Path
 import pymodaq.utils.managers.preset_manager_utils  # to register move and det types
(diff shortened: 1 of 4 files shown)


src/pymodaq/utils/parameter/ioxml.py
```diff
@@ -348,9 +350,12 @@ def parameter_to_xml_file(param, filename: Union[str, Path]):
     fname = parent.joinpath(filename + ".xml")  # forcing the right extension on the filename
     xml_elt = walk_parameters_to_xml(param=param)
     tree = ET.ElementTree(xml_elt)
+    if not overwrite and os.path.isfile(str(fname)):
```
seb5g: fname is a Path object so you should use the corresponding method :
```
fname.exists() and fname.is_file()
```

src/pymodaq/utils/managers/preset_manager.py
```diff
@@ -169,8 +169,28 @@ def show_preset(self):
...
-                self.preset_params, os.path.join(path, self.preset_params.child('filename').value()))
+            filename_without_extension = self.preset_params.child('filename').value()
+
+            try:
+                ioxml.parameter_to_xml_file(self.preset_params,
+                                            os.path.join(path, filename_without_extension),
+                                            overwrite=False)
+            except FileExistsError as currenterror:
+                logger.warning(str(currenterror)+"File " + filename_without_extension + ".xml exists")
+                confirmation = QtWidgets.QMessageBox()
```
seb5g: I create a method for dealing with that, look at: *pymodaq.utils.messenger.dialog*

seb5g (changes requested):
Thanks for implementing this so quickly, I guess 4 hours in a train helps :-)

Could you apply my suggestions?

src/pymodaq/utils/managers/preset_manager.py
loicguilmard: thanks for the guiding, i thought i should implement the box somewhere else but wasn't sure where... :)

src/pymodaq/utils/parameter/ioxml.py
loicguilmard: do i really need to check if it exists and if it is a file ?
i don't think it can be a file if it doesn't exist.

seb5g: in python if the first clause of an *and* fails it doesn't test the second. But it can exist and be a directory... (unlikely I agree)

seb5g (changes requested):
Sounds code,
could you implement a test to check these two changes?

place them in 
/tests/utils/parameter_test_param_ioxml_test
and
/tests/utils/managers/parameter_manager_test

src/pymodaq/utils/parameter/ioxml.py
loicguilmard: I agree those 3 conditions meet the right logic. 
I just feel that not overwrite and fname.is_file() would still produce the same result as it can't be a file if it doesn't exist so is_file() is also a test of its existence, but i am just spliting hairs :)

loicguilmard:
> Sounds code, could you implement a test to check these two changes?
> 
> place them in /tests/utils/parameter_test_param_ioxml_test and /tests/utils/managers/parameter_manager_test

Looking into the parameter_manager_test, i understand why we'd merge the preset_manager_test there but shouldn't we get Qwidget from PresetManager.show_preset to harmonize behavior between our Objets and also between our different Tests ?

loicguilmard:
Any issue left to address ?

src/pymodaq/utils/managers/preset_manager.py
```diff
@@ -169,8 +170,24 @@ def show_preset(self):
...
-                self.preset_params, os.path.join(path, self.preset_params.child('filename').value()))
+            filename_without_extension = self.preset_params.child('filename').value()
+
+            try:
+                ioxml.parameter_to_xml_file(self.preset_params,
+                                            os.path.join(path, filename_without_extension),
+                                            overwrite=False)
+            except FileExistsError as currenterror:
+                logger.warning(str(currenterror)+"File " + filename_without_extension + ".xml exists")
+                userchoice = dialogbox(title='Overwrite confirmation',
```
seb5g: change userchoice to user_agreed to make it more explicit

src/pymodaq/utils/managers/preset_manager.py
```diff
@@ -169,8 +170,24 @@ def show_preset(self):
...
+                                       message="File exist do you want to overwrite it ?")
+                if userchoice == 1:
```
seb5g: then use only:
if user_agreed:
...

src/pymodaq/utils/managers/preset_manager.py
```diff
@@ -169,8 +170,24 @@ def show_preset(self):
...
+                    ioxml.parameter_to_xml_file(self.preset_params,
+                                                os.path.join(path, filename_without_extension))
+                    logger.warning("File " + filename_without_extension + ".xml overwriten at user request")
```
seb5g: construct complex string using the f'' syntax:

f"File {filename_without_extension}.xml overwriten at user request"

src/pymodaq/utils/managers/preset_manager.py
```diff
@@ -169,8 +170,24 @@ def show_preset(self):
...
+                else:
+                    logger.warning("File "+filename_without_extension+".xml wasn't saved at user request")
```
seb5g: use fstring

seb5g (changes requested):
could you change the few minor things I requested?

seb5g (approved):

PR #348, opened 2024-06-26
Add thread_command commands

BenediktBurger:
In this PR I added a file with the allowed commands for ThreadCommands.

The work is almost done (ThreadCommands from viewer UI are still missing).

What do you think?


seb5g:
I like it.  However we could go one step further by also defining and enforcing the type the attribute should take for each command. The init of ThreaCommand could check for the defined type

PR #372, opened 2024-09-10
modified the default config path for mac os

eduvieilbourg:
path problem to log file and preset actions


seb5g (approved):
ok thanks for the work

PR #319, opened 2024-06-12
Binary transfer via PyLECO

BenediktBurger:
This PR switches the LECO communication from base64 to pure binary transport.
However, that it works requires https://github.com/pymeasure/pyleco/pull/82 to be merged and published (pyproject.toml is already adjusted)

Closes #308

PR #404, opened 2024-10-16
Features/binary transport for PyLECO

### LINKED ISSUE #308: PyLECO binary transport
In https://github.com/pymeasure/pyleco/pull/82 I add native support for binary data transport via PyLECO.

I invite you to leave your feedback regarding how it should be implemented for usage.

BenediktBurger:
Follow up of #319 (now for Pymodaq 4.4) , closes #308


src/pymodaq/utils/leco/daq_move_LECODirector.py
```diff
@@ -143,22 +146,29 @@ def stop_motion(self) -> None:
...
+        self, position: Union[str, float, None], additional_payload=None
+    ) -> DataActuator:
+        if position:
+            if isinstance(position, str):
+                deserializer = DeSerializer.from_b64_string(position)
+                pos = deserializer.dwa_deserialization()
+            else:
+                pos = DataActuator(data=position)
+        elif additional_payload is not None:
+            pos = DeSerializer(additional_payload[0]).dwa_deserialization()
```
seb5g, line 159: what would be the additional payload?

seb5g (approved):
sound code

seb5g:
Will you explain a bit the inner mechanisms at the pymodaq days?

src/pymodaq/utils/leco/daq_move_LECODirector.py
BenediktBurger: How the binary transfer works:
The first payload frame contains the JSON response.
The second (and further) frames contain binary objects. These are the `additional_payload`.

Therfore, the sending party sends the binary version of the PyMoDAQ object in the first additional payload frame and here it gets deserialized.

BenediktBurger:
Yes, I will.

seb5g:
Then I propose we discuss further implementation after (or during) your talk. By the way I have trouble really undersanding the vocabulary payload and frame. Maybe you should explicit this in your talk

PR #413, opened 2024-10-23
Add example for qt-less standalone module using LECO

BenediktBurger:


src/pymodaq/examples/qt_less_standalone_module.py
```diff
@@ -0,0 +1,125 @@
...
+            receiver, method, kwargs = self._stored.pop()
+            self.communicator.ask_rpc(receiver=receiver, method=method, **kwargs)
+
+
+if __name__ == "__main__":
+    print("listening endlessly as 'qt_less'")
+    log = logging.getLogger()
+    log.addHandler(logging.StreamHandler())
+    log.setLevel(logging.DEBUG)
+    m = QtLessModule("qt_less")
```
seb5g, line 122: so if executing the script the name would be "qt_less" to be written in the director!

seb5g (changes requested):

src/pymodaq/examples/qt_less_standalone_module.py
BenediktBurger: I added a line to the top.

seb5g (approved):

PR #417, opened 2024-10-24
Add long_description_content_type to setup

rgeneaux:
this avoids warning from `twine check dist/*`

cherrypick to dev if you are still using setuptools in there?


src/pymodaq/resources/setup_plugin.py
```diff
@@ -23,6 +23,7 @@ def setup(path: Path):
         name=PLUGIN_NAME,
         description=config['plugin-info']['description'],
         long_description=long_description,
+        long_description_content_type='text/x-rst',
```
seb5g, line 26: how to do the same thing in a pyproject.toml? do you know?

seb5g (approved):

PR #429, opened 2024-10-28
Dashboard inherits CustomApp

### LINKED ISSUE #412: Improve splash screen when loading a preset
Make the text larger and perhaps enhance the screen, some random ideas:
- a loading bar that gets incremented each time of the module is correctly loaded?
- a list of modules to load with checkmarks that get added as they get loaded?

seb5g:
Implements the CustomApp behaviour.

In particular, this allow a better and cleaner use of QActions both for the menu and the toolbar
Introducing important QActions in the toolbar:

* select preset
* load it
* open DAQ_Scan
* open help
* open plugin manager

image

handle a better splashsreen:
image

### DIFF
--- a/src/pymodaq/control_modules/daq_viewer.py
+++ b/src/pymodaq/control_modules/daq_viewer.py
@@ -1018,7 +1018,8 @@ class DAQ_Viewer(ParameterControlModule):
         if self.settings.child('main_settings', 'overshoot', 'stop_overshoot').value():
             for dwa in dte:
                 for data_array in dwa.data:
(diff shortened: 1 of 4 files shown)


seb5g:
@rgeneaux could you have a look and maybe tell me what could be done on top of that?

seb5g:
solve #412

seb5g:
@rgeneaux could you install this branch and tell me if the user experience is better with the toolbar?
@quantumm idem
@aurore-finco idem

rgeneaux:
Toolbar is great! noticed this tooltip 
image
I think it should simply say "General settings"

seb5g:
good catch

PR #430, opened 2024-10-30
Improve the serializer

BenediktBurger:
Adding comments, type hints and an enum for consistent naming (you can call enum members).

Also makes the serializer symmetrical to the deserializer.

Fixes that a boolean is serialized as an integer, as the number serialization ran before the boolean test. (`isinstance(True, int) returns True`).


src/pymodaq/utils/tcp_ip/serializer.py
```diff
@@ -20,17 +21,38 @@
 
 
 SERIALIZABLE = Union[
-    int,
+    # native
+    bool,
+    bytes,
     str,
-    numbers.Number,
+    complex,  # float and int are subtypes for type hinting
```
seb5g, line 28: ok

src/pymodaq/utils/tcp_ip/serializer.py
```diff
@@ -205,18 +227,19 @@ def string_serialization(self, string: str) -> bytes:
...
         ----------
-        scalar: str
+        scalar: A python number (complex or subtypes like float and int)
 
         Returns
         -------
         bytes: the total bytes message to serialize the scalar
         """
         if not isinstance(scalar, numbers.Number):
+            # type hint is complex, instance comparison Number
```
seb5g, line 242: ok, then is not numbers.Numbers also valid for complex?

src/pymodaq/utils/tcp_ip/serializer.py
```diff
@@ -350,49 +373,56 @@ def list_serialization(self, list_object: List) -> bytes:
...
             bytes_string += self.list_serialization(obj)
 
         elif isinstance(obj, putils.ParameterWithPath):
             path = obj.path
             param_as_xml = ioxml.parameter_to_xml_string(obj.parameter)
-            bytes_string += self.string_serialization('parameter')
+            bytes_string += self.string_serialization(SerializableTypes.PARAMETER.value)
             bytes_string += self.list_serialization(path)
-            bytes_string += self.string_serialization(param_as_xml)
+            bytes_string += self.bytes_serialization(param_as_xml)
```
seb5g, line 422: indeed it is a bynary string

seb5g (approved):

src/pymodaq/utils/tcp_ip/serializer.py
BenediktBurger: The most permissive type hint is complex (allowing also float and int).
For instance comparison, you have to use numbers.Number, which will return true for complex, float, or int.

seb5g: ok then

PR #443, opened 2024-11-08
added the callable csv file in a ugly table in Library reference

loicguilmard:
I can't seem to change the table td font size  so we have to use shift+wheel to see the right side of the table.
It's usable and handy but not beautiful.

Note on the csv generation :
The csv file is not dynamic (yet ?) since i need to load the callable tree in 2 different environements.
I referenced all the methods, functions loops in github.com/loicguilmard/modulexplorer but it doesn't work as is.
It need 2 env and 3 runs (and debug/polishing).

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -7,3 +7,4 @@ pillow
 toml
 pyqt5
 numpydoc
+sphinx-datatables
--- a/docs/src/_static/css/overflow.css
+++ b/docs/src/_static/css/overflow.css
@@ -1,0 +1,3 @@
+.wy-table-responsive {
+    overflow: visible;
+}
--- a/docs/src/api/api_callable_v4to5.rst
+++ b/docs/src/api/api_callable_v4to5.rst
@@ -9,9 +9,10 @@ The following list is organized as such :
 * on the left the path of the object in pymodaq v4.
 * on the right colum the path of the same object in pymodaq v5.
 
-.. csv-table:: Callable from 4 to 5
+.. csv-table::
   :file: api_callable_v4to5/callablesfrom4to5.csv
   :header-rows: 1
+  :class: sphinx-datatable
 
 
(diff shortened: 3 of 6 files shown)


seb5g:
well, once you will have generated the differences, it is no big deal if the table is static? If you really want to, I guess you could use a github action to do so...

Have you checked:

* https://pypi.org/project/sphinx-csv-tools/
* https://github.com/alterat/tables-with-sphinx

loicguilmard:
yes i did.
the csv-tool is more about static filtering, but there is no material searchable table or something fancy like that.
the longtable doesn't break line with the point and i am not sure we want that.
the only solution i could apply was to change the font size in my brower :/
did you check the list ? 
1. It's really exaustive about the callable because it's hard to categorize what we are looking for (functions, methods, obejct type, etc...)
2. It also reference the callable that didn't move. I thought it could be interesting to actually find that a callable didn't move instead of "knowing by omission".

loicguilmard:
Found a sphinx extension to make it responsive.
I had to force overflow in a custom css to handle long entries.
It's a ugly for those but it's much better that without the extension and the overflow.

seb5g:
That's a huge list indeed! Could be a good reference with a ctrl+F you find where a class is now! I noticed you also included class that are imported. The red stuff below are imported things. They should not appear here. I mean on the left side of your table should appear the class that are defined (not imported) and on the right the new place where they are defined. This is important (I think) because I left some modules as they were before where I removed the class definition and imported them from their new location. This allows backcompatibility but should not be reference in your table.

image

loicguilmard:
Then refuse this PR, my method was wrong. It was a module exploration to find callables. If you import it, it's a callable.

seb5g:
the approach is still valid, you just have to check wether things are declared? I don't know how to do that but it should be doable

PR #447, opened 2024-11-13
Update/callables4to5

loicguilmard:
New version of the csv file with project [PyMoDiff](https://gitlab.in2p3.fr/loic.guilmard/pymodiff).

The responsive table is searchable directly : 
searchable-table


seb5g:
looks nice!!
Could you sort the left column before saving the csv? That would put related things together (for instance everything from pymodaq.utils.data

You should change the documentation section name. Callable list is what it is but the most important thing is that is shows the old/new location of the callable between v5 and v4. Maybe also add a dedicated section in developer's guide refering to this list and also explaining the whole thing about v5 (here is the link of my [presentation ](https://drive.google.com/file/d/10IP6i8ZpWbjQkaea0Wezi2EbwwNEO3l9/view?usp=sharing)on the subject)

loicguilmard:
renaming the what's new in pymodaq 4 or a new section what's new in pymodaq 5 ?

seb5g:
I would say a new section in case some oldschooler are still on v3



Le mer. 13 nov. 2024, 23:03, loicguilmard ***@***.***> a
écrit :

loicguilmard:
I am now add mention about the changes on python packaging and the hatch part into this "what changed".
But i will add a more detailled plugin tutorial migration and hatch documentation later.

seb5g:
should I review it now then?

Maybe we should do some heavy testing to check that a 4.x plugin is still working with pmd5 and add whatever is needed in pmd5 for this to work almost effortlessly!

loicguilmard:
You can review it now :)
I'll check our two plugins with pmd5 and let you know.

loicguilmard:
The CI is not working for my branch/update because of the versionning, it goes back to 4.2.2 provoking some dependencies issues.
Should i add a git tag just for documentation to fix that or should you as you get it in ?

seb5g:
> The CI is not working for my branch/update because of the versionning, it goes back to 4.2.2 provoking some dependencies issues.
> Should i add a git tag just for documentation to fix that or should you as you get it in ?

@malik-irain fixed this or is in the process of it!

seb5g:
I'm just looking at your docs and will write down below things I notice while reading:

* nice table of changes
* the link just after 5.6.1. pymodaq_utils (the title could be *PyMoDAQ Utils*) , is pointing to Utility libraries which is actually containing a mess of many different things. The next job on doc should therefore to clean the API doc to reflect the new package structure (that will actually helps)
* pymodaq_gui includes also all the Parameter objects extensions of pyqtgraph. @PauloGitHB is working on a possible PR to pyqtgraph
* 5.6.3.1 Managers: you should separate ParamterManager from the Parameter module. The first helps to deal with Parameters in pymodaq modules while the latter introduces extensions of Paramter objects from pyqtgraph and utils to deal with them
* the plotting doc link is not working

Could you do these modifications? Then in a second time work on reorganising the library api? Thx for the work

loicguilmard:
For the API doc, now that's it's external packages, the submodule makes sense to me :p It's messy to document and debug in a centralized way now :/
I think it didn't really understand the difference between ParameterManager and Parameter module. I think both inherited and extended the pyqtgraph parameter objects. Just Parameter being an attribute of ParameterManager.

seb5g:
do you want to do a visio on the subject?

loicguilmard:
I just sticked to the Manager part, putting the ParameterTree as an attribute in the list.

seb5g (approved):

seb5g:
I will merge your PR and we will make modification from there if need be after few people read it!

seb5g (approved):

seb5g:
loic could you have a look at the conflicts in the merge?

loicguilmard:
@seb5g done !

PR #450, opened 2024-11-16
Feature/documentation arduino ubuntu

quantumm:
Start the Examples section of the documentation, which is intended to host a gloubi-boulga of stories from the users, where they describe their use of Pymodaq.


seb5g:
You should write new piece of documentation on the 5.x branch! 
Seb

seb5g:
or maybe for both... as it will be compatible

quantumm:
I close this PR since it is replaced by #461.
I don't think it is necessary to push it also in the v4 branch since the documentation is already by default displayed for the v5.

Ciao!

PR #453, opened 2024-11-21
fixing build error for tables version depending on python version

loicguilmard:
After talking with an old user reinstalling pymodaq and new users from dijon university, i investigated a small installation issue :
* tables < 3.9 doesn't install with python 3.12
* tables 3.10 fixed the blosc error that imposed tables < 3.9 in pyproject.toml

so modified pyproject.toml accordingly to conditionnally install the right version of tables.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,8 @@ dependencies = [
     "simple_pid",
     "toml",
     "qtconsole",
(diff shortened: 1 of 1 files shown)


seb5g:
I applied your patch to pymodaq_data also as it is there the requirement is for pymodaq >= 5

PR #456, opened 2024-11-25
Use QShortcut in Qspinbox

Ashwolaa:
In a DAQ_move world:
- where you need to type to change the value
- but you need to click to update the value

Life gets cumbersome as us mere humans only have two hands.

As such, I propose as a solution to free some people from this click by implementing a QKeySequence for QSpinBox.

Here is the demo
https://github.com/user-attachments/assets/a2c224b1-baaa-414e-bad0-1ad505dd5c66

In pratice, there is a new class QSpinBoxWithShortcut which stores the shortcuts in a dictionnary.

You can connect the activation signal of the shortcuts to trigger the function of interest.

In the case of DAQ_move:
- I chose Ctrl+Enter for the move abs

For the move_rel:
- Ctrl+Enter goes up
- Ctrl+Shift+Enter goes down


seb5g:
I like the idea but I would love to have these identified in the configuration file for people to specify their own if needed!

seb5g:
you could also add ctrl+g and ctrl+r (for green and red absolute moves)

seb5g:
what happens if you have multiple daq_move in a dashboard? Have you looked at the remote manager features?

Ashwolaa:
> what happens if you have multiple daq_move in a dashboard? Have you looked at the remote manager features?

I need to test it but the implementation is currently done in a "safe" way as you need to be in the spin box widget focus for the event to be detected.


The g/r would not work as you can hope with this PR for example since there is only one focus.

Ashwolaa:
> I like the idea but I would love to have these identified in the configuration file for people to specify their own if needed!

I agree, it just started to get a bit more cumbersome in the event filter when using keys accepted by qspinbox (which is not not only num with the units).

There are some nice features for it with qt6 that I gave up as this will have to wait when the transition is good to go

seb5g:
> > what happens if you have multiple daq_move in a dashboard? Have you looked at the remote manager features?
> 
> I need to test it but the implementation is currently done in a "safe" way as you need to be in the spin box widget focus for the event to be detected.
> 
> The g/r would not work as you can hope with this PR for example since there is only one focus.

is it not a bit limited then if you need to be at the focus of the sb? The remote manager is working when the focus is anywhere within the dahboard. Is that not better?

As for qt6, I've tested it for quite some time and didn't see any issues. but offcourse using new features will make the app not backcompatible but I don't think it is a problem as long everything is running with Qt6

Ashwolaa:
> > > what happens if you have multiple daq_move in a dashboard? Have you looked at the remote manager features?
> > 
> > I need to test it but the implementation is currently done in a "safe" way as you need to be in the spin box widget focus for the event to be detected.
> > 
> > The g/r would not work as you can hope with this PR for example since there is only one focus.
> 
> is it not a bit limited then if you need to be at the focus of the sb? The remote manager is working when the focus is anywhere within the dahboard. Is that not better?

Yes and no,
Yes because it only works in the focus.
No as you need in any case to be in the focus of the widget if you want to edit it. 
In a way, this feature is just here to give back access to the trigger of validating the value of the spin box which is not used for safety reasons if I remember correctly.
I was used to go with tab/shift+tab to switch from red/green as they are successive to one another 


> As for qt6, I've tested it for quite some time and didn't see any issues. but offcourse using new features will make the app not backcompatible but I don't think it is a problem as long everything is running with Qt6

Ok great, I am all fine switching to Qt6. I just want to avoid developing features twice, one for Qt6 and one for qt5

seb5g (approved):

seb5g:
I'l port your implementation to pymodaq_5

PR #461, opened 2024-11-29
Documentation - Introduce lab stories - Arduino on Ubuntu.

quantumm:
Introduce the _Lab stories_ section of the documentation to encourage the users to describe what they do with Pymodaq in their lab, have a poll of examples, discuss physics (beyond Pymodaq), give ideas for a practical work at the university...


quantumm:
Also, as you will see, I put the section in the main menu, which makes it very visible. This may not be a good idea, it should be discussed. It is possible that I have been too much enthusiastic about it ;)

Anyway, what I kept during the last pymodaq days from the beginers is that it is tedious to go through the documentation, and it can be a better way to learn through examples, that is probably the main point.

seb5g:
> Anyway, what I kept during the last pymodaq days from the beginers is that it is tedious to go through the documentation, and it can be a better way to learn through examples, that is probably the main point.

then maybe we could pinpoint within the examples the locations of the corresponding documentation?

seb5g:
Just compiled your docs:

ups:

* it looks very nice
* it fits well as a section in the main menu

downs:

* your pymodaq sticker is in a very bad shape :-(
* seriously: you use pyfirmata2 (which works fine) but all the examples I added and @jerlfan will add will use telemetrix. So maybe at least explain the few options (pyfirmata, telemetrix...)
* you start from the template while one arduino plugin exists, it would be better to update it then!

thx

seb5g:
@quantumm could you answer my comments?

quantumm:
Yes I plan to work on it in the next weeks ;)

seb5g (approved):

quantumm:
Hey! I am not done here! :D

Actually there are still some typos to fix. And also I would like to find a way to directly get the reading of the viewer as a temperature (without coding if possible): see #497

Do you have a solution for that?

seb5g:
sorry I thought it was completed. I reverted the merged so it won't appear yet in the docs, You should probably create a new PR...

PR #463, opened 2024-12-02
Pyside6 GitHub actions

malik-irain:
I added what was needed to run the Github Actions to run test using PySide6 as a Qt backend. I also made two small modifications in the source code, otherwise these tests failed:

- in file `tcp_server_client.py` removed `super().__init__()` in `TCPClientTemplate` that doesn't inherit anything. When  `TCPClient(TCPClientTemplate, QObject)` was instanciated, it was calling  `QObject.__init__(self)`, then  `TCPClientTemplate.__init__(self, ...)`, which was then doing the `super().__init__()` call that is now removed. As `TCPClientTemplate` doesn't inherits anything, it ended up calling `QObject.__init__()`  a second time and that made some test fails. Now it is only called once, in `TCPClient.__init__()`.
- in file `utils.py` and in class `ParameterControlModule`, a `QObject.__init__()` call was also removed for similar reasons, it was called lower in the inheritance hierarchy, by a class directly inheriting QObject.


seb5g:
We should maybe rationalize the number of combination we are testing. I was thinking to focus on 3.11:

* 3.8 pyqt5
* 3.9 pyqt5
* 3.10 pyqt5
* 3.11 pyqt5
* 3.11 pyside2
* 3.11 pyqt6
* 3.11 pyqt6 windows
* 3.11 pyside6
* 3.12 pyqt6

The **readme** should reflect this testing matrix

seb5g:
maybe we can check on how other project are doing because it starts to be a lot of tests... and then we have to update the readme accordingly when we add a new test... maybe adding some automation would be could

malik-irain:
I agree, I actually already started to implement it with a matrix strategy like in [River](https://github.com/online-ml/river/blob/main/.github/workflows/unit-tests.yml), so that github actions automatically try all the combinations, but it seems it's not possible to have a badge for a specific config this way, only the global badge for all combinations (so if one config fails, it's considered as failed).

I will investigate, maybe some external tools or actions allow to generate badges.

seb5g:
yes I was doing this before, but I like the fact to know which combination fail and which doesn't... thanks for the investigation

seb5g:
that's quite an awesome githubaction! 

however seeing the output on the readme, I figure it would be better to have a smaller table with just a green check and a red cross to have the same effect! Do you think you can produce this simpler table in the readme from the githubaction?

seb5g:
and sorry it seems you spend quite some time on those badges...

malik-irain:
Yes, it's possible to change them, I tried a few but the check marks are never centered. Do you prefer something like:

passing/failing

Or maybe:
pass/fail


Or:
check/cross


I prefer the text based ones, anyways, it's configurable in the workflow file.

(If it's ok, don't accept the PR right know, it needs a little bit of setup that I will explain and it's preferable to start with pymodaq_utils and pymodaq_data before going to pymodaq_gui and pymodaq)

seb5g:
I'm happy to go with the passing string!

Issue #491, opened 2024-12-17
Get crosshair position


bzarie26:
On daq_viewer2D, when I use the crosshair, I want to retrieve its position in order to obtain some information. However, it doesn't exist any method available which allow to get this information.

Is it possible to have a method like "get_crosshair_position" that I can use in my plugin?

Thank you in advance for any help or guidance you may be able to provide.

loicguilmard:
Sebastien wrote an exemple in [DataPicker](https://github.com/PyMoDAQ/pymodaq_gui/blob/5.0.x_dev/src/pymodaq_gui/examples/data_picker.py)
I think it's from the viewer :
```
            xs, ys = self.viewer.view.unscale_axis(posx, posy)
            data_at = self.viewer.view.get_data_at('red', (xs, ys))
```

If you are using PyMoDAQ 4, here are the version 4 import path for the data_picker.py :

```
from qtpy.QtCore import Qt
from qtpy import QtWidgets

from typing import Optional

from pymodaq.utils.data import DataRaw, Axis


from pymodaq.utils.gui_utils.widgets.table import SpinBoxDelegate
from pymodaq.utils.parameter.utils import get_widget_from_tree
from pymodaq.utils.plotting.data_viewers.Viewer2D import Viewer2D
from pymodaq.utils.gui_util import CustomApp, Dock, DockArea

from pymodaq.utils.parameter.pymodaq_ptypes.tableview import TableViewCustom
from pymodaq.utils.scanner.scanners.tabular import TableModelTabular
```

seb5g:
solved in #492

seb5g:
you'll need to use version 5 from github to use it for the moment

PR #492, opened 2024-12-17
add a mechanism to obtain the crosshair from a viewer in a plugin

seb5g:
Introduces these functions into the viewer_base which viewer plugins inherit from:
```
    def roi_select(self, roi_info: RoiInfo, ind_viewer: int = 0):
        """ Every time a ROISelect is updated on a 2D Viewer,
        this method receive the corresponding info

        To be subclassed in a plugin to use the info

        Parameters
        ----------
        roi_info: RoiInfo
        ind_viewer: int
            The index of the viewer (if multiple) in which the roi is declared
        """
        pass

    def crosshair(self, crosshair_info: Iterable[float], ind_viewer: int = 0):
        """ Every time a crosshair is updated, this method receive the corresponding info

        To be subclassed in a plugin to use the info

        Parameters
        ----------
        crosshair_info: list of float
        ind_viewer: int
            The index of the viewer (if multiple) in which the crosshair is declared
        """
        pass
```

### DIFF
--- a/src/pymodaq/control_modules/daq_viewer.py
+++ b/src/pymodaq/control_modules/daq_viewer.py
@@ -390,7 +390,7 @@ class DAQ_Viewer(ParameterControlModule):
             return self._viewers
 
     @viewers.setter
-    def viewers(self, viewers):
+    def viewers(self, viewers: List[ViewerBase]):
         for viewer in self._viewers:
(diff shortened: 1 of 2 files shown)


seb5g:
Closes #491

PR #526, opened 2025-02-10
Fix (v4) for "Update GitHub Action Versions" fail

malik-irain:
As suggested by `apt install` output after failing the action, repositories are now updated before installing packages:

```
...
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```

I also manually updated actions versions.


seb5g:
should'nt this be also applied to v5?

malik-irain:
I already did it, but at the time I forgot to also apply it to v4

seb5g:
ok!!

PR #534, opened 2025-03-03
Integrate bayesian-optimization package v2

malik-irain:
It integrates bayesian-optimization in version 2 and correct the way bayesian optimization extension's config is saved.

It also fixes the way thread are terminated (with a less brutal approach when not necessary)


src/pymodaq/control_modules/utils.py
```diff
@@ -174,13 +174,11 @@ def thread_status(self, status: ThreadCommand, control_module_type='detector'):
             try:
                 self.update_status(status.attribute[0])
                 self._hardware_thread.quit()
-                self._hardware_thread.wait()
-                finished = self._hardware_thread.isFinished()
-                if finished:
-                    pass
-                else:
+                terminated = self._hardware_thread.wait(10000)
```
seb5g: maybe remove this from the PR and create another one as it is unrelated...

src/pymodaq/extensions/bayesian/bayesian_optimisation.py
```diff
@@ -58,58 +60,49 @@ class BayesianOptimisation(CustomExt):
...
-         'visible': False, 'children': [
-             {'title': 'Units:', 'name': 'units', 'type': 'str', 'value': ''}]},
-
-    ]
+
+    acquisition_functions_names = list(GenericAcquisitionFunctionFactory.keys())
+
 
     def __init__(self, dockarea, dashboard):
+        self.params = [
```
seb5g: why is it here and not in the class preamble? The main purpose to have it in the class preamble is for introspection and settings generation even if the object is not instanciated!

src/pymodaq/extensions/bayesian/bayesian_optimisation.py
```diff
@@ -662,7 +666,7 @@ def main(init_qt=True):
 
     dashboard = DashBoard(area)
     daq_scan = None
-    file = Path(get_set_preset_path()).joinpath(f"{'beam_steering_mock'}.xml")
```
seb5g, line 665: for this main function, use the mkapp method from pymodaq_gui to simplify the initialization and avoid bugs! check on the daq_scan extension if needed!

src/pymodaq/extensions/daq_scan.py
```diff
@@ -887,11 +888,11 @@ def start_scan(self):
             # mandatory to deal with multithreads
             if self.scan_thread is not None:
                 self.command_daq_signal.disconnect()
-                if self.scan_thread.isRunning():
+                self.scan_thread.quit()
```
seb5g: same as for the control modules remove this out from here into a dedicated branch (you can cherry pick the corresponding commit if this modifcation has been done alone)

src/pymodaq/extensions/pid/pid_controller.py
```diff
@@ -153,11 +153,11 @@ def ini_PID(self):
 
         else:
             if hasattr(self, 'PIDThread'):
-                if self.PIDThread.isRunning():
-                    try:
-                        self.PIDThread.quit()
-                    except Exception:
-                        pass
+                self.PIDThread.quit()
```
seb5g: same as above! it would also be easier to debug if this causes issues at the end..

seb5g (changes requested):
thx Malik for the work! Some stuff to do:

* move Qthread termination in a dedicated PR
* discuss why you remove params from the class preamble

some minor other things

src/pymodaq/extensions/bayesian/bayesian_optimisation.py
malik-irain: it can be moved outside, it was done this way because I first instantiated `acquisition_functions_names` in the constructor, when I changed I didn't move the `param` variable out

malik-irain:
I reverted the changes related to thread termination, used mkQApp function and moved params in the preamble

seb5g:
In fact there is still something to do: check the compat with the eventually existing config file (storing values of settings given the choice of detectors/actuators) and raise a popup window asking to delete this file (and do it if yes) in case of conflict

src/pymodaq/dashboard.py
```diff
@@ -66,6 +66,12 @@
 remote_path = config_mod_pymodaq.get_set_remote_path()
 
 
+def error_dialog(message) -> int:
```
seb5g: you could call the messbox from the pymodaq_gui.messenger module

src/pymodaq/extensions/bayesian/bayesian_optimisation.py
```diff
@@ -43,6 +43,17 @@
 logger = set_logger(get_module_name(__file__))
 config = config_mod.Config()
 
+def find_key_in_nested_dict(dic, key):
```
seb5g, line 46: there is exactly this in pymodaq_utils;utils

src/pymodaq/extensions/bayesian/bayesian_optimisation.py
```diff
@@ -140,6 +151,20 @@ def __init__(self, dockarea, dashboard):
         self.settings.child('models', 'ini_runner').sigActivated.connect(
             self.get_action('ini_runner').trigger)
 
+    def validate_config(self) -> bool:
+        kind = find_key_in_nested_dict(self.bayesian_config.to_dict(), 'kind')
```
seb5g: do we check only this?

seb5g (changes requested):

src/pymodaq/dashboard.py
malik-irain: yes thanks I didn't see that!

src/pymodaq/extensions/bayesian/bayesian_optimisation.py
malik-irain: hmm, I'm not so sure about that?

I need a function that can find the "kind" entry in a nested dict, that can look like this:
```python
{'BayesianModelDefault': 
	{'Move 00': 
		{'main_settings': 
			{'utility': 
				{'kind': 'ucb'}
			}
		}
	}
}
```
or this:
```python
{'BayesianModelDefault': 
	{'Move 00': 
		{'Move 01': 
			{'main_settings': 
				{'utility': 
					{'kind': 'ucb'}
				}
			}
		}
	}
}
```

Or with even more actuators that can be named anything

I didn't find such a function in pymodaq_utils.utils

src/pymodaq/extensions/bayesian/bayesian_optimisation.py
malik-irain: I can check if instantiation is possible, but not so much more

seb5g (approved):

PR #542, opened 2025-03-05
Fix serialize imports in LECO and in deprecated pymodaq folder

rgeneaux:
Ran into this when trying out LECO in v5.0


Ashwolaa (commented):
I had to do a similar change in src/pymodaq/utils/tcp_ip/serializer.py before being able to trigger issue https://github.com/PyMoDAQ/PyMoDAQ/issues/543. 

The deprecation message should be updated in src/pymodaq/utils/tcp_ip/serializer.py to correctly refer to pymodaq_utils

These should be pushed to 5.0.x

Ashwolaa:
> I had to do a similar change in src/pymodaq/utils/tcp_ip/serializer.py before being able to trigger issue #543.
> The deprecation message should be updated in src/pymodaq/utils/tcp_ip/serializer.py to correctly refer to pymodaq_utils
> These should be pushed to 5.0.x

PR #466, opened 2024-12-04
Use new serializer with LECO parts

BenediktBurger:
Close #462 . (EDIT: automatically closing does not work, as it won't merge into the current main branch)

TODOs
- [x] write code
- [x] Test move
- [x] test viewer

PR #552, opened 2025-03-11
Revert "Use new serializer with LECO parts"

seb5g:
Reverts PyMoDAQ/PyMoDAQ#466


seb5g:
reverted it as it was done not on the dev branch...I'l do another PR for this!

Issue #550, opened 2025-03-11
[BUG] Inconsistent use of units between pyleco actor and director as a DAQ_Move


seb5g:
### Describe the bug

A LECO DAQ_Move_Director, because it never receive the actor settings, does'nt know in which units it should express the DataActuaotr it's receiving from the Actor. Hence this should be handled, maybe a message stipulating when the actor is having it's units changed (through self.axis_units setter) would do!

### Reference(s)

```markdown
pymodaq.utils.config.deep_update
```

### Environment

```markdown
- OS:
- Pymodaq:
- python env:
```

seb5g:
closed by #554

Issue #548, opened 2025-03-10
[BUG] Inconsistent behavior of move_rel in daq_move_LECODirector


Ashwolaa:
### Describe the bug

The method `move_rel `of the `DAQ_Move_LECODirector `only works on its first iteration. If one uses `move_rel `after a previous `move_rel `or a `move_abs`, the program crashes as it can no longer acces the attribue `current_value`.

### Expected Behavior

The `current_value `attribue should not disappear after ordering a move_command

### Steps To Reproduce

On a dashboard
-- Start LECO coordinator
-- Load a Mock DAQ_Move with name Move 00
-- Connect Stage to LECO coordinator (localhost)
On a custom script (see below an example)
-- Start a DAQ_Move_LECODirector
-- Connect director to stage

Base script:
```
from pymodaq.utils.leco.daq_move_LECODirector import DAQ_Move_LECODirector
director = DAQ_Move_LECODirector()
director.settings.child("actor_name").setValue("Move 00") #Name of DAQ_MOVE
status = director.ini_stage()
```

Example 1 (working):
`  director.move_rel(50) # Move to 50 (This step works at intended)
`
Example 2 (not working):
```
  director.move_rel(50) # Move to +50 (First times works)
  director.move_rel(50) # Move to +50 (Second times crashes)
```
Error:
```
Exception has occurred: AttributeError
'DAQ_Move_LECODirector' object has no attribute 'current_value'
  File "C:\Users\constant.schouder\Documents\PyMoDAQ\PyMoDAQ\src\pymodaq\utils\leco\daq_move_LECODirector.py", line 126, in move_rel
    position = self.check_bound(self.current_value + position) - self.current_value  # type: ignore  # noqa
                                ^^^^^^^^^^^^^^^^^^
  File "C:\Users\constant.schouder\Nextcloud\Partages_recus\DIRAM\Python\PyLECO\SPECS\daq_move_controllerv3.py", line 98, in main
  File "C:\Users\constant.schouder\Nextcloud\Partages_recus\DIRAM\Python\PyLECO\SPECS\daq_move_controllerv3.py", line 120, in <module>
AttributeError: 'DAQ_Move_LECODirector' object has no attribute 'current_value'
```

Example 3 (not working):
  director.move_abs(50) # Goes to 50
  director.move_rel(50) # Move to +50 (Crashes)

Error:
Same as in example 2

### Reference(s)

```markdown
[...]

Issue #471, opened 2024-12-05
LECOMoveDirector should handle units correctly


BenediktBurger:
I found out, that the LECOMoveDirector does not handle units, yet.

The MockMove is set to 1µm as base unit, thereofre the LECOMoveDirector indicates positions as pure numbers corresponding to steps of 1 µm (unknown to the observer).

Goal: MoveDirector shows/handles the same units as the controlled Move itself.

BenediktBurger:
I guess, that this causes the relative move to not work as expected.

seb5g:
data objects have handling units functionalities and tools to transform them if need be. at some time the plugin method : self.axis_unit = ... is triggered, this should probably send a signal using pyleco for the correspongin director to update!

seb5g:
related to #550 and #548

PR #405, opened 2024-10-16
Features/pyleco events

BenediktBurger:
Continuation of #318 , but for pymodaq 4.4

If someone could test it, I'd appreciate it.

Maybe we can talk about it in Lyon.

TODO
- [ ] Update qt-less example

PR #554, opened 2025-03-11
Bugfix/leco move units

seb5g:
This PR solves three issues: #471 and #550  and #548 

The main issue was related to the fact the director never received the information about what are the actor's units. I recoded the same things as in the TCP/IP legacy, namely sending through the content of the actor settings from which the units are inferred. Once this has been done most issues with relative move and units where fine.

I nevertheless also added (or removed) the polling mechanism in the case of a director because the move_done signal is transmitted by the actor to the director so no need to poll (it would actually quite a bit of overwork)

Modifying settings of the actor from the Director is working and modifying the settings from the actor is updating the corresponding ones on the director

### DIFF
--- a/src/pymodaq/control_modules/daq_move.py
+++ b/src/pymodaq/control_modules/daq_move.py
@@ -715,6 +715,7 @@ class DAQ_Move(ParameterControlModule):
             self._command_tcpip.emit(ThreadCommand(LECOCommands.SET_SETTINGS,
                                                    ioxml.parameter_to_xml_string(self.settings.child('move_settings'))))
 
+
 class DAQ_Move_Hardware(QObject):
     """
         ================== ========================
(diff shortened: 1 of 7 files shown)


seb5g:
still some work on the single param update!

seb5g:
@malik-irain , @Ashwolaa , this has been requiring quite a bit of involvement to really get into it and make it work. Could you check the code in this PR is fine and is not inducing strange side effects? 

 @BenediktBurger  I could see many entangled things between previous serializer b64 encoding and the new one. There is also many things from the TCPServer that are not working anymore. Do you think you could clean it a bit? Or would it be better to wait for the work on PR #405 

I will implement the same things about settings for the DAQ_Viewers

seb5g:
you can use the [bugfix/leco_move_units](https://github.com/PyMoDAQ/PyMoDAQ/tree/bugfix/leco_move_units) branch as I will be merging the PR.

Issue #551, opened 2025-03-11
LECO director should have their corresponding actor settings and should be able to update them and be updated from them


seb5g:
This is actually done by the legacy TCP/IP communication protocol so should be ok to implement!

seb5g:
done by #558

Issue #549, opened 2025-03-10
Mismatch between the detector and director's point counts, when using LECO


malik-irain:
It seems that the director "can't grab" as many points as the detector outputs. It may be because it calls the wrong method to grab.

![Image](https://github.com/user-attachments/assets/79c4666b-7c93-4636-b941-9184bddd4f53)

seb5g:
solved by #558

PR #558, opened 2025-03-14
implementation of leco plugin settings in viewer and grab/stop implemented

seb5g:
solve #549 and #551 

I implemented for viewer what I did in #554 for the daq_move to the daq_viewer. Also the grab/snap was not behaing properly and the stop method was not implement by the director

started to clean a bit by using enum allowed methods


seb5g:
hi @malik-irain @Ashwolaa @BenediktBurger 
Both types of control module seems to be fully compatible now with pyleco with settings management and also stuff we saw missing

could you try it and let me know if anything seems missing

PR #559, opened 2025-03-14
Cleanup plyeco

### LINKED ISSUE #555: Clean up LECO serializer use
@BenediktBurger  I could see many entangled things between previous serializer b64 encoding and the new one. There is also many things from the TCPServer that are not working anymore. Do you think you could clean it a bit? Or would it be better to wait for the work on PR #405 

_Originally posted by @seb5g in https://github.com/PyMoDAQ/PyMoDAQ/issues/554#issuecomment-2718265434_

Do you mean that with the new serializer, we can remove the base64?
What do you mean regarding non working parts of the TCPServer to clean up?

BenediktBurger:
Maybe closes #555 

Most importantly:
- remove base64 en/decoding
- remove obsolete relicts from TCPIP protocol in PyMoDAQ
- improve documentation


seb5g (approved):

seb5g:
thx Benedikt !

PR #627, opened 2025-05-20
Allow JSON data with LECODirector

### LINKED ISSUE #464: "Lite" version to support old equipment
Sometimes, old hardware or OS (generally only able to run old python versions) are needed for some equipment.

One solution is to use an old PyMoDAQ version but it isn't always possible and doesn't provide more recent functionalities.

So, It would be nice to have a lite version of PyMoDAQ (with special daq_move / daq_viewer) that only interact with the hardware plugin on the one side, and communicates with a full remote PyMoDAQ version using PyLECO on the other.

Also documentation should be updated to provide an example on how to deploy on these old computers.

It may raise some security issues as an old OS like Windows XP for example **should not be connected to the Internet**. But as long as its only the lab internal network or a direct connection to a control computer it should be fine (at least as much as possible).

malik-irain:
Allow "basic" communication outside of PyMoDAQ in which actuators are able to share their units and detectors are able to share a label, and their axes. Detectors may be 0, 1 or 2D.

When in 2D mode, axes need to be indexed, I made it so that the first axis is the x-axis (index = 1) and the second one the y-axis (index = 0). So it's kind of the inverse of what PyMoDAQ does. Is it ok @seb5g or do you prefer to keep the same order?

closes #464


seb5g:
it would be better to keep the same order (numpy style) and let people know in the doc.

seb5g:
Also could you point where to find the example code of a NoPyMoDAQ actor?

src/pymodaq/utils/leco/daq_move_LECODirector.py
```diff
@@ -98,6 +102,7 @@ def ini_stage(self, controller=None):
         else:
             self.controller = controller
 
+        self.json = False
```
seb5g, line 108: why not in the class preamble as for daq_viewer?

src/pymodaq/utils/leco/daq_move_LECODirector.py
```diff
@@ -109,13 +114,17 @@ def move_abs(self, position: DataActuator) -> None:
...
+            position = position.value(self.axis_unit)
         self.controller.move_abs(position=position)
 
     def move_rel(self, position: DataActuator) -> None:
         position = self.check_bound(self.current_value + position) - self.current_value  # type: ignore  # noqa
         self.target_value = position + self.current_value
 
         position = self.set_position_relative_with_scaling(position)
+        if self.json:
+            position = position.value(self.axis_unit)
```
seb5g: maybe add a comment here saying that for json one has to send a float. Does it mean it cannot be something other than a float? What if it is a SLM with pixels?

src/pymodaq/utils/leco/daq_xDviewer_LECODirector.py
```diff
@@ -108,19 +112,41 @@ def stop(self):
...
             dte = SerializableFactory().get_apply_deserializer(additional_payload[0])
+        elif data is not None:
+            axes = []
+            labels = []
+            if isinstance(data, dict):
+                axes = [
+                    Axis( label=axis.get('label', ''),
+                          units=axis.get('units', ''),
+                          data=np.array(axis.get('data', [])),
+                          index=i
```
seb5g: never use just i as index for enumeration ;-), please use ind or something clearer

seb5g: is there an interface documentation  of how should be constructed such data as json dict?

src/pymodaq/utils/leco/daq_xDviewer_LECODirector.py
```diff
@@ -108,19 +112,41 @@ def stop(self):
...
+                data = data.get('data', [])
+
+            np_data = np.atleast_1d(data)
+            if np_data.ndim <= 2:
+                data = [np_data]
+            else:
+                data = [np.atleast_1d(d) for d in data]
+
+            dfp = DataFromPlugins(self.controller.actor, data=data, axes=axes[:data[0].ndim], labels=labels)
+            dte = DataToExport('Copy', data=[dfp])
```
seb5g, line 149: what's the reason for the copy?

seb5g (changes requested):

src/pymodaq/utils/leco/daq_xDviewer_LECODirector.py
malik-irain: We have (almost) the same result when the payload is binary. A `DataToExport` named "Copy" containing a `DataFromPlugins`

src/pymodaq/utils/leco/daq_xDviewer_LECODirector.py
malik-irain: The documentation is not ready yet!

src/pymodaq/utils/leco/daq_move_LECODirector.py
malik-irain: Yeah, it needs to be reworked, I didn't think about more dimensional data types

src/pymodaq/utils/leco/daq_move_LECODirector.py
malik-irain: My mistake, it shouldn't even be here in `daq_viewer` as it is directly determined by if we received an additional_payload or not

I think it should be in the constructor as its value is not meant to be shared across instances but determined by the context

src/pymodaq/utils/leco/daq_move_LECODirector.py
malik-irain: @seb5g How are data organized in a DataActuator sent to a SLM? What is the shape of `data` attribute?

seb5g: just the same as a DataWithAxes: data is a list of numpy arrays of fixed shape but could be 0D, 1D or 2D

PR #630, opened 2025-05-30
Bugfix/check data type

Ashwolaa:
I do not fully understand how the check_data_type in the DAQ move really behave. It seems to be related to old compatibility but it seems to behave in a different manner according to the input.

If one gives a list ==) Take the first element of the list ==) but if the first element is a number (which is accepted here) then the 
`        data_act.name = self.title`
spits out an error
If ones gives a np array, then it gets converted to a DataActuator type and no problem is observed.

In this PR, I updated it to behave in the following manner:
==) if we detect a list, make a np.array from it then we indeed get a DataActuator out
==) if it is a Number (either int/float), we could specify Real, convert it to a np array

This forces the passage to the following line which ascertain that it correctly gets out with all the attributes.
`            data_act = DataActuator(data=[data_act], units=self.units)`

### DIFF
--- a/src/pymodaq/control_modules/daq_move.py
+++ b/src/pymodaq/control_modules/daq_move.py
@@ -616,6 +616,8 @@ class DAQ_Move(ParameterControlModule):
         """
         if isinstance(data_act, list):  # backcompatibility
             data_act = np.array(data_act)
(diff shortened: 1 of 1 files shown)


seb5g:
too many mixu things (formating and code modification) . PR #661  do a proper PR on the subject

PR #633, opened 2025-06-01
Bugfix/PID_add_move_from_extension

Ashwolaa:
When creating a DAQ_move from a setpoint, the function add_move_from_extension is triggered which launches the function:
`        actuator = self.add_move(name, None, instrument_name, [], [], []) (see dashboard.py)`

This triggers a TypeError: 'NoneType' object is not subscriptable as plug_settings is None

This PR solves this issue by only trying to access the settings if plug_settings is not None.

Below is the full error.

  File "D:\Work\PyMoDAQ\src\pymodaq\extensions\pid\pid_controller.py", line 505, in create_setp_actuators
    self.dashboard.add_move_from_extension(setp, "PID", PIDController(self))
  File "D:\Work\PyMoDAQ\src\pymodaq\dashboard.py", line 1207, in add_move_from_extension
    actuator = self.add_move(name, None, instrument_name, [], [], [])
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\Work\PyMoDAQ\src\pymodaq\dashboard.py", line 1136, in add_move
    ui_identifier = plug_settings["main_settings", "ui_type"]
                    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'NoneType' object is not subscriptable

During handling of the above exception, another exception occurred:

  File "D:\Work\PyMoDAQ\src\pymodaq\extensions\pid\pid_controller.py", line 509, in create_setp_actuators
    raise PIDError(
pymodaq.utils.exceptions.PIDError: Could not load the PID extension and create setpoints actuators'NoneType' object is not subscriptable


seb5g (changes requested):
hello,
thanks for this. however you mixed in syntax (a lot of it, thanks Claude I guess?) and the tiny patch object of the PR. Could you create two of them: one for the syntax and then the patch.
As for the patch I think than better than setting ui_identifier to None (as I did also), the best would be to read the defulat GUI type from the configuration file. Could you do that in that other PR?
thx

seb5g:
as there is two commits you could easily create two branches from it to do the PR separately

src/pymodaq/dashboard.py
```diff
@@ -1132,9 +1132,12 @@ def add_move(
...
         actuators_modules,
     ) -> DAQ_Move:
-        try:
-            ui_identifier = plug_settings["main_settings", "ui_type"]
-        except KeyError:
+        if plug_settings is not None:
+            try:
+                ui_identifier = plug_settings["main_settings", "ui_type"]
+            except KeyError:
+                ui_identifier = None
```
seb5g: something like would be better for both None lines!

ui_identifier = config('actuator', 'ui')

seb5g (changes requested):

seb5g (approved):

PR #634, opened 2025-06-01
Readability/daq

Ashwolaa:
Little PR related to the readability of the code. 
It sorts the import and use Ruff style 
https://docs.astral.sh/ruff/formatter/

In addition, its original intent was to initialize self.ui with the expected DAQ_Move_UI_Base attribute (in DAQ_move) as in daq_viewer


seb5g:
we should have a global discussion about what we would like as syntax in pymodaq before changing the code using automatic formating... and then impact it to all modules once we're ok!

PR #646, opened 2025-06-09
Feature/pid upgrade

Ashwolaa:
Hey everyone,

As first introduced in #635, I looked a bit deeper into the PID extension and came up with some "upgrades". In practice, it should be pretty harmless with previous versions starting from 5.1.x_dev so you should be safe to test it without ruining your setup!

 Here is the list of the new functionalities introduced:
- Add a display timer to know how long is spent in the PID loop: this is useful to setup the "best" sample time for your PID
- Add a button in the parameter tree to refresh both plots (at the same time)
- Running loop should be slightly faster now as it only waits enough to match the expected sampling time (time between two measurements)
- There is now a variable that tracks the previous values collected by the PID and storing them as a dict of deque (see collections.deque). There is one deque /setpoint. Its main purpose is to calculate the relative stability of the PID. This stability is calculated from the standard deviation over the elements in the queues. The length of the deque can be changed in the parameter tree. In practice, it is useful to adapt the constants of the PID to your needs as you mostly want to lower this number.

stability

- This deque is now also accesible from the DAQ_move_PID which makes use of it to add a new move_done signal. In pratice, by toggling the check stability in the parameter tree, the move_done_signal (used in a daq_scan for example) will only trigger if the stability is lower than the desired threshold. The deque length is the expected amount of data_points used to calculate the standard deviation. This length cannot be higher than the ones from the PID. The LED from the parameter turns green when it passes the stability_test.

daq_move_PID

- Before, once a DAQ_move_PID was created, it would not be deleted from the dashboard when quitting the model. For this reason, I added a new method in the `dashboard.py` to remove modules from it (actuators/detectors). The DAQ_move_PID are now properly destroyed (with their docks) using this syntax. I believe this is the kind of tasks that could be transfered to the module manager.py at some point (see https://github.com/orgs/PyMoDAQ/discussions/622#discussioncomment-13410542)
- I disabled the close button from the PID window for now. This mostly helps to make sure the program correctly goes through the `quit_fun` (and remove the DAQ_move_PID)

Future developments are foreseen:
1) Related to the speed at which the values are being updated. Currently, everything is decided from the `timerEvent` which makes it dependent on the value given to the `refresh_plot_time`. 
In fact, I already added an additional queue system in the PID_runner so the values could already be used to calculate the relative stability but I need some help from someone (@seb5g?) as I am very very loss on how to manage and properly define the data before sending them in the `process_output`.

2) Currently, any models can be loaded despite not having the correct actuators/detectors to make it work. This is prone to errors, so additional checks should be done to avoid that and only allow to load the models if they match the current state of the dashboard.


seb5g:
did you figure out the issue with your spectrometer?

Ashwolaa:
> did you figure out the issue with your spectrometer?

I got it to work as expected by removing the  `QtWidgets.QApplication.processEvents()` in the `grab_data `of `modules_manager.py`. 
I feel like the processEvent was making some parts of the pid_thread to pursue while the data were not properly collected.

I did not include this change in this PR as it works fine with some Mocks.

seb5g:
but does it make any difference at all for the mocks? If not we should remove it for everything, don't we?

Ashwolaa:
It did not make a difference using the PIDModelBeamSteering and it was required in my setup. I believe it should be safe to do. 

I am all inclined to remove as much as possible these `QtWidgets.QApplication.processEvents()`.

src/pymodaq/dashboard.py
```diff
@@ -1204,20 +1227,20 @@ def save_layout_state_auto(self):
...
-        plug_name,
-        plug_settings,
-        plug_type,
-        move_docks,
-        move_forms,
-        actuators_modules,
+        plug_name: str = None,
+        plug_settings: Parameter = None,
+        plug_type: str = None,
+        move_docks: list[Dock] = [],
```
seb5g: one should avoid passing empty list (mutable) as default argument. This creates many issues... Pass them as None by default and add a line in the method to instantiate it to an empty list of the default is None

seb5g (changes requested):

src/pymodaq/dashboard.py
Ashwolaa: Indeed, I was a bit overzealous in the arguments to avoid possible bugs if a list was not being given but this is safer

src/pymodaq/dashboard.py
```diff
@@ -297,6 +297,132 @@ def add_status(self, txt):
         except Exception as e:
             logger.exception(str(e))
 
+    def remove_detectors(self, detector_modules: List[DAQ_Viewer] = None):
```
seb5g, line 300: why do you need this for?

src/pymodaq/dashboard.py
```diff
@@ -297,6 +297,132 @@ def add_status(self, txt):
...
+                if dock:
+                    dock.close()
+                dock = self.dockarea.docks.get(f"{detector_module.title} viewer", None)
+                if dock:
+                    dock.close()
+            self.update_module_manager()
+        except Exception as e:
+            logger.exception(str(e))
+
+    def remove_actuators(self, actuator_modules: List[DAQ_Move] = None):
```
seb5g, line 327: why do you need this for?

src/pymodaq/dashboard.py
seb5g: is it for a cleaner quit_fun?

seb5g (changes requested):
could you explain still the laset questions? I will try to run your branch locally to make sure evertyhing is ok. But again for next time please separate things. Here dashboard stuff (remove, ...) is nice refactoring but mixed in with PID stuff?? If we make a mistake in the review, both work will be lost while one could still be valid...

src/pymodaq/dashboard.py
Ashwolaa: Yeah I understand, I was initially not willing to edit the dashboard and I added them in separate commits to disentangle it but this brings some confusion. So far, only the PID makes use of it so it should be safe for the rest of the code.

The reason for including it is indeed related to the `quit_fun` of the PID_controller.
When ones create a PID actuator (see `create_setp_actuators `in the pid_controller.py), a new DAQ_move was added to the dashboard. However, when one would leave the PID module, the DAQ_move would not be deleted from the dashboard. This was possibly leading to some conflicts if one would stop/restart the PID module.

I implemented these functions to treat this case in a more general frame and in addition to extend some of the control over the dashboard.

Ashwolaa: We only need it for the actuators but it was too tempting to not make it for the detectors at the same time.

seb5g: :-DD

seb5g:
alright I looked at it with my beamsteering mock and it looks quite nice. The stability feature is really good!!!! I have two things to ask from you now: could you update the documentation in this PR concerning the PID with the features you introduced? And would you be willing next october to present this (all the PID stuff + your addons) at the pymodaq days?

Ashwolaa:
> alright I looked at it with my beamsteering mock and it looks quite nice. The stability feature is really good!!!! I have two things to ask from you now: could you update the documentation in this PR concerning the PID with the features you introduced? And would you be willing next october to present this (all the PID stuff + your addons) at the pymodaq days?

I am happy that it suits you !
I can answer yes to both questions.

seb5g:
@Ashwolaa let me know when this is ready for final review

malik-irain:
@Ashwolaa Is it ready for final review now?

Ashwolaa:
> @Ashwolaa Is it ready for final review now?

I made some mistakes in the merging and struggled to get it back to where I wanted. I made a fresh branch from a safe position to have a clean PR #703

PR #658, opened 2025-06-25
Very basic handling of running presets directly

astroloba:
Added basic command-line handling to `dashboard` : the (optional) first command-line argument can be used to specify a preset file:
`$ dashboard /etc/.pymodaq/preset_configs/preset_default.xml`

### DIFF
--- a/src/pymodaq/utils/gui_utils/loader_utils.py
+++ b/src/pymodaq/utils/gui_utils/loader_utils.py
@@ -8,7 +8,7 @@ from pymodaq.utils.config import get_set_preset_path
 from pymodaq.extensions.utils import CustomExt
 
 
-def load_dashboard_with_preset(preset_name: str, extension_name: str="") -> \
+def load_dashboard_with_preset(preset_name: str, extension_name: str = "") -> \
         (DashBoard, CustomExt, QtWidgets.QMainWindow):
     """ Load the Dashboard using a given preset then load an extension
 
--- a/src/pymodaq/utils/gui_utils/loader_utils.py
+++ b/src/pymodaq/utils/gui_utils/loader_utils.py
@@ -8,7 +8,7 @@ from pymodaq.utils.config import get_set_preset_path
 from pymodaq.extensions.utils import CustomExt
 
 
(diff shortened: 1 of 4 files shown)


malik-irain:
There is a `load_dashboard_with_preset` function, could you try to use it?

https://github.com/PyMoDAQ/PyMoDAQ/blob/5.0.x/src/pymodaq/utils/gui_utils/loader_utils.py#L11-L31

Also to handle command-line arguments, [argparse](https://docs.python.org/3/library/argparse.html) would be nice with the `-p` or `--preset` argument

astroloba:
These two additional commits should do the job.
Note that because of an import conflict (`loader_utils.py` already imports `DashBoard`), i had to move the `main()` code from `dashboard.py` to a new `dashboard_loader.py`. I modified the `pyproject.toml` file so that the `dashboard` command starts `dashboard_loader:main`.

astroloba:
...and finally using `argparse` to implement a `-p` (alias `--preset`) command-line argument:
`$ dashboard -p preset_default`

src/pymodaq/utils/gui_utils/loader_utils.py
```diff
@@ -8,7 +8,7 @@
 from pymodaq.extensions.utils import CustomExt
 
 
-def load_dashboard_with_preset(preset_name: str, extension_name: str) -> \
+def load_dashboard_with_preset(preset_name: str, extension_name: str="") -> \
```
malik-irain: a space before and after the equals sign would be perfect!

astroloba: Done! Commit [36fcba2](https://github.com/PyMoDAQ/PyMoDAQ/pull/658/commits/36fcba2e8a71334624f2b997f05064d0dbd458f5)

malik-irain:
Sorry, one last thing I forgot about: It is a new feature so it should be on branch 5.1.x_dev. I tried to modify the merge target in this PR but it adds unwanted modifications.

It would be great if you could create a branch on 5.1.x_dev, then cherry-pick your commits into it and open a new PR

PR #632, opened 2025-05-31
Feature/host argument for leco director

Ashwolaa:
DAQ_move_LECODirector and DAQ_viewer_LECODirector now accepts explicitely the host argument during initialization.

 If host argument is given, the settings is updated with the given value:
```
        if host is not None:
            self.settings["host"] = host
        LECODirector.__init__(self, host=self.settings["host"])

```
This permits to create an instance from an external script (not pymodaq) to change / request the value of an initialized DAQ_move / DAQ_viewer 
Related to #547 and #557

PR #659, opened 2025-06-27
Allow actor/director to connect to arbitrary coordinator host/port

ccabello99:
I was having issues connecting my actor to a coordinator running on another computer or with an arbitrary port number. I found that we never actually passed the host and port that a user inputs in the GUI to the PymodaqListener during actor initialization. 

I added a function that will take these values and pass them to the listener class when the connect_leco() method is called.

### DIFF
--- a/src/pymodaq/control_modules/utils.py
+++ b/src/pymodaq/control_modules/utils.py
@@ -469,13 +469,25 @@ class ParameterControlModule(ParameterManager, ControlModule):
             name = self.settings.child("main_settings", "leco", "leco_name").setValue(name)
         return name
 
+    def get_leco_host_port(self) -> tuple:
+        host = self.settings["main_settings", "leco", "host"]
+        port = self.settings["main_settings", "leco", "port"]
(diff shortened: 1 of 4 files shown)


Ashwolaa:
I suggest adding these two changes on this PR (taken from #632) on the LECO director for the viewer and the actuator.

image
image

malik-irain:
It works correctly for actors, could you add the changes mentioned by @Ashwolaa and  a port setting on LECODirectors?

malik-irain:
@seb5g It's the same as #689. Should we also accept that on 5.0? I don't consider it as a real feature and more of a fix as it's just "fixing" the GUI to allow for settings that are already taken into consideration in the code, but I'm not sure.

(It's not complete, last commit of #689 is missing for this to fully work, so don't accept if the answer is yes!)

seb5g:
then yes, let me know when done

malik-irain:
should be good now

seb5g (approved):

Issue #668, opened 2025-07-06
[BUG] Deprecation of class properties (python>3.11)


Ashwolaa:
### Describe the bug

Using python 3.13, the dashboard will not be able to handle the following lines when the splash screen is required giving the following error in the log.

2025-07-06 17:37:50,596 - pymodaq.dashboard - ERROR - 'property' object has no attribute 'show'
Traceback (most recent call last):
  File "D:\Work\PyMoDAQ\src\pymodaq\dashboard.py", line 1783, in set_preset_mode
    self.splash_sc.show()
    ^^^^^^^^^^^^^^^^^^^
AttributeError: 'property' object has no attribute 'show'

The splash_sc property is initialized here:
```
@classmethod
@property
def splash_sc(cls) -> QtWidgets.QSplashScreen:
    if cls._splash_sc is None:
        cls._splash_sc = get_splash_sc()
    return cls._splash_sc
```
The reason for it is that Class properties are deprecated in Python 3.11 and will not be supported in Python 3.13 

### Reference(s)

```markdown
dashboard
```

### Environment

```markdown
- OS:Windows 11
- Pymodaq: 5.1
- python env: 3.13
```

seb5g:
I want to reopen it because simply converting it back to a regular method may create a bug somewhere else where the splashscreen was called using the property.

I propose to implement what is discussed in https://github.com/PyMoDAQ/pymodaq_data/issues/49 and use it here. @malik-irain could you do it please?

seb5g:
it's been kept as a property but not as a class one

PR #669, opened 2025-07-06
Bugfix/dashboard python3.13

Ashwolaa:
This PR is related to #668

The class method looks obsolete looking at its use in the dashboard. This PR removes it to make the use of the dashboard compatible with python 3.13

### DIFF
--- a/src/pymodaq/dashboard.py
+++ b/src/pymodaq/dashboard.py
@@ -1800,7 +1800,7 @@ class DashBoard(CustomApp):
                     severity="critical",
                     title="Preset loading error",
                     text=f"""
-                            <p>{error}<\p>
+                            <p>{error}</p>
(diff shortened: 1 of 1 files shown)


src/pymodaq/dashboard.py
```diff
@@ -259,12 +259,11 @@ def __init__(self, dockarea):
             if self.check_update(show=False):
                 sys.exit(0)
 
-    @classmethod
     @property
```
seb5g, line 262: is it never called from a class and only from an instance? don't remember why I coded it like that but there should be a reason...

Ashwolaa: Could be but I did not find it ^^

seb5g (approved):

PR #674, opened 2025-07-14
Doc/2.1.3.2 execute a given python file

YannickDM09:
I rewrote the section 2.1.3.2 of the doc ! I tried to explain simply what makes it possible to launch PyMoDAQ's modules using a console.


malik-irain:
It seems you made your commit on branch 5.0.x on your side and tried to do a PR on branch 5.1.x_dev. It creates a lot of commits and tries to merge other files. Could you redo that but on a branch based on 5.1.x_dev?
```bash
# create the new branch
git checkout -b doc/2.1.3.2_execute_a_given_python_file_5.1 5.1.x_dev
# cherry-pick the commit to apply to this branch
git cherry-pick fcbe676
```

Regarding the content now, the image and explanation are more related to  section `2.1.3.1. Load installed scripts`. Project scripts can execute any function (even though it is often `main`) but `python -m <module>` will necessarily start the main function.

Also, I think 
>All PyMoDAQ’s modules are Python files with an entry point so each of them can be launched directly with your console

should be:
> Most PyMoDAQ's modules ...

YannickDM09:
Thank you for your help and your recommendations ! I realized the commands you sent and if I understood correctly, "git checkout -b" creates a new branch based on the branch 5.1.x_dev and switch automatically on this branch.
"git cherry-pick" copy and paste the chosen commit on the new branch.

Also, I don't understand why the commit was made on the branch 5.0.x because I created my first branch (doc/2.1.3.2_execute_a_given_python_file) from the 5.1.x_dev branch and when I made the commit I was on that same first branch (using PyCharm). I did the "git remote set-url origin" command (following the tutorial guide 3.5 How to contribute to PyMoDAQ’s documentation?) to link my local repository to my remote repository but is there a command to specify the branch where we want to commit ?

malik-irain:
Your default branch is 5.0.x so when creating a new branch without specifying the base (`git checkout -b <branch_name>`) it use the default as base.

I don't think `git remote set-url origin` is necessary if you cloned the repo but it won't hurt. For branches, you can have a different name between your local copy and remote one but it will still keep the same base!

YannickDM09:
Aah ok ! I see !
I will do the modifications and try to do a pull request again. Thanks !

PR #678, opened 2025-07-17
Improvements in LECO parts

BenediktBurger:
Some improvements from #405 

Also: added Enums of rpc methods available, in order to link a request with the associated method.
Test ensure, that these methods are available via rpc.

The enum names are open for discussion.


BenediktBurger:
some review?

src/pymodaq/utils/leco/daq_move_LECODirector.py
```diff
@@ -123,7 +123,6 @@ def move_home(self):
 
     def get_actuator_value(self) -> DataActuator:
         """ Get the current hardware value """
-        self.controller.set_remote_name(self.communicator.full_name)  # to ensure communication
```
seb5g, line 130: why remove it? is the communication checked elsewhere?

src/pymodaq/utils/leco/leco_director.py
```diff
@@ -13,18 +13,19 @@
 
 from pymodaq.utils.leco.director_utils import GenericDirector
 from pymodaq.utils.leco.pymodaq_listener import PymodaqListener
+from pymodaq.utils.leco.rpc_method_definitions import GenericDirectorMethods, MoveDirectorMethods
 from pymodaq_utils.serialize.factory import SerializableFactory
 from pymodaq.control_modules.thread_commands import ThreadStatusMove
 
 config = Config()
 
 class DirectorCommands(StrEnum):
```
seb5g, line 26: with the definition of the enums in the rpc_method_definition, this class seems overkill, should be not remove it?

src/pymodaq/utils/leco/pymodaq_listener.py
```diff
@@ -16,8 +17,17 @@
...
+    MoveMethods,
+    ViewerMethods,
+    GenericDirectorMethods,
+    MoveDirectorMethods,
+    ViewerDirectorMethods,
+)
 
 
+# Commands for ThreadCommand
 class LECOClientCommands(StrEnum):
```
seb5g, line 31: is this class still needed with the proper definition of the enums in rpc_method_definition?

src/pymodaq/utils/leco/pymodaq_listener.py
```diff
@@ -185,7 +206,7 @@ class PymodaqListener(Listener):
     :param host: Host name of the communication server.
     :param port: Port number of the communication server.
     """
-    remote_name: str = ""
+    remote_names: set[str]
```
seb5g, line 209: why a set here?

seb5g: is it for testing the remote is still connected later?

src/pymodaq/utils/leco/pymodaq_listener.py
```diff
@@ -272,52 +299,64 @@ def queue_command(self, command: ThreadCommand) -> None:
             finally:
                 self.cmd_signal.emit(ThreadCommand('disconnected'))
 
+        elif command.attribute is None:
```
seb5g: its a bit dangerous to put it here like this as it means that all methods below have a use of attribute. The commands that need an attribute should maybe be nested  in such a if case like 

```
elif command.attribute is not None:  # add below all methods requiring the presence of an attribute
    if command.command == LECOViewerCommands.DATA_READY:
        ...
else:
    raise IOError("Unknown TCP client command")

```

src/pymodaq/utils/leco/leco_director.py
seb5g: I think I mixed enums for allowed methods and allowed ThreadCommands as some are sharing the same enum value...

seb5g (commented):
some questions on a few points but this looks quite thorough. The enum names are clear

src/pymodaq/utils/leco/pymodaq_listener.py
BenediktBurger: Currently, an actor module allows only one director module. Once other director module tries to connect (in order to view some data etc.), the first director won't get any data anymore, as the remote_name has been changes.

With these changes, we can add as many different directors to an actor as you wish. The `set` has the advantage, that every entry appears just once, so no duplication.

src/pymodaq/utils/leco/pymodaq_listener.py
seb5g: I see, that's very good!!

src/pymodaq/utils/leco/pymodaq_listener.py
seb5g: I see that's very good!

BenediktBurger: I'm not sure whether it is good to add the data protocol to PyMoDAQ and used my draft of a "event based transfer via control protocol" in pyleco to enhance PyMoDAQs usage.

src/pymodaq/utils/leco/daq_move_LECODirector.py
BenediktBurger: I had to set the remote name at each get value, as someone else might have changed the remote name in the mean time.
Now with the `set` of remote names (see below), this problem does not arise anymore and it is sufficient to set the remote name at stage initialization.

src/pymodaq/utils/leco/leco_director.py
BenediktBurger: You are right, that this class is not needed anymore.
I kept it for backward compatibility, but linked it to the proper definitions.
If you think, that nobody will use it, we can remove that class as well.

Also for the class below.

src/pymodaq/utils/leco/pymodaq_listener.py
BenediktBurger: These are now `ThreadCommand` contents, while the Director classes where RCP method names.

src/pymodaq/utils/leco/pymodaq_listener.py
BenediktBurger: Changed.

seb5g (approved):

src/pymodaq/utils/leco/pymodaq_listener.py
seb5g: Should we leave it as is for the moment and have a discussion at the pymodaq days to talk about it? Could you prepare a light summary of what would be the pros and cons?

seb5g:
is it ready for a merge?

src/pymodaq/utils/leco/pymodaq_listener.py
BenediktBurger: Sorry, this was meant as a comment, how I got these improvements here.

It should be merged as is. Using the data protocol could come later, independently.

BenediktBurger:
Yes, it is ready for merge. I needed to resolve the merge conflict.

Issue #307, opened 2024-05-28
PyLECO event based feedback


BenediktBurger:
I propose to use the LECO data protocol to tell the dashboard plugins about changes in the remote plugin.

## Issue
LECO offers two communication channels: The control protocol (currently used) for exchanging messages.
This works great for requests and their responses.
However, if a remote plugin wants to tell about a change (e.g. settings change) by its own accord, it has to know the name of the plugin in the dashboard.
This (currently implemented) version requires that the remote plugin always knows, whom to inform about changes (or position reached or new values...). It also limits the contact to one controlling plugin.

## Proposal
LECO offers also another protocol (the data protocol) for a broadcasting style, which is great for event propagation.

Implementation:
* This requires another style of Coordinator to be run as well, the `proxy_server`.
* The dashboard plugins use their Listener (already present and capable) to subscribe to changes in the remote plugin
* The remote plugin publishes (broadcasts) different events (easy to add to the present event handling)
* The dashboard plugins have to react to the events arriving to the Listener (should be quite simple)

seb5g:
For sure this is the next step after the initial implementation you made as for the moment the remote control module cannot update the dashboard "Fake" control modules. Do you plan to work on that soon?

BenediktBurger:
It's itching me to do that, but I have to postpone it.
That the idea is not lost, I opened this issue.

I might get back to it, if I need a break.

BenediktBurger:
I tried to understand how everything works, here is my summary as a flowchart:
~~~mermaid
flowchart LR
    subgraph main_thread[main module thread]
        subgraph Listener
            queue_commandL[[queue_command]]
            cmd_signal([cmd_signal])
        end
        subgraph DAQ_Move["DAQ_Move(ParameterControlModule, ParameterManager, ControlModule, QObject)"]
            init_hardware[[init_hardware]]
            command_hardware(["command_hardware(ThreadCommand)"])
            _update_settings_signal(["_update_settings_signal(edict)"])
            _command_tcpip(["_command_tcpip(ThreadCommand)"])
            status_sig(["status_sig(str)"])
            process_tcpip_cmds[["process_tcpip_cmds"]]
            thread_status[[thread_status]]
        end
    DAQ_Move===>|"contains('ui')"|DAQ_Move_UI
    end

    subgraph hardware_thread[hardware thread]
        subgraph DAQ_Move_Hardware["DAQ_Move_Hardware(QObject)"]
[...]

PR #683, opened 2025-07-19
Feature/more robust leco directors

BenediktBurger:
The director checks every second (time is up to debate, can be made a setting), whether the actor is reachable. If not, it disables the "stage". This is to give feedback, when the connection is down and you have to reconnect (and tell the actor about your name again).

Belongs to #307 as it allows more robust remote handling.


BenediktBurger:
Someone to review?

seb5g:
on it!

src/pymodaq/utils/leco/leco_director.py
```diff
@@ -90,6 +92,26 @@ def close(self) -> None:
         self.settings.child('settings_client').clearChildren()
         self.listener.stop_listen()
 
+    def start_timer(self) -> None:
+        """To be called in child classes."""
+        self.timer = QTimer()
+        self.timer.timeout.connect(self.check_actor_connection)
+        self.timer.start(1000)  # in seconds
```
seb5g: maybe add this timeout in the main leco config file for user to find it and modify it easily? it needs to be added in the pymodaq_utils config:

image

then can be read as:
```
try:
    timeout = config('network', 'leco-server', 'check-timeout')
except KeyError:
    timeout = 1000
self.timer.start(timeout)
```

the try to make sure it works even if the pymodaq_utils config is not uptodate

seb5g (changes requested):
just a change on the timeout value to be added into the configuration file

src/pymodaq/utils/leco/leco_director.py
BenediktBurger: Done. pymodaq_utils PR is coming.

seb5g (approved):

PR #684, opened 2025-07-21
Improve typing

BenediktBurger:
Fix some wrong typings and add some more.

### DIFF
--- a/src/pymodaq/control_modules/move_utility_classes.py
+++ b/src/pymodaq/control_modules/move_utility_classes.py
@@ -657,7 +657,6 @@ class DAQ_Move_base(QObject):
     def close(self) -> None:
         raise NotImplementedError
 
-    @abstractmethod
     def move_abs(self, value: Union[float, DataActuator]):
         if hasattr(self, 'move_Abs'):
             deprecation_msg('move_Abs method in plugins is deprecated, use move_abs', 3)
@@ -665,7 +664,6 @@ class DAQ_Move_base(QObject):
         else:
             raise NotImplementedError
 
-    @abstractmethod
     def move_rel(self, value: Union[float, DataActuator]):
         if hasattr(self, 'move_Rel'):
             deprecation_msg('move_Rel method in plugins is deprecated, use move_rel', 3)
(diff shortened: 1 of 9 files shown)


BenediktBurger:
some review?

src/pymodaq/control_modules/move_utility_classes.py
```diff
@@ -637,35 +645,39 @@ def check_bound(self, position: DataActuator) -> DataActuator:
...
         else:
             raise NotImplementedError
 
-
-    def close(self):
+    @abstractmethod
+    def close(self) -> None:
         raise NotImplementedError
 
+    @abstractmethod
```
seb5g: adding the decorator here would prevent the call of the old move_Abs method.

src/pymodaq/control_modules/move_utility_classes.py
```diff
@@ -637,35 +645,39 @@ def check_bound(self, position: DataActuator) -> DataActuator:
...
     def move_abs(self, value: Union[float, DataActuator]):
         if hasattr(self, 'move_Abs'):
             deprecation_msg('move_Abs method in plugins is deprecated, use move_abs', 3)
-            self.move_Abs(value)
+            self.move_Abs(value)  # type: ignore
         else:
             raise NotImplementedError
 
+    @abstractmethod
```
seb5g: adding the decorator here would prevent the call of the old move_Rel method.

src/pymodaq/control_modules/move_utility_classes.py
```diff
@@ -637,35 +645,39 @@ def check_bound(self, position: DataActuator) -> DataActuator:
...
     def move_rel(self, value: Union[float, DataActuator]):
         if hasattr(self, 'move_Rel'):
             deprecation_msg('move_Rel method in plugins is deprecated, use move_rel', 3)
-            self.move_Rel(value)
+            self.move_Rel(value)  # type: ignore
         else:
             raise NotImplementedError
 
+    @abstractmethod
```
seb5g: adding the decorator here would prevent the call of the old move_Home method.

seb5g (changes requested):
some use of the abstract decorator should be removed

BenediktBurger:
I removed them. Thanks for the review.

PR #694, opened 2025-08-28
Feature/fix badge

### LINKED ISSUE #693: [BUG] Skip badge on PR from fork does not work
Reopening of #679 .

It does not work as expected, there are still

BenediktBurger:
Might close #693 

The owner is considered, where the workflow is run: In a PR, it is the target repository, i.e. Pymodaq itself.

Here, we check for the fork.


BenediktBurger:
As we see: this PR does not create the badge step, but after merge, the push to main should create it.

Issue #700, opened 2025-09-09
[BUG] ROI/Preset creation do not trigger the connection with the proper slot


seb5g:
### Describe the bug

When creating a new preset or a new ROI in the dashboard, the menu is updated but the action does not trigger the corresponding slot. When the dashboard is restarted, it works!

### Expected Behavior

The action creation should allow the immediate triggering of the slot connected to the action!

PR #701, opened 2025-09-09
add the right action/slot connection just after action creation

seb5g:
solve #700


malik-irain:
It's on 5.0.x, don't forget to port it to 5.1.x_dev (unless it's already fixed there)

PR #703, opened 2025-09-11
Feature/pid upgrade+doc

Ashwolaa:
This PR is the updated version of  #646.
The reason for this new PR is that  I ***ed up my last merging with the last changes of Pymodaq_5.1.x_dev, by not updating my local version wrt to the remote. 
Anyway,  I decided to start a new branch before the merging to avoid possible conflicts/bugs.

This PR should be good for review, I also updated the docs with the new buttons/options.


seb5g (approved):

seb5g:
I accepted your PR. As you are the one using this extension the most. Make sure to patch anything needed. It was a huge PR and I'm not sure I could pinpoint everything... 

I tested it and it seems it's doing what it should. However, a few time, using the beam_steering model, I obtained nan values (I guess this is when the beam is out of the camera screen) but then one should maybe add some fake value in that case to keep the system stable. Could you investigate?

seb5g:
Are you going to use the beam steering as demo at tje pymodaq days?

Ashwolaa:
Okay, I will try to be reactive if there are some issues from users on the discord. I already want to discuss with you for an additional patch but this can wait for the pymoday's as it is about the managing correctly the data class objects.

The issue with the beam steering comes from the scipy that calculates the center of mass. I think it triggers when the beam gets out of the screen.
 I looked a bit into it but I struggled to find a quick fix and I did not spend additional time but I will give it another shot.

I plan to show either the beam steering or my custom plugin for stabilizing some spectral fringes.

seb5g:
yes please try to stabilise the beam steering example as this is the one we describe in the doc and the one people will try and mess with

PR #704, opened 2025-09-11
Feature/add units to scan

seb5g:
Units are everywhere in the control modules but where not enforced within the daq_scan extension. This PR add in each scanner setting the correct suffix corresponding to the native actuator unit

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ classifiers = [
 ]
 dependencies = [
     "pymodaq_utils >= 1.0.2", #use strictly this version (yanked) for the moment until we found a better way to handle versioning
-    "pymodaq_gui>=5.0.23",
+    "pymodaq_gui>=5.0.25",
     "pymodaq_data>=5.0.21",
     "easydict",
     "multipledispatch",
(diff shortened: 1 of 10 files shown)


seb5g:
all tests pass, @malik-irain maybe an external view will show some missing bits...

malik-irain:
It's working but a few improvements could be made:

- Units handling options "Display units" and "Use same units" are mutually exclusive but it is not so clear why. Anyways, we could achieve the same simply by keeping "Use same units" with a mouseover text/tooltip saying that activating it disable displaying units, or using a list of options to choose but not display both. The tooltip solution is not trivial to implement, as it needs to be added manually after the parametertree is instanciated, but it could be the object of another PR to add tooltips to GUI elements.

- When choosing "Use same units", "Common units" text field is editable but when changed it doesn't do anything, if it is the wanted goal it could be defined with `'readonly' : true` ?

seb5g:
> It's working but a few improvements could be made:
> 
>     * Units handling options "Display units" and "Use same units" are mutually exclusive but it is not so clear why. Anyways, we could achieve the same simply by keeping "Use same units" with a mouseover text/tooltip saying that activating it disable displaying units, or using a list of options to choose but not display both. The tooltip solution is not trivial to implement, as it needs to be added manually after the parametertree is instanciated, but it could be the object of another PR to add tooltips to GUI elements.

indeed we could have only one boolean. Tooltip is easy to add in the settings.setOpts(tip='blabla')
> 
>     * When choosing "Use same units", "Common units" text field is editable but when changed it doesn't do anything, if it is the wanted goal it could be defined with `'readonly' : true` ?

I guess it could be readonly indeed... wanted to allow the user to change it but then I guess it's better no to and just use the regular units defined in the instrument plugin

seb5g:
I addressed your comments, I think this is ready for a merge!

PR #708, opened 2025-09-15
Feature/pid upgrade 2

Ashwolaa:
Small patches to complete PR#703
- Now the queues used to calculate the stability are taken from the PID runner instead of the process_output (which was only updated when plotting) ==) much faster convergence
- Adding the possibility for a weighted calculations on the stability (last values have more impact than first ones)
- Fixing other small bugs (sample time not being updated / update_runner_setpoints not properly called / ...)
- Some factorizing


seb5g:
I'm not sur I understand. You mentioned the queue are used from the pidrunner but I can still see it in the process_output. Also I don't understand why you don't use queue object meant for this: https://docs.python.org/fr/3/library/queue.html
They are quite appropriate for threading producer/consumer.

Ashwolaa:
> I'm not sur I understand. You mentioned the queue are used from the pidrunner but I can still see it in the process_output. Also I don't understand why you don't use queue object meant for this: https://docs.python.org/fr/3/library/queue.html They are quite appropriate for threading producer/consumer.

There are two queues:
- one in the pid_runner which is just getting filled through each iteration of the pid_runner loop
- one in the DAQ_PID which is being filled by all the accumulated values from the PID_runner. This queue is then used in the process_output to calculate the stability.

In the previous implementation, the one from the pid_runner was implemented but not used (as I struggled on how to properly manage DataRaw type).

I did not pay much attention in the Queue object, the deque from collections just felt more versatile all in all and is also thread-safe. I am not sure that there is much change in terms of performances.

seb5g:
ok let's leave it like this if there is no issues with thread safety

seb5g (approved):

PR #711, opened 2025-09-18
Name for scan and logger windows

### LINKED ISSUE #369: MIssing a proper window name
When opening windows from *dashboard*, i.e. for instance opening Scan window with "Extensions > Do scans",
the window has its name set to dashboard, and so will name of other such windows.

This makes navigation between windows a bit confusing, and a name like "Scan" or such (and oither specific name for other windows) would be more convenient.

Tested with pymodaq 4.3.0 under windows OS.

malik-irain:
Closes #369

### DIFF
--- a/src/pymodaq/dashboard.py
+++ b/src/pymodaq/dashboard.py
@@ -450,6 +450,7 @@ class DashBoard(CustomApp):
             win = QtWidgets.QMainWindow()
         area = DockArea()
         win.setCentralWidget(area)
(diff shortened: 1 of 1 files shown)


seb5g:
malik your pr points to pymodaq 5.0.x, but is based on 5.1.x

malik-irain:
Corrected!

seb5g (approved):

PR #712, opened 2025-09-18
Feature/use group parameter nested structure

Ashwolaa:
In this PR, I updated the preset_manager_utils.py to benefit from the nested structure available from GroupParameter
https://github.com/PyMoDAQ/pymodaq_gui/pull/88

This allows to easily sort the plugin according to their type:
I decided to use three subtypes:
- Remote (lecodirector/TPCserver)
- Mock (every plugin with Mock in its name)
- Plugin (all the "true" plugins)

This allows the user to easily find the relevant plugins. This could later be enhanced to allow the user to make his own listing

For now the sorting is done by the name of the plugin, but this could instead be through an additional keyword in the plugin itself which would sort it accordingly.

https://github.com/user-attachments/assets/bc9b9894-b636-4f5d-928e-258e14ed18cb


seb5g (approved):
it's very good!!!

PR #718, opened 2025-09-25
Using class attributes instead of instances attributes

runrunm:
Hello PyMoDAQ,

I was experiencing issues with the dashboard using PySide6:
- **File → Restart** not working
- **File → Configuration file** or **Preset Modes → New Preset** windows not opening

I found out issue was being taken care of in pymodaq_gui [here](https://github.com/PyMoDAQ/pymodaq_gui/issues/63). This PR is for PyMoDAQ repository and I only checked [dashboard.py](https://github.com/PyMoDAQ/PyMoDAQ/blob/5.0.x/src/pymodaq/dashboard.py) file here. A similar issue is most likely to appear somewhere else (extensions or  but these changes should fix **Restart** and **New preset** buttons in dashboard.

Cheers,

Bastien

### DIFF
--- a/src/pymodaq/dashboard.py
+++ b/src/pymodaq/dashboard.py
@@ -819,10 +819,11 @@ class DashBoard(CustomApp):
             mssg.setText('You have to restart the application to take the'
                          ' modifications into account!')
             mssg.setInformativeText("Do you want to restart?")
-            mssg.setStandardButtons(mssg.Ok | mssg.Cancel)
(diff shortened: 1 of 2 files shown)


seb5g:
why did you closed it?

runrunm:
I was not working in dev branch and I noticed part of the issue was already solved there (5.1.x_dev)...

I noticed a similar issue in pymodaq_plugin_manager though so I will make a PR for this one.

PR #721, opened 2025-09-26
Bugfix/close extensions

### LINKED ISSUE #720: Can't close the logger
When opening the logger and trying to close it directly after it fails to close, leaving with a dangling windows with the following error:

```
ERROR:pymodaq.daq_logger:'NoneType' object has no attribute 'close'
  File "/home/mairain/original-pymodaq/pymodaq/src/pymodaq/extensions/daq_logger/daq_logger.py", line 183, in quit_fun
    self.logger.close()
    ^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'close'
```

malik-irain:
Fixed closing the logger using the internal close button and the automatic close of all the extensions when the main dashboard is closed.

Fixes #720


seb5g (changes requested):
Your PR is fine, but we should add in the CustomApp base class an abstract class called quit_fun to make sure extensions implement it! Could you add this?

seb5g:
@malik-irain could you port this into the monorepo and apply it with the requested modification (adding a quit_fun abstract kind method)

malik-irain:
There's still some work to handle everything completely. I keep it open for now, but I'll go back to it later

PR #728, opened 2025-10-04
Factorizing/load module dashboard

Ashwolaa:
- Factorizing extension loading in the dashboard by implementing a new function dedicated to making the window + dock area
- DataMixer was not using the correct method to quit when validate_config was False

### DIFF
--- a/src/pymodaq/dashboard.py
+++ b/src/pymodaq/dashboard.py
@@ -523,7 +523,7 @@ class DashBoard(CustomApp):
                     <p>Please delete the file at <b>{self.datamixer_module.config_path}</b>.</p>
                 """,
             )
-            self.datamixer_module.quit()
+            self.datamixer_module.quit_fun()
         return self.datamixer_module
 
     def load_console(self):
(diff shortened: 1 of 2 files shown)


src/pymodaq/dashboard.py
```diff
@@ -445,94 +446,32 @@ def clear_move_det_controllers(self):
...
-        area = DockArea()
-        win.setWindowFlags(
-            Qt.Window
-            | Qt.WindowTitleHint
-            | Qt.WindowMinimizeButtonHint
-            | Qt.WindowMaximizeButtonHint
-        )
-        win.setCentralWidget(area)
-        win.setWindowTitle("Scanner")
+    def load_scan_module(self):
```
seb5g: to be sure of backcompatibility, add the default win argument as None in the method signature. Remark valid for all load methods

seb5g (changes requested):
changes to apply to methods signature for backcompatibility

seb5g (approved):

PR #735, opened 2025-10-10
Load last preset option

loicguilmard:
Added 3 entries in the [user] section of the pymodaq toml :
loadlast = true
lastloadok = true
lastpreset = "preset_name"

Added a checkbox in the preset menu, to activate default loading last preset

Testing parameters, file existence and proper loading of the preset to change config value accordingly.

Bonus : added a default to None for extension when loading a dashboard from command line with preset as an argument

### DIFF
--- a/src/pymodaq/dashboard.py
+++ b/src/pymodaq/dashboard.py
@@ -5,6 +5,7 @@ import sys
 import datetime
 import subprocess
 import logging
+from functools import total_ordering
 from pathlib import Path
 from importlib import import_module
 from packaging import version as version_mod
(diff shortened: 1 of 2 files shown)


seb5g (changes requested):
nice!
Could you change the config names with underscore for better readability?

malik-irain:
With the last modifications on how presets are managed, it is hard to port this code to the last version of PyMoDAQ.  @loicguilmard can you restart this PR with:
- names with underscores (`load_last`, `load_last_ok` and `last_preset`)
- creating your branch from the dev branch so it can be the PR target

PR #737, opened 2025-10-13
Doc/action manager

Ashwolaa:
Adding doc for action manager and parameter manager (with inclusion of https://github.com/PyMoDAQ/pymodaq_gui/pull/100)


docs/src/api/utility_api/parameter/parameter_manager.rst
```diff
@@ -0,0 +1,198 @@
+Parameter Manager
+==================
+
+Overview
+--------
+
+The ``ParameterManager`` inherits from the ActionManager.It provides a user interface for managing hierarchical parameter configurations with search, save/load capabilities, and a collapsible toolbar.
+
+Quick Start
+-----------
```
seb5g: should say it's a layer on top of the Parameter and ParameterTree helping classes to handle changes in the settings and providing a unique interface for all modules in pymodaq dealing with Parameters

docs/src/api/utility_api/parameter/parameter_manager.rst
```diff
@@ -0,0 +1,198 @@
...
+            else:
+                advanced.hide()
+
+Accessing Values
+~~~~~~~~~~~~~~~~
+
+.. code-block:: python
+
+    # Get value
+    mode = manager.settings.child('settings', 'mode').value()
```
seb5g: there is also the shortcut 
manager.settings['settings', 'mode']

docs/src/developer_folder/managers.rst
```diff
@@ -4,12 +4,791 @@ Managers and Mixin Objects
...
+    print(self.actions_names)
+    
+    # Get all action objects
+    all_actions = self.actions
+
+Best Practices
+--------------
+
+1. **Organize actions logically**: Group related actions together in your ``setup_actions`` method
+2. **Use meaningful short names**: Choose clear, descriptive short names for easy reference
```
seb5g: Best use case is to set for a particular application the action names within a StrEnum (I started to do so in some modules), that avoids mistakes when writing strings while you have automatic autocompletion with a StrEnum

seb5g (changes requested):
Very nice and thorough documentation. A few suggestions in the review to address

docs/src/developer_folder/managers.rst
Ashwolaa: Do you have an example that I can refer to?

seb5g: Look at https://github.com/PyMoDAQ/PyMoDAQ/blob/5.1.x/src/pymodaq/extensions/optimizers_base/optimizer.py line 59

Ashwolaa: that's sweet, both in the implementation and the calling

seb5g:
Be carfull to target 5.2.x_dev now (but soon we'll have the monorepo)

seb5g:
I cherry picked your PR in 5.2.x_dev!

PR #748, opened 2025-10-30
Feature/configurator extension

seb5g:
This large PR introduces the Configurator: a manager to create configurations: list of settings for control modules in the Dashboard. 
Allows also to predefine absolute "positions" for configured Actuator

It simplifies also the Preset Manager

A simplification and refactorisation of all managers ill follow

https://github.com/user-attachments/assets/5a6c31c2-f661-4cb3-8792-22e0a53e42a9


malik-irain:
It complains about `find_dicts_in_list_from_key_val` missing from `pymodaq_utils.utils`, isn't the associated PR from utils missing?

Ashwolaa:
I am not at home/work until Sunday evening.
I will be able to take a good look at it next week

seb5g:
I solved the reuqirement issues! @malik-irain a few test don't pass... as you noticed earlier

malik-irain:
A few comments on usage:

- Preset manager starts with `preset_default` selected but not showing anything (I first though it was loading the Configurator). And we can't rename a preset anymore?
-  I don't know if it was done in this PR but the preset manager remove too much configuration options
   - For example, I can't make a preset that would connect automatically to a LECO Actor.
- When adding an actuator value
   - The window could be smaller   
   - Units are always prefixed with `m`, even when they already have a prefix, i.e. Mock actuators for Power (in `mW`) and Xaxis (in `µm`)  are respectively in `mµW` and `mµm`. They work as they name say, 1000 `mµW` = 1 `µW`. They are showed as their normal units (`mW` and `µm`) in the table though.
- Some configuration entries can be selected for configuration but they have no effect, so they shouldn't be modifiable. For example: Axes, Show Config, ...
- Parameter tree "Title nodes" (I don't know how they're called but "Actuator", "LECO Options", etc.) that don't contain value can be picked  for configuration, they fallback to the first configurable entry in the tree.
- A `del` key shortcut, or a button to remove the selected entry could be nice

Also I got this error but nothing crashed and I don't know what I did to trigger it:

```python
  File "/home/mairain/.virtualenvs/original-pymodaq/lib/python3.12/site-packages/pyqtgraph/parametertree/Parameter.py", line 785, in child
    param = self.names[names[0]]
            ~~~~~~~~~~^^^^^^^^^^
KeyError: 'name'

The above exception was the direct cause of the following exception:

  File "/home/mairain/original-pymodaq/pymodaq/src/pymodaq/utils/managers/configurator/utils.py", line 404, in mimeData
    module, module_type = get_module_from_param(param_with_path)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mairain/original-pymodaq/pymodaq/src/pymodaq/utils/managers/configurator/utils.py", line 109, in get_module_from_param
    module = param_module.child('name').value()
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mairain/.virtualenvs/original-pymodaq/lib/python3.12/site-packages/pyqtgraph/parametertree/Parameter.py", line 787, in child
    raise KeyError(f"Parameter {self.name()} has no child named {names[0]}") from e
KeyError: 'Parameter actuator has no child named name'
QLayout: Cannot add a null widget to QVBoxLayout/
```

src/pymodaq/utils/managers/configurator/utils.py
```diff
@@ -0,0 +1,410 @@
...
+import os
+from pathlib import Path
+import sys
+from dataclasses import dataclass
+from typing import Union, Tuple
+
+
+from qtpy import QtWidgets, QtCore
+
+from qtpy.QtCore import QMimeData, Qt, QVariant, QModelIndex
```
malik-irain: `QVariant` is not available in pyside6, use `QVariant` from `pymodaq_gui.qvariant`

seb5g:
> A few comments on usage:
> 
>     * Preset manager starts with `preset_default` selected but not showing anything (I first though it was loading the Configurator). And we can't rename a preset anymore?
It now load it directly. I added a button/functionality to copy the preset and rename it . We could add an entry in config to write which preset should be set by default on the UI.
> 
>     * I don't know if it was done in this PR but the preset manager remove too much configuration options
>       
>       * For example, I can't make a preset that would connect automatically to a LECO Actor.
That's done on purpose, because all this will be done using the Configurator. Preset only declare ControleModules. A default configuration will be created for each preset and loaded just after it's loaded. You' ll do things like leco connection in there. It's important as the configurator handles order in the actions it performs.
> 
>     * When adding an actuator value
>       
>       * The window could be smaller
weird it's minimal on my screen

>       * Units are always prefixed with `m`, even when they already have a prefix, i.e. Mock actuators for Power (in `mW`) and Xaxis (in `µm`)  are respectively in `mµW` and `mµm`. They work as they name say, 1000 `mµW` = 1 `µW`. They are showed as their normal units (`mW` and `µm`) in the table though.
i removed the Si prefix (milli, kilo...)
> 
>     * Some configuration entries can be selected for configuration but they have no effect, so they shouldn't be modifiable. For example: Axes, Show Config, ...
Done. Readonly settings or the one having a dedicated keywork won't be selectable in the configurator
> 
>     * Parameter tree "Title nodes" (I don't know how they're called but "Actuator", "LECO Options", etc.) that don't contain value can be picked  for configuration, they fallback to the first configurable entry in the tree.
you can now pick a "group", that will create one configuration entry for each setting within the group!
> 
>     * A `del` key shortcut, or a button to remove the selected entry could be nice
there was a button already (the left arrow) and I added the Suppr shortcut on it.
>

malik-irain:
>>     * I don't know if it was done in this PR but the preset manager remove too much configuration options
>>       
>>       * For example, I can't make a preset that would connect automatically to a LECO Actor.
> 
> That's done on purpose, because all this will be done using the Configurator. Preset only declare ControleModules. A default configuration will be created for each preset and loaded just after it's loaded. You' ll do things like leco connection > in there. It's important as the configurator handles order in the actions it performs.

Then,  something else need to be changed as when loading a preset with a LECODirector  it will be initialized by default, using the current state of settings. Meaning that it will try to connect to `actor_name` (default value) and the configuration can not be edited nor loaded before automatic initialization leading to:

```python
FO:pymodaq.pymodaq.daq_viewer.detector 02:Initializing DAQ_Viewer: detector 02
New Component b'detector 02_8585_director' at b'\x00k\x8bEo'.
Receiver 'b'actor_name'' is not in the addresses list.
ERROR:pymodaq.daq_viewer:Hardware couldn't be initialized
  File "/home/mairain/original-pymodaq/pymodaq/src/pymodaq/control_modules/daq_viewer.py", line 1322, in ini_detector
    infos = self.detector.ini_detector(controller)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mairain/original-pymodaq/pymodaq/src/pymodaq/utils/leco/daq_xDviewer_LECODirector.py", line 82, in ini_detector
    self.controller.set_remote_name(self.communicator.full_name)
  File "/home/mairain/original-pymodaq/pymodaq/src/pymodaq/utils/leco/director_utils.py", line 26, in set_remote_name
    self.ask_rpc(method=GenericMethods.SET_REMOTE_NAME, name=name or self.communicator.name)
  File "/home/mairain/.virtualenvs/original-pymodaq/lib/python3.12/site-packages/pyleco/directors/director.py", line 125, in ask_rpc
    return self.communicator.ask_rpc(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mairain/.virtualenvs/original-pymodaq/lib/python3.12/site-packages/pyleco/core/internal_protocols.py", line 129, in ask_rpc
    return self.interpret_rpc_response(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mairain/.virtualenvs/original-pymodaq/lib/python3.12/site-packages/pyleco/core/internal_protocols.py", line 105, in interpret_rpc_response
    result = self.rpc_generator.get_result_from_response(response_message.payload[0])
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mairain/.virtualenvs/original-pymodaq/lib/python3.12/site-packages/pyleco/json_utils/rpc_generator.py", line 75, in get_result_from_response
    raise exc(error)
pyleco.json_utils.errors.ServerError: -32093: Receiver is not in addresses list.
Error Data: actor_name
```
I don't know what the simpliest would be but:

-  Add an init setting in the configurator
    - Force LECODirector to start uninitialized in the preset manager and
    - Start everything uninitialized
- Allow these settings to be set in the preset manager for LECODirector

Then we could select the values we want for `actor_name` (and server, port, if needed) and then initialize from loading the configuration. But when selecting `actor_name` and set a new value in the Configurator the following Exception arise:

```python
ValueError: could not convert string to float: 'actuator'
  File "/home/mairain/original-pymodaq/pymodaq/src/pymodaq/utils/managers/configurator/utils.py", line 75, in setModelData
    model.setData(index, editor.value(), Qt.ItemDataRole.EditRole)
  File "/home/mairain/original-pymodaq/pymodaq_gui/src/pymodaq_gui/utils/widgets/table.py", line 164, in setData
    if self.validate_data(index.row(), index.column(), self.cast(value)):
                                                       ^^^^^^^^^^^^^^^^
```
>>  * When adding an actuator value
>>      
>>      * The window could be smaller
>
> weird it's minimal on my screen

I believe you, but mine is like that, not that disturbing anyways

Fill in information about this actuator target value_001

src/pymodaq/utils/managers/modules_manager.py
```diff
@@ -276,6 +282,43 @@ def value_changed(self, param):
...
+        """Retrieve settings from all control modules and return it as a grouped Parameter"""
+        settings = Parameter.create(
+            title="Control Modules Settings",
+            name="control_modules_settings",
+            type="group",
+            children=[{'title': 'Actuators:', 'name': ModuleType.Actuator.value, 'type': 'group'},
+                      {'title': 'Detectors:', 'name': ModuleType.Detector.value, 'type': 'group'},],
+        )
+
+        for ind_act, actuator in enumerate(self.actuators_all):
```
Ashwolaa: Not super fond of the name, why not use a setter for self.actuators or directly called self._actuators?

src/pymodaq/utils/managers/modules_manager.py
```diff
@@ -276,6 +282,43 @@ def value_changed(self, param):
...
+
+            settings.child(ModuleType.Detector).addChild(
+                {'title': detector.title, 'name': f'{ModuleType.Detector}_{ind_det:03.0f}', 'type': 'group',
+                 'children': [
+                     {'title': 'Name:', 'name': 'name', 'type': 'str', 'value': detector.title}
+                 ]},
+            )
+            settings.child(ModuleType.Detector,
+                           f'{ModuleType.Detector}_{ind_det:03.0f}').addChildren(detector_settings.children())
+
```
Ashwolaa: I would factorize with a single function for any module.
I do not understand why you do not use directly detector.settings which is already a Parameter.

src/pymodaq/utils/managers/preset/preset_manager.py
```diff
@@ -0,0 +1,372 @@
...
+# check if preset_mode directory exists on the drive
+preset_path = config_mod_pymodaq.get_set_preset_path()
+overshoot_path = config_mod_pymodaq.get_set_overshoot_path()
+layout_path = config_mod_pymodaq.get_set_layout_path()
+
+
+
+
+class PresetManager(CustomApp):
+
```
Ashwolaa, line 38: How about separating Actuators and Detectors into two ParameterTree? The layout might be even better

src/pymodaq/utils/managers/preset/preset_manager.py
```diff
@@ -0,0 +1,372 @@
...
+                presets.sort()
+                index = presets.index(text.lower())
+                self.get_action('presets').insertItem(index-1, text)
+
+            self.get_action('presets').setCurrentText(text)
+            self.save_check()
+
+    def copy_preset(self):
+        text, ok = QtWidgets.QInputDialog.getText(None, 'Enter a NEW Preset name',
+                                                  'Preset name:', QtWidgets.QLineEdit.Normal)
```
Ashwolaa: QtWidgets.QLineEdit.EchoMode.Normal

src/pymodaq/utils/managers/preset/preset_manager.py
```diff
@@ -0,0 +1,372 @@
...
+            params_det = [
+                {'title': 'Detectors:', 'name': ModuleType.Detector.value, 'type': 'groupdet'}
+            ]  # [PresetScalableGroupDet(name='Detectors')]
+            self.settings = Parameter.create(title='Preset', name='Preset', type='group',
+                                             children=params_act + params_det,)
+        self.get_action('presets').setCurrentText(preset_file.stem)
+
+    def create_preset(self):
+        text, ok = QtWidgets.QInputDialog.getText(None, 'Enter a NEW Preset name',
+                                                  'Preset name:', QtWidgets.QLineEdit.Normal)
```
Ashwolaa: QtWidgets.QLineEdit.EchoMode.Normal

Ashwolaa:
wow the review comes directly, I though I would give them all in one go

Ashwolaa (commented):
nonlinearscanner does not work, the import are no longer good

src/pymodaq/examples/nonlinearscanner.py
```diff
@@ -105,7 +105,7 @@ def main():
     dash = DashBoard(area)
     file = Path(get_set_preset_path()).joinpath(f"{config('presets', 'default_preset_for_scan')}.xml")
     if file.exists():
-        dash.set_preset_mode(file)
+        dash.apply_preset_to_dashboard(file)
```
Ashwolaa: Imports are no longer good here, I cannot test it

Ashwolaa:
Overall, this looks very good and quite robust. 


In terms of usage:
- I would prefer a list view compared to a QCombobox to select the config with +- buttons connected to it.
- I do not like the current delegates
  - spinbox : It is half cropped
  - checkbox: Can barely see the tickbox
  - pushbutton: Not good
- the saving should be more "automatic". In practice, I would like to directly apply my configurator just pressing the load button on the dashboard or having a binding in the configurator itself. Here, one has to save it before being able to use, it makes the testing a bit more laborious. Maybe one could think differently, as the loaded one is always a temp copy of a saved one. Then the reload would still work as intended and one could use it directly to test it.
- move up/down should be at a different position (below/top of the table?).
- A waiting time between two actions should be added. 
- Multiple selections is lacking to remove more than one entries
- The context menu is nice but it should be more well organized.
  - Action on the table: Action/Copy/Remove/Clear
  - Separator
  - Load/Save (the base folder should be config associated to the current preset); By the way, the refresh should be immediate when closing the popup, e.g. if I delete a file, then it still shows up in the QCombobox and if I save, it does not show up. This is also the case for the actions in the dashboard. By the way, we should only use one QAction if possible between the two.
- I feel like we are lacking some actions, maybe some buttons could be added instead of using the context menu, e.g. Action==)Actuator==)Init / Go to / ... ; Action ==) Detector ==) Init / Grab / ...
- A visual indicator to know what can be "configured" would be great
- I can change the name in the settings, is that intended?
- The ... sel is redundant with the +. Moreover, one can just close the pop up without problem, this makes it quite confusing as I can now save a file with ... . In my opinion, selecting this should either open the config directory (to select a file stored somewhere else) or give a blank config. Then, it should only ask for the config name when one tries to save it. 
- We are lacking shortcuts to directly access the directory!
- An indicator for which settings is being initialized would be nice. This would help to keep track. Maybe adding an extra column with some red/green area? By the way, is there a feedback mechanism? what happens if an action fails? Does it stop?
- the shortcut does not work well in my case (only the delete)
- in addition of dragging, a doubleclick on the parameter could also send it directly to the configurator table

I'll keep playing a bit with it and come up with additional remarks.

seb5g:
> Then, something else need to be changed as when loading a preset with a LECODirector it will be initialized by default, using the current state of settings. Meaning that it will try to connect to `actor_name` (default value) and the configuration can not be edited nor loaded before automatic initialization leading to:
> 
> I don't know what the simpliest would be but:
> 
>     * Add an init setting in the configurator
that's what I thought and implemented it! This and setting an actuator value, I called it SpecialConfiguratorEntry (theres an enum for that). We can later add other stuff if needed

> 
> Then we could select the values we want for `actor_name` (and server, port, if needed) and then initialize from loading the configuration. But when selecting `actor_name` and set a new value in the Configurator the following Exception arise:
> 
> ```python
> ValueError: could not convert string to float: 'actuator'

fixed

> 
> > > * When adding an actuator value
> > >   
> > >   * The window could be smaller
> > 
> > 
> > weird it's minimal on my screen
> 
> I believe you, but mine is like that, not that disturbing anyways

It's really small on mine :-D

Ashwolaa:
It is also small on mine ^^

seb5g:
> Overall, this looks very good and quite robust.
> 
> In terms of usage:
> 
>     * I would prefer a list view compared to a QCombobox to select the config with +- buttons connected to it.

you mean a list with all configs displayed? Why?

> 
>     * I do not like the current delegates
>       
>       * spinbox : It is half cropped
>       * checkbox: Can barely see the tickbox
>       * pushbutton: Not good
Agreed, I guess this need some styling that will be done a bit later. I get the right delegate widget out of the ParameterType and  pyqtraph factory. We could probably make them look better by doing some DisplayRole, BackgroundRole, EditRole in the data method of the underlying model. Keep it in mind for later
> 
>     * the saving should be more "automatic". In practice, I would like to directly apply my configurator just pressing the load button on the dashboard or having a binding in the configurator itself. Here, one has to save it before being able to use, it makes the testing a bit more laborious. Maybe one could think differently, as the loaded one is always a temp copy of a saved one. Then the reload would still work as intended and one could use it directly to test it.

this will be somehow implemented in the next PR where I refactorise all the managers
> 
>     * move up/down should be at a different position (below/top of the table?).
this is usual for such control to be at this position... but the usual way will be to actually drag drop entries in the list. Did you try?
> 
>     * A waiting time between two actions should be added.
This will be a SpecialConfiguratorEntry indeed!!
> 
>     * Multiple selections is lacking to remove more than one entries
Yes but I didn't want to handle this for the drag mechanism... will do later. There is a shortcut "suppr" to delete quickly entries
> 
>     * The context menu is nice but it should be more well organized.
ok keep that in mind for later polishing
>       
>       * Action on the table: Action/Copy/Remove/Clear
>       * Separator
>       * Load/Save (the base folder should be config associated to the current preset); By the way, the refresh should be immediate when closing the popup, e.g. if I delete a file, then it still shows up in the QCombobox and if I save, it does not show up. This is also the case for the actions in the dashboard. By the way, we should only use one QAction if possible between the two.
Adressed in the next PR on managers
> 
>     * I feel like we are lacking some actions, maybe some buttons could be added instead of using the context menu, e.g. Action==)Actuator==)Init / Go to / ... ; Action ==) Detector ==) Init / Grab / ...
> 
>     * A visual indicator to know what can be "configured" would be great
> 
>     * I can change the name in the settings, is that intended?
no. I will adress this using the special paramter keyword and set all settings to readonly
> 
>     * The ... sel is redundant with the +. Moreover, one can just close the pop up without problem, this makes it quite confusing as I can now save a file with ... . In my opinion, selecting this should either open the config directory (to select a file stored somewhere else) or give a blank config. Then, it should only ask for the config name when one tries to save it.
> 
>     * We are lacking shortcuts to directly access the directory!
> 
>     * An indicator for which settings is being initialized would be nice. This would help to keep track. Maybe adding an extra column with some red/green area? By the way, is there a feedback mechanism? what happens if an action fails? Does it stop?
could you elaborate?
> 
>     * the shortcut does not work well in my case (only the delete)
there is only this one??
> 
>     * in addition of dragging, a doubleclick on the parameter could also send it directly to the configurator table
ok
> 
> 
> I'll keep playing a bit with it and come up with additional remarks.

I'll summarize here what you said for future PR (for the moment I'll concentrate on the manager refactorisation as I'm already almost done!

- [ ]  list view or a QCombobox
- [ ] delegates styling
- [ ] automatic saving
- [ ] move up/down buttons
- [ ] waiting time entry
- [ ] context menu styling
- [ ] buttons for specific actions a bit like in the preset to select control modules
- [ ] A visual indicator to know what can be "configured" would be great
- [ ] readonly settings
- [ ] The ... was experimental to be quick in creating a new one. The ... is normally never saved.
- [ ] shortcuts?
- [ ] doucleclick

seb5g:
For the moment I propose to merge it as is (make sure it is still ok running) and we'll get back to styling UI/UX after refactorisation has been done

Ashwolaa:
> For the moment I propose to merge it as is (make sure it is still ok running) and we'll get back to styling UI/UX after refactorisation has been done

Looks good for me, I already rebased my branches on it 


> > ```
> > * I would prefer a list view compared to a QCombobox to select the config with +- buttons connected to it.
> > ```
> 
> you mean a list with all configs displayed? Why?

I am too used to standard IDE where it feels easy to switch from one file to another.

> > ```
> > * move up/down should be at a different position (below/top of the table?).
> > ```
> 
> this is usual for such control to be at this position... but the usual way will be to actually drag drop entries in the list. Did you try?


Yeah I have  mostly been using the drag/drop, the buttons position just feel a bit off


> > * An indicator for which settings is being initialized would be nice. This would help to keep track. Maybe adding an extra column with some red/green area? By the way, is there a feedback mechanism? what happens if an action fails? Does it stop?

> could you elaborate?

If one can apply directly the config from the configurator, a visual indication to which config has been successfully setup would be nice, like a little LED. 


> > ```
> > * the shortcut does not work well in my case (only the delete)
> > ```

> there is only this one??

I tried arrows but did not manage to trigger reaction, it switches between columns in the table or between parameter name/value in the settings.

src/pymodaq/utils/managers/preset/preset_manager.py
seb5g: done

src/pymodaq/utils/managers/modules_manager.py
seb5g: modules manager handles selection of a few of all the modules from the dashboard. 
self.actuators is referring to the one that are selected by the user

seb5g: and this is  part of the api for external use so self._actuators won't do

src/pymodaq/utils/managers/modules_manager.py
seb5g: done

seb5g: >  do not understand why you do not use directly detector.settings which is already a Parameter.

because passing Parameters by reference is creating huge mess everywhere, sometime removing them from the ParameterTree they belong to (so here would go from the control module to the configurator...) I'm sure you don't want that ;-)

src/pymodaq/utils/managers/preset/preset_manager.py
seb5g: not convinced...

src/pymodaq/utils/managers/preset/preset_manager.py
seb5g: done

Ashwolaa:
A tool to convert old presets would be very useful by the way.

seb5g:
Yes!

PR #753, opened 2025-11-03
Refactoring the Managers

seb5g:
Using a Base class to handle common actions to all managers to be and external actions (ones that should appear in other widgets: the DashBoard for instance).

This is using new stuff from pymodaq_gui (updated action manager essentially)


tests/utils/managers/preset_manager_test.py
```diff
@@ -1,33 +1,123 @@
...
+
 from qtpy import QtWidgets
 from pymodaq.utils.managers import PresetManager
+from pymodaq_gui.parameter import Parameter
+from pymodaq_gui.parameter import utils as putils
+from pymodaq_gui.parameter import ioxml
 
+@pytest.fixture
+def init_qt(qtbot):
+    return qtbot
```
malik-irain, line 19: not needed

tests/utils/managers/preset_manager_test.py
```diff
@@ -1,33 +1,123 @@
...
+    external_ui.addToolBar(toolbar)
+    external_ui.menuBar().addMenu(menu)
+
+    preset_manager = PresetManager(menu=menu, toolbar=toolbar)
+    preset_manager.update_entry_base()
+    qtbot.addWidget(preset_manager.mainwindow)
+    preset_manager.mainwindow.show()
+    qtbot.addWidget(external_ui)
+
+    yield preset_manager, qtbot
```
malik-irain, line 37: It should use qtbot directly
I would prefer if the fixture was called "preset" or "preset_manager" and yield only the preset manager

We don't need to "keep" the qtbot object: https://github.com/pytest-dev/pytest-qt/blob/master/tests/test_screenshot.py

tests/utils/managers/configurator_manager_test.py
```diff
@@ -0,0 +1,53 @@
...
+from qtpy import QtWidgets
+from pymodaq.utils.managers import Configurator
+from pymodaq_gui.parameter import Parameter
+from pymodaq_gui.parameter import utils as putils
+from pymodaq_gui.parameter import ioxml
+
+@pytest.fixture
+def init_qt(qtbot):
+    return qtbot
+
```
malik-irain, line 18: Not needed

tests/dashboard_test.py
```diff
@@ -13,9 +12,9 @@
 def init_qt(qtbot):
```
malik-irain, line 12: not needed

tests/dashboard_test.py
```diff
@@ -28,8 +27,7 @@ def test_main_setfilepreset_quit(self, init_qt):
         win.setWindowTitle('PyMoDAQ Dashboard')
```
malik-irain, line 27: win should be attached to qtbot

src/pymodaq/dashboard.py
```diff
@@ -273,6 +224,26 @@ def __init__(self, dockarea):
...
+                                         toolbar=self.get_toolbar('configurator'),
+                                         preset_filename=self.preset_manager.entry)
+        self.get_toolbar('configurator').setEnabled(False)
+
+    def update_after_preset(self, preset_name: str):
+        self.configurator.preset_filename = preset_name
+        self.configurator.entry = 'default'
+        self.get_menu('configurator').setEnabled(True)
+        self.get_toolbar('configurator').setEnabled(True)
+        self.configurator.apply_entry(self.configurator.entry_filename)
```
malik-irain, line 245: maybe make the configurator modification part a method? (Except if order is important)
```python
class Configurator:
...
def update_with_new_preset(self, preset_name: str):
     self.preset_filename = preset_name
     self.entry = 'default'
     self.apply_entry(self.entry_filename)
...
```

malik-irain:
From an usage point of view, I'm not a fan of the new icons for loading preset or configuration

seb5g:
> From an usage point of view, I'm not a fan of the new icons for loading preset or configuration

Well me neither I actually tried to use a standardized set but for sure we have to look around the qt awesome and actually kind of vote for what we would prefer.  or we could create a mapping between different sets of icons and declare in the configuration the one we want to use . the mappings could point to either the old style icons , the qt6 ones or the qt awesome set

PR #757, opened 2025-11-12
remove other type of suffix... encountered during install

seb5g:
While trying to create locally a "dev tag" :

git tag -a 5.2.0.dev (without triggering a release)

the installation of the packages was failing because the version extracted by the git_describe returned something with the "^" character different from the "~" one already taken into account...
This PR strip this character BUT we should try and see where it comes from and what are the other bits that could be added...

### DIFF
--- a/git_describe.py
+++ b/git_describe.py
@@ -28,6 +28,7 @@ try:
 
     # Remove "~1" or similar suffix
     version = version.split("~")[0]
+    version = version.split("^")[0]
 except Exception:
     version = FALLBACK_VERSION
 


seb5g:
and we should modify the python-publish action to not publish on pypi if there is "dev" in the tag!

malik-irain:
I don't understand why you want to create a dev tag, we didn't do it before

seb5g:
Well it doesn't matter much for the main subject of this PR. I don't know where these suffixes come from!!!

Concerning the tag, it was an attempt to have a clearer version number while working on the dev branch. We discussed that already. At the moment the version of the dev branch is indexed from the last tag which is 5.1.2 while we know the next tag of the dev branch will be 5.2.0 so on dev we should have something like 5.2.0.dev... as a version when installing from source and in editable mode.

seb5g:
the tests are also failing at the readme stage.

malik-irain:
> the tests are also failing at the readme stage.

Yes because the branch name if detected as `/refs/pull/757/merge/` and it conflicts with `sed` command separators. It's not  important on this PR as it doesn't modify the README but definitely need improvement!

> Concerning the tag, it was an attempt to have a clearer version number while working on the dev branch. We discussed that already. At the moment the version of the dev branch is indexed from the last tag which is 5.1.2 while we know the next tag of the dev branch will be 5.2.0 so on dev we should have something like 5.2.0.dev... as a version when installing from source and in editable mode.

When it wasn't a mono repo, branch 5.2.x_dev reported a version like `5.1.<last_tag>` or `5.1.<last_tag+1>-dev...`  depending on modifications on the branch after the last tag or not. I mean, I'm not against the modifications but isn't it overthought?


Anyways, **I can't accept** any PR until you aknowledge my email 😉 (I need you to delete dev branch and rename dev2 as dev, more information in the email I sent)

PR #767, opened 2025-11-13
Documentation about monorepo and install script update

malik-irain:
Updated documentation about monorepo and the new install script
Install script installs in editable mode by default and can install optional dev dependencies

It's still missing a figure about the new branches organization (but it will be ready soon)

### DIFF
--- a/docs/src/developer_folder/contributing.rst
+++ b/docs/src/developer_folder/contributing.rst
@@ -19,7 +19,7 @@ If you're willing to help, there are several ways to do it:
 For the last point, here are some pointers:
 
 You should `fork and clone <https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/fork-a-repo#forking-a-repository/>`_
-the up-to-date GitHub repo: https://github.com/PyMoDAQ using git command line or GitHub Desktop
+the up-to-date GitHub repo: https://github.com/PyMoDAQ/PyMoDAQ using git command line or GitHub Desktop
 (the latter is recommended for newcomers, just make sure to uncheck "Copy the ... branch only").
 Then create a dedicated branch name from the change you want to work on
 (using git).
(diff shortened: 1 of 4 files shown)


docs/src/developer_folder/contributing.rst
```diff
@@ -29,10 +29,10 @@ Finally it is advised to create a dedicated virtual environment for this and ins
 - ``mamba create -n dev_env``
 - ``mamba activate dev_env``
 - ``cd`` to the location of the folder where you downloaded or cloned the repository.
-- install the package as a developer with test tools using the command ``pip install -e ".[dev]"``.
+- install the package as a developer with test tools using the command either by using the provided script ``python install-packages.py -d``.
```
seb5g, line 32: why not using the same command -e?

docs/src/developer_folder/contributing.rst
```diff
@@ -29,10 +29,10 @@ Finally it is advised to create a dedicated virtual environment for this and ins
...
+    - it calls `pip install -e` on each pymodaq package (pymodaq_utils, pymodaq_data, pymodaq_gui and pymodaq) in the package folder.
     - the `-e` option tells pip to install the package in editable mode, meaning that any change done to the source will be reported to the installed package.
-    - the `".[dev]"` argument indicate to install the current folder as a package alongside its optional dependancies, declared as `dev` in the `pyproject.toml` file.
-    - the quotes around `[dev]` just allow to escape the sequence in order to not be interpreted by the shell.
+    - the `-d` option activates `[dev]` in the package names transmitted to `pip install` (e.g. `pymodaq_utils[dev]`). It indicate to install the package alongside its optional dependancies, declared as `dev` in the `pyproject.toml` file.
```
seb5g, line 35: then can we do -e -d with your script? . Maybe you should be more explicit on what it means the [dev] optional install

docs/src/tutorials/write_documentation.rst
```diff
@@ -118,7 +118,7 @@ We will now create a new branch from *pymodaq-dev* so that we can isolate our ch
 
 Finally, install our local repository in edition mode in our Python environment
 
-``(pmd_dev) >pip install -e .``
+``(pmd_dev) >python install-packages.py``
```
seb5g, line 121: should we not explicitly call -e to be on the same line?

install-packages.py
```diff
@@ -8,19 +8,25 @@
...
+
+    parser.add_argument(
+        "-d", "--dev",
+        action="store_true",
+        help="Also install dev requirements"
+    )
+
     parser.add_argument(
-        "-e", "--editable",
+        "-n", "--non-editable",
```
seb5g, line 21: strange behavior isn't it?

seb5g (changes requested):

docs/src/developer_folder/contributing.rst
malik-irain: Because it doesn't activate the same thing, `-d` triggers `[dev]` to be sent to pip, `-e` is a pip option. My reasoning is that someone using the script needs the packages to be installed as editable (an average user will still use `pip install pymodaq`).

When `[dev]` dependencies are activated, all qt backends are installed and it can be troublesome to chose which one to use or to correctly uninstall them (they install several packages not automatically uninstalled with them).
 
Also having `-d` as an option would lead an user to read the documentation as not using it would cause issues when developing for pymodaq ;)

docs/src/developer_folder/contributing.rst
malik-irain: no, `-e` was removed as it's now by default

docs/src/tutorials/write_documentation.rst
malik-irain: not if its functionality is activated by default

install-packages.py
malik-irain: Not if you think it should be activated by default ;)

PR #771, opened 2025-11-14
An attempt to fix coverage report uploads

malik-irain:
The goal is to have only one coverage report uploaded on codecov (not that we have a choice with a monorepo!)

To do that, each test action for a package generate a coverage report and uploads it as an artifact. Then another action, wait for any running test, then try to download each coverage artifact, if no artifact is found (they are only kept one week) tests are restarted. Then all coverage artifacts are combined and sent to codecov!

I don't know if it works, so I'm testing it now!

### DIFF
--- a/.github/workflows/coverage-upload.yml
+++ b/.github/workflows/coverage-upload.yml
@@ -63,9 +63,8 @@ jobs:
           path: artifacts
           merge-multiple: true
 
-
-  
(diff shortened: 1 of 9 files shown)


seb5g:
ok have a good test!

malik-irain:
It's not really working for now, I keep it open until I have some time to do it

PR #775, opened 2025-11-18
Feature/singleton on Config

seb5g:
Implement the Singleton Pattern on Config Base class to make sure one use the same instance everywhere even if saved in between!!!

Expose also the load method to reload if config has just been saved

apply this to always get the right timeout in module manager


seb5g:
@malik-irain could you merge this together with the things you implemented at the time?

malik-irain:
So here are my modification:
1. I moved the Singleton metaclass to its own file as it could be reused by other component without needed to load config.py
2. I added read/write locks for a multiple readers / single writer scenario, so we can have several readers at once but only one writer. I needed to change library in the end as I also need reentrancy (one can lock an already acquired lock) and deescalation (acquire a read lock when a write lock is acquired), so it took some time to adapt!
3. I removed the load function as it shouldn't be needed, but then I stumbled into the same issues as you did I think. So I changed the mecanism in which TreeFromToml saved the config. 

To achieve the last point, the idea is to cache any modification into an object (here a dict) so that:
- We can still retrive the path of the modification to do in the config file ('general', 'check_version') for example
- Any new modification should mask the previous one
- We don't want to apply the modification unless the Save button is pressed

I struggled a little bit but remembered that a dict with a the config path tuple as key and the modified value as value would work like a charm for that!


Did you have this configuration update issues somewhere else?

(test fails but they work locally)

packages/pymodaq_gui/src/pymodaq_gui/utils/widgets/tree_toml.py
```diff
@@ -5,13 +5,22 @@
...
 import datetime
+from typing import Tuple
 
 from qtpy import QtWidgets, QtCore
 from qtpy.QtCore import QObject
 
 from pymodaq_gui.parameter import ParameterTree, Parameter
 from pymodaq_utils.config import Config, create_toml_from_dict
 
+def get_param_path(param : Parameter, include_top_level_parent=False) -> Tuple[str]:
```
seb5g: this method already exists in pymodaq_gui.parameter.utils: get_param_path

seb5g:
did you do specific tests to make sure the object is singleton? That the update from tree is synchronized? And that the read/write lock work as expected? It looks thorough but a bot complex to be checked only by reading it!

seb5g:
and the failing tests seems to be because I removed the timeout entry from the toml (and in the source code) but not from the tests where its presence was tested. While you're at it, could you clean the tests? (it weird that they pass, probably because you kept the old config file so the timeout entry was still present!)

malik-irain:
I can test the Tree synchronization, if by that you mean that changing an entry in the tree and accepting the changes updates the config and when cancelling nothing is changed

But I have no idea on how to test the locks except reading carefully until I trust my code

seb5g:
should I do something?

malik-irain:
I wanted to add you as a reviewer now that I made some modifications! It should be ready now

malik-irain:
There's now a "cache" at the config class level. Or to be exact I have now two config dict inside:

- A one from the fusion of system and user config that gets modified and read as usual.
- Another one that contains the user file contend only and on which modification are copied. This one gets saved as the user file when asked to save

Also config is automatically saved at shutdown as much as possible (destructor + atexit)

I guess, it's ready for review @seb5g

packages/pymodaq_utils/src/pymodaq_utils/config.py
```diff
@@ -273,81 +280,72 @@ class BaseConfig:
...
     def __init__(self):
-        self._config = self.load_config(self.config_name, self.config_template_path)
+        self._lock = ReaderWriterLock()
+        self._config = {}
+        self._modified_config = {}
+        self.load()
+        atexit.register(self.save)
+
+    def __del__(self):
+        print("bye bye")
```
seb5g: remove!

seb5g:
Hi @malik-irain I pushed two commits:

* first to save/cache the correct type of a value (for a list it should be the limits with the value first in the list)
* using ParameterManager base class to refactor the code on settings/tree/signals handling!

can you check?

malik-irain:
It's good for me now! If it's good for you and tests pass, it can be accepted

PR #779, opened 2025-11-24
  Documentation: Quick Start for Ubuntu

malik-irain:
As @quantumm said in #777

> We separate the Quick Start in two sections: one for Windows and one for Ubuntu. It allows to properly separate the installation instructions, and highlights the compatibility with Ubuntu.


malik-irain:
I think this PR is causing doc to not compile on readthedocs

seb5g:
Could you identify what's wrong?

malik-irain:
Yes, actually I'm not 100% sure it's coming from these modifications, as when it started failing, it seems  nothing happened

seb5g:
issue is coming from pyqtgraph latest release with an error in the docstring. Completeley unrelated to this PR so I accept it!

PR #794, opened 2025-11-27
Removed support for python 3.9 and added 3.13

malik-irain:


malik-irain:
Before accepting, let's upgrade python version!

Issue #786, opened 2025-11-26
[BUG] Pyqtgraph enforcing Value error if value is missing


seb5g:
### Describe the bug

Since the new version of pyqtgraph 0.14.0 the ValueError is raised if a parameter is created without specific value (so far it could take it's default from other mechanisms). They said it was to be deprecated for a while but we didn't tackle it. We should do so now. Most issues should appear by running pymodaq_gui tests but maybe a more general survey of our modules is needed (and include their test with that repsect)

malik-irain:
Fixed by #798

PR #797, opened 2025-11-28
Pyqtgraph default value

malik-irain:
Solves #786 by providing a monkey patch to allow Parameter types object to ensure returning "None" if no value is found.

The issue is that Parameter object with no value are allowed, and can be linked together as children, but then using the `getValue()` method will fail if a value is not set. 

We use that method _(mainly in tests...)_ and it can't just be surrounded by a  `try: ... except: ...` block  as it calls recursively  `value()` and `getValue()` on the children of the current Parameter and will fail if one the children value is not set. We can't really easily provide a replacement value by catching an exception. 

I first tried to subclass it to make it cleaner but as some used `pygtgraph.Parameter` objects are actually already subclasses (`ListParameter` for example), they would still use the `pyqtgraph.Parameter.value` method.

Another way would be to find all instances of Parameter and subclasses of Parameter objects and build them with `value=None` if no value is provided but it is time consuming.

Yet another way would be to change stop using `getValue()` and implement a function that replaces it's mechanism (OK for that). But if `value()` needs to also change, it will be troublesome as there is a lot of occurrence of its usage in the code.

### DIFF
--- a/packages/pymodaq/src/pymodaq/control_modules/daq_viewer_ui.py
+++ b/packages/pymodaq/src/pymodaq/control_modules/daq_viewer_ui.py
@@ -370,7 +370,8 @@ class DAQ_Viewer_UI(ControlModuleUI, ViewerDispatcher):
 
 def main(init_qt=True):
     from pymodaq.utils.gui_utils.dock import DockArea
-    from pymodaq.utils.parameter import ParameterTree, Parameter
+    from pymodaq.utils.parameter import ParameterTree
+    from pymodaq_gui.parameter import Parameter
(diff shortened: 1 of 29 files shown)


malik-irain:
@seb5g don't accept yet, I want to try the last point I mentioned

malik-irain:
The PR about this doesn't say much about motivations https://github.com/pyqtgraph/pyqtgraph/pull/3286
Even the one that did the deprecation isn't more verbose https://github.com/pyqtgraph/pyqtgraph/pull/2853

But there is an issue on pyqtgraph about this problem https://github.com/pyqtgraph/pyqtgraph/issues/3378

It seems the "official" solution from one of the maintainers is to give a None value to the groups

PR #799, opened 2025-12-01
apply list selection, value chosen in combobox become the first element of the list

seb5g:
This is the application of a PR from @Ashwolaa on pymodaq_gui


seb5g:
@malik-irain all the tests pass but for this one has to use the new config file template (so delete the old one). I print a message in the logger or on the console before exiting the process to make that clear!

seb5g:
@malik-irain this is ready for review!

packages/pymodaq_utils/src/pymodaq_utils/resources/config_template.toml
```diff
@@ -6,8 +6,7 @@ language = "English"
 country = "UnitedStates"
 
 [qtbackend]
-backends = [ "pyqt5", "pyqt6", "pyside2", "pyside6",]
-backend = "pyqt6"
+backend = ["pyqt6",  "pyqt5", "pyside2", "pyside6",]
```
mari-kuri: We don't support pyside2 anymore, we can take advantage of this PR to remove it

packages/pymodaq_utils/src/pymodaq_utils/config.py
```diff
@@ -26,6 +26,16 @@
 KeyType = TypeVar('KeyType')
 
 
+def replace_item_in_list(items: list[str],
+                         item_to_check: str,
+                         item_to_replace:str):
+    if not isinstance(items, list):
```
mari-kuri, line 38: wrong parameter naming? item_to_replace is the replacement and not the one being replaced, to keep it simple I suggest:

items -> items
item_to_check -> old
item_to_replace -> new


Also for the typing, why not using the KeyType for the typing, generic types are generally called T so refactor to use T where KeyType is used and rewrite this function to be:

```python
def replace_item_in_list(items: list[T],
                         old: T,
                         new:T):         
```

packages/pymodaq_data/src/pymodaq_data/data.py
```diff
@@ -2089,7 +2089,7 @@ def errors_as_dwa(self):
         else:
             raise ValueError(f'Cannot create a dwa from a None, should be a list of ndarray')
 
-    def plot(self, plotter_backend: str = config('plotting', 'backend'), *args, viewer=None,
+    def plot(self, plotter_backend: str = config('plotting', 'backend')[0], *args, viewer=None,
```
mari-kuri, line 2092: This uses "utils" config file where there is a "plotting", "backend" entry but it is not a list!

packages/pymodaq_utils/src/pymodaq_utils/config.py
```diff
@@ -384,5 +386,41 @@ def dict_to_add_to_user(self):
...
+                    return super().__call__(*args)
+                except (KeyError, ConfigError):
+                    return [entry]
+            else:
+                return entry
+
+        return super().__call__(*args)
+
+
+
```
mari-kuri, line 474: 1.  A comment to explain what it does would be nice!
2.  another branch for "plotting", "backend" if it becomes a list
3. I feel it's a little bit convoluted for what is should do, but I wait on the comment to finish the review of this part!

mari-kuri (changes requested):

packages/pymodaq_utils/src/pymodaq_utils/config.py
seb5g: We don't want to have multiple entries for the backend list and the chosen backend. So we use the first in the list as default

this code is to make sure it still works even if you have an old configuration file from 5.1 template...

packages/pymodaq_data/src/pymodaq_data/data.py
seb5g: yes this is using a new template for config where we only use list when we want a user to pick a default.

packages/pymodaq_utils/src/pymodaq_utils/resources/config_template.toml
seb5g: done

packages/pymodaq_utils/src/pymodaq_utils/config.py
seb5g: I just added dynamics that I missed at first round

packages/pymodaq_data/src/pymodaq_data/data.py
malik-irain: This particular entry management is strange, no? It is not a list neither in the GUI nor the toml file, yet as it a backend it gets converted to list internally.

Why not make it a full list then?

packages/pymodaq_utils/src/pymodaq_utils/config.py
malik-irain: With a `T = TypeVar('T')` it would have linked the type of the elements to the same. Basically it means `Any` but all the `T` are the same whereas `Any` can represent different types each time it is used.

packages/pymodaq_utils/src/pymodaq_utils/config.py
```diff
@@ -295,22 +305,14 @@ def to_dict(self):
     def get(self, key: Union[str, Iterable[str]], default=None):
         try:
             return self[key]
-        except KeyError:
+        except (KeyError, ConfigError):
```
malik-irain: I don't think KeyError can happen as:
`self[key]` call `__getitem__`, in which the only way to get `KeyError` is when `self(*item)` is called, but in `__call__` `KeyError` is already caught and transformed in a `ConfigError`

packages/pymodaq_utils/src/pymodaq_utils/config.py
```diff
@@ -384,5 +386,56 @@ def dict_to_add_to_user(self):
         """To subclass"""
         return dict(user=dict(name=USER))
 
+    def __call__(self, *args):
+        if 'backends' in args:
+            try:
+                return super().__call__(*args)
+            except KeyError:
```
malik-irain, line 428: I don't think KeyError can happen as:
`self[key]` call `__getitem__`, in which the only way to get `KeyError` is when `self(*item)` is called, but in `__call__` `KeyError` is already caught and transformed in a `ConfigError`. Unless it can come from somewhere else?

packages/pymodaq_utils/src/pymodaq_utils/config.py
```diff
@@ -384,5 +386,56 @@ def dict_to_add_to_user(self):
...
+               args = replace_item_in_list(args, 'backends', 'backend')
```
malik-irain, line 429: Not coherent with the Typing of replace_item_in_list, args become a tuple when it is defined as a list

packages/pymodaq_utils/src/pymodaq_utils/config.py
malik-irain: Ok, I rewrote it to make it clearer I think:

```python
for plural, singular in [
    ('backends', 'backend'),
    ('debug_levels', 'debug_level'),
    ('dynamics', 'dynamic'),
]:
    #new toml file don't have plural, so check and fallback to singular
    if plural in args:
        try:
            return super().__call__(*args)
        except KeyError:
           args = replace_item_in_list(args, plural, singular)
    # old singular was a text entry, new singular is a list
    # we now expect a list all the times so if it isn't
    elif singular in args:
        entry = super().__call__(*args)
        if isinstance(entry, list):
            return entry
       else:
            try:
                args = replace_item_in_list(args, singular, plural)
                return super().__call__(*args)
            except (ConfigError, KeyError):
                return [entry]

return super().__call__(*args)
```

But you still didn't explain the algorithm, I don't really understand the `elif singular` part. Let's say I'm trying to get 'backends' but it's not in my config, so it will be replaced by 'backend', should the code go directly to `super().__call__(*args)` or also do the `singular` part?

And in the other case where I use `backend`, if it's not a list it tries to make it a list (either by getting the plural `backends` or by making it a list, but is there really a case when only `backend` exists in the config, is not a list and there's no `backends`?

Also, what happens when the config of these entries are modified?

malik-irain (commented):
It's more or less only comments, but it seems to work and passes the tests. I didn't test thoroughly the edge cases with old config files

PR #802, opened 2025-12-03
make sure one can switch between dark and light style

seb5g:
so far once the darkstyle was activated, it was always dark...

### DIFF
--- a/packages/pymodaq_gui/pyproject.toml
+++ b/packages/pymodaq_gui/pyproject.toml
@@ -37,6 +37,7 @@ dependencies = [
     "pyqtgraph>=0.12 , <0.14.0",
     "qtpy",
     "easydict",
+    "qt_themes",
 ]
 
 [project.optional-dependencies]
--- a/packages/pymodaq_gui/src/pymodaq_gui/__init__.py
+++ b/packages/pymodaq_gui/src/pymodaq_gui/__init__.py
@@ -41,16 +41,9 @@ def set_and_check_qt_backend_or_die(config):
 
 def set_check_style():
     current_styles = config('style', 'style')
-    styles_default = ['default', 'dark']
+    styles_default = ['Fusion']
     styles = QtWidgets.QStyleFactory.keys()
-    if 'Fusion' in styles:
-        styles.remove('Fusion')
     styles_default.extend(styles)
-    try:
-        import qt_themes
-        styles_default.extend(list(qt_themes.get_themes().keys()))
-    except ImportError as e:
-        pass
     for style in current_styles[:]:
         if style not in styles_default:
             current_styles.remove(style)
--- a/packages/pymodaq_gui/src/pymodaq_gui/__init__.py
+++ b/packages/pymodaq_gui/src/pymodaq_gui/__init__.py
@@ -46,9 +46,17 @@ def set_check_style():
     if 'Fusion' in styles:
         styles.remove('Fusion')
     styles_default.extend(styles)
+    try:
+        import qt_themes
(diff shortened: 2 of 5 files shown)


Ashwolaa:
It changes!

darkstyle=False
image

darkstyle=True
image

mari-kuri:
The issue is different actually, the code is not really faulty, it's just that the default style in pyqt6 and pyside6 are dark themed so only a little bit of difference are visible. It works with pyqt5!

Your modification feels weird, calling `set_dark_palette` unconditionally to check inside if we should apply the dark style, is the same as calling it conditionally. And the second one is more semantically correct, so it doesn't really change anything

A better way to handle the issue would be to have:

```python
def set_dark_theme(app):
    #the current dark palette + fixup css
    
def set_light_theme(app):
    # same as set_dark_theme but with light colors
    
def set_theme(app, dark=True):
    from qtpy.QtGui import QPalette, QColor
    app.setStyle("Fusion")
    # separated functions so its easier to modify a style
    if dark:
        set_dark_theme(app)
    else:
        set_light_theme(app)

 ...
def mkQApp(name: str):
    app = mkQApppg(name)
    set_theme(app, config('style', 'darkstyle'))
```

But we would have to find a light palette...

seb5g:
dark = True:
image

dark = False:
image


image

For me it work even with qt6... I just read this morning that calling setStyle revert first the Palette to the default then it applies the style and its eventual pallette. So it may be that your default style (on your computer) is dark and that it applies it while on mine it applies a light style??

mari-kuri:
My default theme is dark yes, that's it!

But with pyqt5 it works nicely, I can set and unset darkstyle without it being stuck. Even using pyqt6 or pyside6 it changes (between two dark themes) without being stuck

I guess we have to find a lightstyle to force it's application

Ashwolaa:
I agree with Malik and his suggestions. The different themes should come up as a combobox/list instead of using a checkbox. This would allow people to add their own them if they want. For now, I guess something like:
- default (standard)
- dark 
- light 
would do well.

seb5g:
Fusion Style
The Fusion style is a desktop-oriented style.

The Fusion style is a platform-agnostic style that offers a desktop-oriented look and feel. It implements the same design language as the Fusion style for Qt Widgets.

The Fusion style automatically switches dark and light themes according to the system settings. For information on how to set custom theme colors, refer to the Customization section.

seb5g:
I have an idea and will try it...

seb5g:
https://www.youtube.com/watch?v=JK-B-CT34EU

Ashwolaa:
Our good friend make some themes too
https://github.com/beatreichenbach/qt-themes

seb5g:
Alright new try with all styles from a given machine (either linux or windows, the list should adapt itself to the machine):

image

default is fusion + style from your os
dark is fusion + dark style from Malik's Palette
all the other are light styles from the os (shoudl adapt to the ones from windows or linux dynamically)

seb5g:
good stuff also: who tells him to use qtpy :-DDD

seb5g:
> Our good friend make some themes too https://github.com/beatreichenbach/qt-themes

I'll add them to the ones above!

seb5g:
Just added the qt_themes:
here are some of the ones available

monokai:
image

atome one:
image



Modern light:
image

I like that!! We just need to use the theme colors for the docks label and it would be cool!

malik-irain:
On Linux, using qt-theme I get these:


Configuration entries_003

I guess that's enough and we could rely only on these themes

If they don't make the necessary modifications, we can still fork the project, do the modifications and integrate it in PyMoDAQ one way or another, as the license allows it

PR #804, opened 2025-12-05
Feature/widget synchronizer

Ashwolaa:
Context: Qt aplications often need to keep multiple widgets synchronized (toolbar/menu actions, compact/full views, enable/disable controls). This is usually done with manual signal connections, leading to boilerplate code. In addition, one has to be particularly careful in some situations with bidirectional sync as it can lead to infinite trigger or to desync.

In this PR: I propose a new class `WidgetSync` to deal with all this silly connections. In practice, this permits to easily synchronize multiple widgets with different types or properties. The `WidgetSync` acts as a single source of truth so any widgets or functions can refer to it, aka the truth signal source is the value attribute and any change from it can be captured by the value_changed Signal.

For now, there is no direct use of it in pymodaq but I plan to use it for future developments as I believe in the potential value of this tool to easily manage connections between QWidgets/QActions.

As always, this comes with an example which does not say much as this can be achieved with standard Qt syntax but the docs show more clearly the proposed syntax.

https://github.com/user-attachments/assets/95040437-ec45-4873-aae7-dd79d2e9d20a

### DIFF
--- a/docs/src/developer.rst
+++ b/docs/src/developer.rst
@@ -9,6 +9,7 @@ Developer's Guide
    /developer_folder/plugins
    /developer_folder/custom_app
    /developer_folder/managers
+   /developer_folder/widget_sync
    /developer_folder/whats_new34
    /developer_folder/whats_new45
 
--- a/docs/src/developer_folder/widget_sync.rst
+++ b/docs/src/developer_folder/widget_sync.rst
@@ -1,0 +1,464 @@
+.. _widget_sync:
+
+Widget Synchronization
+======================
+
+.. contents::
+   :local:
+   :depth: 2
+
+Overview
+--------
+
+The ``widget_sync`` module provides a simple, powerful way to synchronize widget properties across multiple Qt widgets.
+It's perfect for keeping toolbar buttons, menu items, and settings panels in sync without manual signal management.
+
+**Key features:**
+
+* Automatic bidirectional synchronization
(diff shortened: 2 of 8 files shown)


seb5g (approved):
that's quite an incredible feast...... I don't see yet the potential for pymodaq indeed but I can imagine we could have a use for that. Actually I do have an idea!!! I'll try and let you know. As it doesn't modify anything on pymodaq yet I'll accept it readily

seb5g:
actually @Ashwolaa how would you synchronised the items AND the current text of a combo box?

Ashwolaa:
This is a bit tedious for now and is something I was looking to dig into for multiple properties.

You would need two syncs for now. One for the items and one for the index/current text. I will forward you an example.

Envoyé : vendredi 5 décembre 2025 22:31
Cc : Ashwola ***@***.***>; Mention ***@***.***>
Objet : Re: [PyMoDAQ/PyMoDAQ] Feature/widget synchronizer (PR #804)

[https://avatars.githubusercontent.com/u/6609955?s=20&v=4]seb5g left a comment (PyMoDAQ/PyMoDAQ#804)<https://github.com/PyMoDAQ/PyMoDAQ/pull/804#issuecomment-3618665224>

actually @Ashwolaa<https://github.com/Ashwolaa> how would you synchronised the items AND the current text of a combo box?

PR #806, opened 2025-12-05
to create a tree from a given start point in the config

seb5g:
@malik-irain could you incorporate this in your work on the treefromtoml?


malik-irain:
Done!

PR #807, opened 2025-12-05
switch icons given the checked status

seb5g:
if the action is checkable and the keywaord argument icon_checked is a valid icon creation argument, the action icon will switch between two icons given the checked status

### DIFF
--- a/packages/pymodaq_gui/src/pymodaq_gui/managers/action_manager.py
+++ b/packages/pymodaq_gui/src/pymodaq_gui/managers/action_manager.py
@@ -47,8 +47,6 @@ class QAction(QtQAction):
         super().__init__(*args, **kwargs)
 
         if icon_checked is not None and not isinstance(icon_checked, QtGui.QIcon):
-            if icon_checked == 'unshow':
-                pass
             icon_checked = create_icon(icon_checked)
 
         self.icon_checked = icon_checked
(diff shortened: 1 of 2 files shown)


packages/pymodaq_gui/src/pymodaq_gui/managers/action_manager.py
```diff
@@ -12,32 +12,50 @@
...
+                 icon_checked: Union[str, QtGui.QIcon] = None, **kwargs):
+        if icon_unchecked is not None:
+            icon_unchecked = create_icon(icon_unchecked)
+            args = list(args)
+            args.insert(0, icon_unchecked)
         super().__init__(*args, **kwargs)
 
+        if icon_checked is not None and not isinstance(icon_checked, QtGui.QIcon):
+            if icon_checked == 'unshow':
+                pass
```
Ashwolaa: What is this weird condition? If None, do nothing

packages/pymodaq_gui/src/pymodaq_gui/managers/action_manager.py
```diff
@@ -91,14 +114,17 @@ def addaction(name: str = '', icon_name: Union[str, Path, QtGui.QIcon]= '', tip=
         Using this shortcut will trigger the action
     enabled: bool
         set the enabled state
+    icon_checked: : str / Path / QtGui.QIcon / enum name
+        str/Path: the png file name/path to produce the icon
+        QtGui.QIcon: the instance of a QIcon element
+        ThemeIcon enum: the value of QtGui.QIcon.ThemeIcon (requires Qt>=6.7)
+        Optional, if set, will be the icon when the action is checked
```
Ashwolaa: In theory, one could add a warning that checkable needs to be activated or activate if by default

packages/pymodaq_gui/src/pymodaq_gui/managers/action_manager.py
seb5g: that was a silly thing to add a debug stop point here!! I removed it

PR #809, opened 2025-12-06
craeting an example for double sync of combobox

seb5g:


seb5g:
@Ashwolaa I think I got the gist of it and created the double sync for combo box

packages/pymodaq_gui/src/pymodaq_gui/examples/widget_sync_example.py
```diff
@@ -718,6 +721,14 @@ def create_custom_widget_example(self):
             setter=lambda c: self.color_combo.setCurrentText(c)
         )
 
+        # Bind ComboBox
+        self.color_sync.bind(
+            self.color_combo_bis,
+            signal=self.color_combo_bis.currentTextChanged,
+            getter=lambda: self.color_combo_bis.currentText(),
+            setter=lambda c: self.color_combo_bis.setCurrentText(c)
+        )
```
Ashwolaa, line 730: I think you forgot to fill it with items in this one.

packages/pymodaq_gui/src/pymodaq_gui/examples/sync_combo.py
```diff
@@ -0,0 +1,84 @@
...
+        if index is None:
+            index = self.index_sb.value()
+        self.combos[index].items = []
+
+    def set_items(self, list_elt: list[str] = None, index: int = None):
+        if index is None:
+            index = self.index_sb.value()
+        if list_elt is None or list_elt is False:
+            list_elt = [f'elt {elt:03.0f}' for elt in range(index+1)]
+        self.combos[index].items = list_elt
```
Ashwolaa, line 70: This does not work so well here. As the index can be above 2 which is the number of ComboBox you have. Above it, the index is out of range.

Ashwolaa:
Yes that's it! Very nice. Careful with your example. You cannot go above N=3 in your example.

 I will come up with an additional extension that allows a single sync to manage multiple values through a dict approach. 
For example, one can either access:
- items / index 
- Value + Unit + range

seb5g:
It's not meant to be an example to merge but just to check with you that's the way of doing it and if it's correct I will just add this combo box in the widget list so we can use it easily in other applications.  I was thinking in particular to the combo box in configurator manager and preset manager where it's difficult to keep the sync now it will be easy

Ashwolaa:
Please wait a bit with it, I will soon PR an additional update that will simplify a lot the syntax and allow syncing multiple properties

seb5g:
no worries I don't want to merge it , I'm just referencing everything related to it!

PR #818, opened 2025-12-11
Bugfix/daq move dataset

### LINKED ISSUE #815: [BUG] Fake Dataset when initiliazing dashboard
### Describe the bug

Some h5 files are created when the dashboard gets launched. One dataset per instrument. See below

This also triggers when launching a DAQ_move or a DAQ_viewer (from daq_move / daq_viewer)

### Expected Behavior

No h5 file should be created.

### Steps To Reproduce

1.  Launch dashboard
2. Load Preset
3. Check you Data folder

https://github.com/user-attachments/assets/0538e951-92b4-4ab6-996d-9ee621ee6fd9

### Reference(s)

```markdown
pymodaq.utils.config.deep_update
```

### Environment

```markdown
- OS:Windows 11
- Pymodaq: 5.1 and dev
- python env:3.13
```

### Additional context

_No response_

seb5g:
fixes #815 

The DAQ_Move created at instanciation an instance of a H5Saver to store DataActuator data when looping the get_actuator_value. But it's not needed at startup, it should be started only when looping.

Data were not even saved... now they are stored everytime one clicks the loop action

### DIFF
--- a/packages/pymodaq/src/pymodaq/control_modules/daq_move.py
+++ b/packages/pymodaq/src/pymodaq/control_modules/daq_move.py
@@ -20,6 +20,7 @@ from qtpy import QtWidgets
 
 from easydict import EasyDict as edict
 
+from pymodaq.utils.h5modules.module_saving import ActuatorSaver
 from pymodaq_utils.logger import set_logger, get_module_name
 from pymodaq_utils.utils import find_keys_from_val
 from pymodaq_utils import utils
@@ -166,7 +167,6 @@ class DAQ_Move(ParameterControlModule):
         if len(ACTUATOR_TYPES) > 0:  # will be 0 if no valid plugins are installed
             self.actuator = kwargs.get("actuator", ACTUATOR_TYPES[0])
 
-        self.module_and_data_saver = module_saving.ActuatorTimeSaver(self)
 
         self._move_done_bool = True
 
@@ -564,6 +564,10 @@ class DAQ_Move(ParameterControlModule):
                     "show_graph"
                 ):
                     self.ui.show_data(DataToExport(name=self.title, data=[data_act]))
+                if self.ui.has_action("refresh_value") and self.ui.is_action_checked('refresh_value'):
+                    data_node = self.module_and_data_saver.get_set_node()
+                    self.module_and_data_saver.add_data(data_node, DataToExport('Actuator', data=[data_act]))
+
             self._current_value = data_act
             self.current_value_signal.emit(self._current_value)
             if (
@@ -685,12 +689,14 @@ class DAQ_Move(ParameterControlModule):
         The current timer period is set by the refresh value *'refresh_timeout'* in the actuator main settings.
         """
         if get_value:
+            self.module_and_data_saver = module_saving.ActuatorTimeSaver(self)
             self._refresh_timer.setInterval(
                 self.settings["main_settings", "refresh_timeout"]
             )
             self._refresh_timer.start()
         else:
             self._refresh_timer.stop()
+            self.module_and_data_saver.h5saver.close_file()
 
     @property
     def actuator(self):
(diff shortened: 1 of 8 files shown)


packages/pymodaq/src/pymodaq/control_modules/daq_move.py
```diff
@@ -20,6 +20,7 @@
 
 from easydict import EasyDict as edict
 
+from pymodaq.utils.h5modules.module_saving import ActuatorSaver
```
Ashwolaa: Why import if not used?

Ashwolaa (commented):
Looks good on usage, no extra dataset with DAQ_move. The update value seems to behave correctly

seb5g:
worked also on the daq_viewer dataset creation at startup. A small polish on the daq_move side still to do to replicate some work done on the viewer and we'll be ok.

@Ashwolaa do you think feasible a widgetsync for parameters that would sync their widget tree representation and their values that would ease a lot of implementation in pymodaq.

packages/pymodaq/src/pymodaq/control_modules/daq_move.py
seb5g: some automatic import added by pycharm I guess...

Ashwolaa:
> worked also on the daq_viewer dataset creation at startup. A small polish on the daq_move side still to do to replicate some work done on the viewer and we'll be ok.
> 
> @Ashwolaa do you think feasible a widgetsync for parameters that would sync their widget tree representation and their values that would ease a lot of implementation in pymodaq.

I will look into it, I just PR an extension to dict base entries into the syncing mechanism so I am fairly optimistic but we should discuss about it.

seb5g:
@Ashwolaa the PR is finished and solved the issues. Since yesterday I incorporated a mechanisms to tell the DAQ_Move to log its values independently of the refresh state/button.

Could you check all this?

Ashwolaa:
When trying to launch a scan, I receive the error
ERROR:pymodaq.daq_scan:'DAQ_Viewer' object has no attribute 'module_and_data_saver'
  File "D:\Work\PyMoDAQ\packages\pymodaq\src\pymodaq\extensions\daq_scan.py", line 797, in set_scan
    res = self.update_scan_info()
  File "D:\Work\PyMoDAQ\packages\pymodaq\src\pymodaq\extensions\daq_scan.py", line 575, in update_scan_info
    scan_node = self.module_and_data_saver.get_set_node(new=False)
  File "D:\Work\PyMoDAQ\packages\pymodaq\src\pymodaq\utils\h5modules\module_saving.py", line 414, in get_set_node
    module.module_and_data_saver.main_module = False
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'DAQ_Viewer' object has no attribute 'module_and_data_saver'

Ashwolaa:
> @Ashwolaa the PR is finished and solved the issues. Since yesterday I incorporated a mechanisms to tell the DAQ_Move to log its values independently of the refresh state/button.
> 
> Could you check all this?

I am not sure to understand how it should work. When using the RefreshValue, nothing is being recorded. Should it appear in the C:\Data? aka the filepath given in config?

seb5g:
here a small use case. It is the same usage as continuous saving from the daq_viewer. Contrary to what I said above, the new implementation need you to tick `do_save` to trigger a "continuous saving", a king of single module logger. Then whenever get_actuator_value is called, the value is saved (works also with the refresh obviously) 


https://github.com/user-attachments/assets/9e1a50bf-8e0d-4926-a163-01726b3f3eaa

and I fixed the issue with the daq_scan and all the files that were also created with it. FOr future records, it is related to the way the property: module_and_data_saver is encoded (creating H5 files when calling the property and if no H5 has been set beforehand). This is what I do here by using not the property but the _module_and_data_saver variable.

I don't like too much this patch but I feel like the DAQ Scan would need a full cleanup in the future anyway...

Do you see other bugs?

PR #819, opened 2025-12-11
Feature/widget synchronizer: Expanding to dict and more

Ashwolaa:
This PR is the follow up of Feature/widget synchronizer #804 

Sorry it is huge, but it tries to do a lot and to stay fairly flexible/accessible in terms of usage outside of the PyMoDAQ's scope.

Its main interest is the expansion of the widget_synchronizer to dict entries. As such, it now provides a way to use a single object to synchronize combobox (items + sel) or in more complex scenarion (value + units), ...

There is a large number of examples that are provided to see different use cases. 

The main goal is to profit from it to easily connect the relevant parts in Pymodaq
==) Connected status propagated to all widgets/actions
==) Facilitate units conversion for actuators (hopefully)
==) Possibility to easily add/remove temporary widget that are synchronized 

I think it is best to navigate through the examples and the doc to check it. I will try to put it in action after the PR is accepted. For now, it is still harmless throughout PyMoDAQ's codebase.


seb5g (approved):
excellent!

seb5g:
Did you think about creating a separated package for that? It is entirely out of scope of pymodaq by itself even if I'm sure we will use it in pymodaq more and more in the future. I'm ok to leave it here but that could be useful elsewhere without the need of pymodaq.

Ashwolaa:
I was thinking indeed to do it as a separate library initially but after some thoughts, it was easier to stick first with PyMoDAQ to facilitate the first debugging steps and also maximize the practical uses in it throughout the code.

But I agree that whence it is in a stabilized form, it would be better to get it out and publish it separately.

PR #829, opened 2025-12-16
Feature/actuator UI config

### LINKED ISSUE #826: Use lists in config editor for Actuator Ui / Debug Level
Hi,

**Is your feature request related to a problem? Please describe.**
Currently, switching between Original / Simple / ... UI for daq moves is done here

Image

The user does not know what are the available options.

The debug level does it like this
Image
It is a bit confusing (I've seen people changing the "debug_levels" drop-down list with no effect) but at least the available options are here. 

**Describe the solution you'd like**
Replace both Actuator/Ui and General/Debug_level by a list with acceptable options!

seb5g:
fix  #826 partially on the UI level

for the rest see #799

### DIFF
--- a/packages/pymodaq/src/pymodaq/__init__.py
+++ b/packages/pymodaq/src/pymodaq/__init__.py
@@ -48,6 +48,7 @@ try:
         config = Config()  # to ckeck for config file existence, otherwise create one
         copy_preset()
 
+
         from pymodaq_utils.config import Config
         from pymodaq.utils.scanner.utils import register_scanners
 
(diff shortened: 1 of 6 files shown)


rgeneaux:
sorry for the noob question, but how can I quickly install from git directly now?
`pip install git+https://github.com/PyMoDAQ/PyMoDAQ@feature/actuator_ui_config`
doesn't work since it doesn't find the setup.py or pyporject.toml

seb5g:
You have to fetch/clone the PR code (see code button up right on that webpage) then in the root of the repository in your console:
```
python install-packages.py
```
this will install in editor mode all 4 packages (I guess we can modify the install-packages script to fetch a PR and install the code from it. @malik-irain ?)

seb5g:
and if you're already in editor mode, you could checkout directly from the upstream branch (but I'm not fluent enough with git to be sure)

rgeneaux:
I am getting some error which probably doesn't have anything to do with the PR
```
File "C:\...\PyMoDAQ\packages\pymodaq\src\pymodaq\control_modules\instruments.py", line 9, in <module>
    from pymodaq import CONTROL_MODULES
ImportError: cannot import name 'CONTROL_MODULES' from 'pymodaq' (C:\...\PyMoDAQ\packages\pymodaq\src\pymodaq\__init__.py)
```
I won't have much time to debug this so maybe don't wait for me to accept the PR!

seb5g:
ok i'll let malik check quickly before merging

malik-irain:
I got the same error when trying to start the dashboard in a fresh env, by cloning, switching to this branch and installing with the script

malik-irain:
It seems to come from pymodaq `__init__.py` file, when check for ui:

```python

        if not isinstance(uis_config, list):
            uis = [uis_config]
        for ui in uis_registered:
            if ui not in uis_config:
                uis_config.append(ui)
        pymodaq_config['actuator', 'ui'] = uis_config
        pymodaq_config.save()
 ```
 
 I suppose `uis = [uis_config]` should be `uis_config = [uis_config]`

rgeneaux:
Everything good after your commit @malik-irain 
thanks 
image

seb5g:
indeed!!

PR #833, opened 2025-12-21
Bugfix/configurator

seb5g:
solving last bugs


seb5g:
hi @Ashwolaa I tried, in order to simplify combos sync, to use widget sync for my 2 combos. Current value is fine however both creating an entry of deleting one do not synchronize the two combobox?? I update the entries_sync value attribute to do so but to no avail... I do not understand what is interfering with the process of synchronization. Could you have a llok at the last commit of this ongoing PR? cheers

Ashwolaa:
Hey, I found a little bit of time to look into it. 
I made a PR #835 that adresses it. 
The main issue comes from the check of change in `set_value`. The internal value isactually changed when updating elements in the list due to shallow copies and it was always giving False. The signal for the widget was then never emitted.
It should be good now. I recommend using some of the new helpers method.
```
update_key
append_to_list
remove_from_list
pop_from_list
```

seb5g:
thks @Ashwolaa and merry Xmas

seb5g:
Thx for working on it!

Tries #835 with this issue to no avail. I used:

```
self.entries_sync.append_to_list('items', entry)
self.entries_sync.update_key('current', entry)
```

but still the items in the other combo are not updated!

I created a derived combo class to have specific getter/setter and check whether they are called or not. Following the append/set_value there is only one call to the setter: only one widget has its values updated. Don't get why it's working with currentText on not settings items. And I don't understand what is different compared to your examples...

Ashwolaa:
Damn... Merry Christmas! I will look into this more deeply in the next few days

seb5g:
actually I'm wondering whether the issue comes from the fact that the addwidget method is returning a proxy class to both the action and the underlying widget. Maybe this is messing with the signal mechanisms? But don't know why currentText is  working then and not items...

seb5g:
I GOT IT!!
it was indeed the proxyclass that was messing, passing the underlying widget to the sync class is working just fine!!!

PR #837, opened 2026-01-03
Feature/common window to encapsulate common actions (menu and toolbar)

seb5g:
Most of the standalone modules should/do share common actions: 

* close
* show log
* show config
* ...

It means the code can be refactored to use only one instance of such actions.

This PR introduces a CustomApp inherited object named SharedUI. It instantiate or use a QMainWindow in which it adds common actions in the form of toolbars and or menubar:
image

Then the central widget of the menubar will hold the UI content of the app one want to load:

* DashBoard
* DAQ_Move
* DAQ_Viewer

For instance, to create the SharedUI and a DAQ_Move:

```python
widget = QtWidgets.QWidget()
daq_move = DAQ_Move(widget, title="test")

shared_ui = SharedUI(widget)
shared_ui.affect_application(daq_move)
```

Here we use the `affect_application` method to connect the shared `Quit` action with the DAQ_Move `quit_fun` method and the proper restart of the app!

While removing the common action, there was not much left for the DAQ_Viewer UI settings. This PR is therefore also merging this almost empty Dock into a way compacter toolbar:

image

One see the SharedUI and the new toolbar (using MaterialIcons on the way)! Detector selection is also simplified by using the same mechanism as within the preset manager:

image

The `SaveNew` action has been dropped and the background ones modified to be simple actions in the toolbar

Below is the result for a loaded Preset in the DashBoard (with a mix of simple/original Actuator UIs)

image

Of course, further PR/discussion on the UX can be done @runrunm and @Ashwolaa but I think this is a good start for what we discussed last time


Ashwolaa:
great job, this looks very nice. I will look at it in the coming days!

packages/pymodaq/src/pymodaq/control_modules/daq_viewer_ui/ui_base.py
```diff
@@ -0,0 +1,361 @@
...
+    tree = ParameterTree()
+    tree.setParameters(param, showTop=False)
+    timer = QtCore.QTimer()
+
+
+
+    widget = QtWidgets.QWidget()
+    prog = DAQ_Viewer_UI(widget, title='myViewer')
+    shared_ui = SharedUI(prog, app_class_file=__file__)
+
```
Ashwolaa: This does not run, app_class_file is not expected

packages/pymodaq/src/pymodaq/control_modules/daq_move_ui/ui_base.py
```diff
@@ -220,7 +232,7 @@ def setup_docks(self):
         self.abs_value_sb_2.setStyleSheet("background-color : lightcoral; color: black")
 
         self.abs_value_sb_bis = QSpinBoxWithShortcut(step=0.1, dec=True, siPrefix=config('actuator', 'siprefix'))
-        self.ini_actuator_pb = PushButtonIcon('ini', 'Initialization', checkable=True,
+        self.ini_actuator_pb = PushButtonIcon('cable', 'Initialization', checkable=True,
```
Ashwolaa, line 235: This layout feels a bit off compared to the compact versions. I think we should keep the 
Name / Plugin selection / Initialization / (Settings?) as the standard approach for any instrument.

packages/pymodaq/src/pymodaq/control_modules/daq_viewer_ui/ui_base.py
```diff
@@ -0,0 +1,361 @@
...
+    def setup_actions(self):
+        self.add_widget('name', QtWidgets.QLabel(self.title))
+        self.get_action('name').widget.setFont(
+            create_font(font_name="Tahoma",
+                        font_size=14, isbold=True, isitalic=True))
+        self.add_widget('selector', self.selector.add_widget)
+        self.add_action('ini_detector', 'Ini. Detector', ActionIconNames.INI, checkable=True,
+                        tip='Connect to selected detector', icon_color=qt_themes.get_theme().red,
+                        )
+        self.add_action('grab', 'Grab', 'videocam', "Grab data from the detector", checkable=True,
```
Ashwolaa, line 147: I would add a separator here (and maybe put the settings before the grab).
In fact, I am quite inclined to have the settings very close to the init button as some plugins need to select the connection port after start up.

packages/pymodaq/src/pymodaq/control_modules/daq_viewer_ui/ui_base.py
```diff
@@ -0,0 +1,361 @@
...
+                        icon_checked='videocam_off')
+        self.add_action('snap', 'Snap', ActionIconNames.SNAP, "Take a snapshot from the detector")
```
Ashwolaa, line 149: Not my favorite icon ^^
The behavior is a bit strange here, when grabbing and then snapping, the grab icon stays ticked.
I'd suggest to disable the snap icon when using grab and vice/versa

packages/pymodaq/src/pymodaq/control_modules/daq_viewer_ui/ui_base.py
Ashwolaa: I would also add some colors, I love this green / orange part

packages/pymodaq/src/pymodaq/control_modules/daq_viewer_ui/ui_base.py
```diff
@@ -0,0 +1,361 @@
...
+        self.add_action('save_current', 'Save Current Data', 'save_as', "Save Current Data")
+        self.toolbar.addSeparator()
+        self.add_action('background_snap', 'Snap Background', 'background_replace',
+                        tip='Take a snapshot a set it as background')
+        self.add_action('background_subtract', 'Subtract Background', 'texture_minus', checkable=True,
+                        tip='If checked, apply background substraction',
+                        icon_checked_color=qt_themes.get_theme().green)
+        self.add_action('show_settings', 'Show Settings', 'settings', "Show Settings", checkable=True)
+
```
Ashwolaa, line 160: I would add an extra action which would hide/show the viewer(s) similarly as for the show_graph in the daq_move

packages/pymodaq_gui/src/pymodaq_gui/utils/utils.py
```diff
@@ -346,3 +346,15 @@ def mkQApp(name: str):
...
+def create_font(font_name=None, font_size=None, isbold=False, isitalic=False) -> QtGui.QFont:
+    font = QtGui.QFont()
+    if font_name is not None:
+        font.setFamily(font_name)
+    if font_size is not None:
+        font.setPointSize(font_size)
+
+    font.setBold(isbold)
+    font.setItalic(isitalic)
+    return font
```
Ashwolaa: out of topic in this PR but the create_icon / create_color of the action manager could be moved also in this utils.py file

By the way, there is a widget LabelWithFont that could use this method.

packages/pymodaq_utils/src/pymodaq_utils/utils.py
```diff
@@ -664,6 +664,14 @@ def get_new_file_name(base_path=Path(config('data_saving', 'h5file', 'save_path'
     file = f'{base_name}_{index:03d}'
     return file, curr_dir
 
+def get_module_path(module_name: str):
+    """ Get Path of a given module represented by its name
+
+    You get the name usually with the dunder __module__ on an object or class
+    """
+    module = sys.modules[module_name]
+    return module.__file__
```
Ashwolaa: expecting Path(module.__ file __) from the docstring

packages/pymodaq/src/pymodaq/control_modules/control_module_selector.py
```diff
@@ -0,0 +1,150 @@
...
+        """Called when a menu item is selected from the nested add menu
+
+        To be subclassed for particular signal emission
+
+        """
+        # Call the parameter's addNew method with the selected type
+        self.add_widget.setText('/'.join(path_tuple))
+        self.add_widget.adjustSize()
+        self.module_changed.emit(path_tuple)
+
```
Ashwolaa, line 83: Since this is also used by the GroupParameter, maybe we could factorize the methodology in a utils.py to be applicable to any Qmenu

packages/pymodaq/src/pymodaq/dashboard.py
```diff
@@ -26,6 +26,8 @@
 from pymodaq_plugin_manager.manager import PluginManager
 from pymodaq_plugin_manager.validate import get_pypi_pymodaq
 
+from pymodaq.control_modules.daq_viewer_ui.viewer_selector import SelectedModule
+from pymodaq.utils.shared_ui import SharedUI
```
Ashwolaa: Unused import

packages/pymodaq/src/pymodaq/control_modules/daq_viewer_ui/ui_base.py
```diff
@@ -0,0 +1,361 @@
...
+        self.enable_actions(status, all_except=('ini_detector', 'selector', 'show_settings'))
+        self.set_action_enabled('selector', not status)
+
+        if status:
+            icon = create_icon(ActionIconNames.INI,
+                               icon_color=qt_themes.get_theme().green,)
+        else:
+            icon = create_icon(ActionIconNames.INI,
+                               icon_color=qt_themes.get_theme().red,)
+        self.get_action('ini_detector').set_icon(icon)
```
Ashwolaa, line 293: - Might be worth to use a common init function for detector/actuator. Then use a pre/post_init for modules specific such as enabling actions, and so on
- Wouldn't it be easier to use a checked/unchecked icon/icon_color by the way?

packages/pymodaq/src/pymodaq/control_modules/daq_viewer_ui/ui_base.py
```diff
@@ -0,0 +1,361 @@
...
+            self.command_sig.emit(ThreadCommand(UiToMainViewer.DETECTOR_CHANGED, sel_mod))
+            if self.viewer_types != [sel_mod.daq_type.to_viewer_type()]:
+                self.update_viewers([sel_mod.daq_type.to_viewer_type()])
+        except ValueError as e:
+            pass
+
+    def show_settings(self, show=True):
+        if (self.is_action_checked('show_settings') and not show) or \
+                (not self.is_action_checked('show_settings') and show):
+            self.get_action('show_settings').trigger()
```
Ashwolaa, line 218: Similar comment as for init, this should be a common feature so a method in control_modules_ui

packages/pymodaq/src/pymodaq/control_modules/daq_viewer_ui/ui_base.py
```diff
@@ -0,0 +1,361 @@
...
+        if self.is_action_checked('grab'):
+            self.get_action('grab').trigger()
+
+    def send_init(self, checked: bool):
+        self._enable_detchoices(not checked)
+        self.command_sig.emit(ThreadCommand(UiToMainViewer.INIT,
+                                            [checked,
+                                             self.selector.selected_module]))
+        if not checked and self.is_action_checked('background_subtract'):
+            self.get_action('background_subtract').trigger()
```
Ashwolaa: Why not but I would be more inclined to disable any actions except settings when the detector is not initialized

packages/pymodaq/src/pymodaq/utils/managers/modules_manager.py
```diff
@@ -571,6 +571,17 @@ def reset_signals(self):
         self.move_done_flag = True
         self.det_done_flag = True
 
+    def poll_init(self, module):
+        is_init = False
+        tstart = time.perf_counter()
+        while not is_init:
```
Ashwolaa: Why not use ?
while not module.initialized_state:

packages/pymodaq/src/pymodaq/utils/managers/modules_manager.py
```diff
@@ -571,6 +571,17 @@ def reset_signals(self):
...
+            QThread.msleep(1000)
+            QtWidgets.QApplication.processEvents()
+            is_init = module.initialized_state
+            if time.perf_counter() - tstart > 60:  # timeout of 60sec
```
Ashwolaa: This 60 is fine but arbitrary, value could be chosen in config.

Ashwolaa (commented):
Here is a first review, mostly code organization and some layouts questions. Very nice anyway, this is very close to the layout I was imagining at first so I am super pleased by this PR!

Ashwolaa (changes requested):
Here is a first review, mostly code organization and some layouts questions. Very nice anyway, this is very close to the layout I was imagining at first so I am super pleased by this PR!

Ashwolaa:
The current menu is not so great now though. Some naming conventions should be chosen to make it clearer.
image

packages/pymodaq/src/pymodaq/control_modules/daq_viewer_ui/ui_base.py
seb5g: corrected

packages/pymodaq/src/pymodaq/control_modules/daq_move_ui/ui_base.py
seb5g: this is only for stand alone as it feels a bit empty if everything is solely in the toolbar. Another idea is to keep it minimal and show below the plot, really just like a daq_viewer?

packages/pymodaq/src/pymodaq/control_modules/daq_viewer_ui/ui_base.py
seb5g: done

packages/pymodaq/src/pymodaq/control_modules/daq_viewer_ui/ui_base.py
seb5g: our UX manager @runrunm will come up with some nicer icons I'm sure!!

Ok for the disabling

packages/pymodaq/src/pymodaq/control_modules/daq_viewer_ui/ui_base.py
seb5g: done

packages/pymodaq_gui/src/pymodaq_gui/utils/utils.py
seb5g: both are done

packages/pymodaq_utils/src/pymodaq_utils/utils.py
seb5g: true changed the return

packages/pymodaq/src/pymodaq/control_modules/control_module_selector.py
seb5g: agreed, I will let you do so just after the PR! ok?

packages/pymodaq/src/pymodaq/dashboard.py
seb5g: removed

packages/pymodaq/src/pymodaq/control_modules/daq_viewer_ui/ui_base.py
seb5g: not so simple, here I don't want the icon to change when I click but only when I'm sure the detector/actuator is initialized... Do you see another solution?

packages/pymodaq/src/pymodaq/control_modules/daq_viewer_ui/ui_base.py
seb5g: actually this one I'm not sure it is used ... anymore. Let's have a round of refactorisation between the two out of this PR...

packages/pymodaq/src/pymodaq/control_modules/daq_viewer_ui/ui_base.py
seb5g: this is to prevent to start a new detector with background substration activated without a right background. Then it is also disabled

packages/pymodaq/src/pymodaq/utils/managers/modules_manager.py
seb5g: sure :-D

packages/pymodaq/src/pymodaq/utils/managers/modules_manager.py
seb5g: done

packages/pymodaq/src/pymodaq/control_modules/daq_viewer_ui/ui_base.py
Ashwolaa: - The icons are opposite / flipped between daq_move and daq_viewer
- I was expecting that we can hide the whole dock while it only hides the inner dockarea here

packages/pymodaq/src/pymodaq/control_modules/daq_viewer_ui/ui_base.py
Ashwolaa: We can leave it for this now for this PR as it behaves as exepected. I might look at it later

seb5g:
> The current menu is not so great now though. Some naming conventions should be chosen to make it clearer. image

I made it such like if menus have same title (within sharedui and app) they will be merged together. In a next UI round we'll have to add actions to show/hide logger dock and remote dock.

seb5g:
@Ashwolaa I completed/updated all the things you mentioned. Do you want to have a final look at it then we'll merge it and open another one to fine tweak the icons and other UX things!

Ashwolaa (approved):
Looks good on usage overall, let's merge

PR #845, opened 2026-01-08
patching the signal signature for pyside6

seb5g:
When using the tree_layout widget (pymodaq_gui) used for instance in the h5browser, then simple click or double click is raising an error in pyside6 due to signal signature mismatch (which is not the case for pyqt6...)

This PR patch this together with a test suite !


seb5g:
to be ported on 5.1.x branch also

malik-irain:
Good for me, I removed the yielded `qtbot`  from `tree` fixture as:

1. It's more easy to understand and make more "beautiful" code 
2. We don't use qtbot variable
3. In the official `qtbot` implementation, they don't do it in similar situations and add the qtbot fixture as a parameter of their test when needed:
https://github.com/pytest-dev/pytest-qt/blob/master/tests/test_screenshot.py#L8-L26

malik-irain (approved):

PR #861, opened 2026-01-16
widget editor in table view

seb5g:
the connection creates two modification the call to setData is done when the widget is done editing


malik-irain:
Some modifications are still needed a space is added at the end of text values when an entry is selected and is validated when no editions are made

PR #868, opened 2026-01-19
Toolbar uses title from CustomApp constructor as self.title is not defined

malik-irain:


malik-irain:
@seb5g  I modified #864 to use title from the constructor instead of self.title, otherwise Dashboard couldn't be loaded. I don't know if that's what you needed the modification for.

Anyways, in PyMoDAQ source code, title is never provided to CustomApp...

seb5g:
Could you check in the latest as I included it this morning. 
```python
If title is none: 
    title = self.__class__.__name__

self.title = title
```

Sorry for the case , I'm on my phone in the plane... That is 4 hours late

PR #869, opened 2026-01-19
Create title attribute in CustomApp that defaults to class name

malik-irain:
@seb5g the logic to check for title wasn't included in any of the accepted PRs


seb5g:
strange maybe it's on the one still opened... anyway thx

PR #877, opened 2026-01-19
Allow to load dashboard with preset and config

malik-irain:
It is a fix and a feature:

1. Loading a preset from command line was broken, it just set the selected preset to the one provided but didn't load it. This is fixed.
2. We can also tell the program to load a specific configuration (after the default one that is always automatically loaded)

### DIFF
--- a/packages/pymodaq/src/pymodaq/dashboard.py
+++ b/packages/pymodaq/src/pymodaq/dashboard.py
@@ -1631,9 +1631,14 @@ def main():
     app = mkQApp('Dashboard')
 
     # Command-line argument parsing
-    parser = argparse.ArgumentParser(prog="dashboard", description="PyMoDAQ dashboard")
-    parser.add_argument("-p", "--preset", metavar="PRESET_NAME", help="preset name to load")
-    parser.add_argument("-c", "--config", metavar="CONFIG_NAME", help="config name to execute")
+    parser = argparse.ArgumentParser(prog="dashboard",
+                                     description="PyMoDAQ dashboard. "
+                                                 "Command-line options only affect GUI initial state."
+                                     )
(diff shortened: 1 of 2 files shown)


packages/pymodaq/src/pymodaq/utils/gui_utils/loader_utils.py
```diff
@@ -45,13 +45,15 @@ def load_dashboard_with_preset(preset_name: str, extension_name: str = None)  ->
...
+    preset_name = preset_path.stem
     extension = None
 
     if preset_name in dashboard.preset_manager.entries:
-        dashboard.preset_manager.entry = preset_name
-
+        dashboard.preset_manager.execute_entry_base(preset_path)
+        if configuration_name is not None:
+            configuration_path = get_set_configurator_path().joinpath(preset_name).joinpath(f'{configuration_name}.config')
+            dashboard.configurator.execute_entry(configuration_path)
```
seb5g: I would say it's better to call execute_entry_base just like for the preset!

packages/pymodaq/src/pymodaq/dashboard.py
```diff
@@ -1633,11 +1633,12 @@ def main():
     # Command-line argument parsing
     parser = argparse.ArgumentParser(prog="dashboard", description="PyMoDAQ dashboard")
     parser.add_argument("-p", "--preset", metavar="PRESET_NAME", help="preset name to load")
+    parser.add_argument("-c", "--config", metavar="CONFIG_NAME", help="config name to execute")
```
seb5g: maybe add in the help that if config is supplied then preset is mandatory

seb5g (changes requested):

seb5g (approved):

PR #881, opened 2026-01-20
Feature/daq scan update

seb5g:
Applies the CustomApp/EXt inheritance and uses the SharedUI concept together with new MaterialIcons


seb5g:
hi @Ashwolaa 
this is a first round of UX application to daq_scan/dashboard. 

We will still need to discuss again with @runrunm in order to select the final icons, here is just a proposal (same for colors that should be coherent also throughout all the modules: like settings in blue, execution of stuff in magenta, ok state in green, stop/close in red... @runrunm did you make any progress along these lines

packages/pymodaq_gui/src/pymodaq_gui/managers/manager_base.py
```diff
@@ -258,36 +258,45 @@ def setup_actions_base(self):
...
-        self.add_action(ManagerActions.COPY, f'Copy {self.entry_type.capitalize()}', 'EditCopy')
+        self.add_action(ManagerActions.COPY, f'Copy {self.entry_type.capitalize()}',
+                        'file_copy')
         self.add_action(ManagerActions.NEW,
-                        f'New {self.entry_type.capitalize()}', 'ListAdd',
+                        f'New {self.entry_type.capitalize()}', 'add_circle',
                         tip=f'Create a new {self.entry_type} file ("Ctrl+N")',
-                        shortcut=QKeySequence(Qt.Modifier.CTRL | Qt.Key.Key_N))
+                        shortcut=QKeySequence(Qt.Modifier.CTRL | Qt.Key.Key_N),
+                        icon_color=self.get_theme().green,)
```
Ashwolaa, line 268: out of topic but create/copy are to close currently.
I suggest changing it such that create starts with a blank preset since this action is currently lacking.

packages/pymodaq/src/pymodaq/extensions/scan/daq_scan_ui.py
```diff
@@ -46,14 +46,14 @@ def __init__(self, parent):
...
+        self.add_action('start', 'Start Scan', 'motion_play', "Start the scan",
+                        menu=self.action_menu, icon_color=self.get_theme().green)
         self.add_action('start_batch', 'Start ScanBatches', 'run_all', "Start the batch of scans", menu=self.action_menu)
-        self.add_action('stop', 'Stop Scan', 'stop', "Stop the scan", menu=self.action_menu)
-        self.add_action('move_at', 'Move at doubleClicked', 'move_contour',
+        self.add_action('stop', 'Stop Scan', 'stop_circle', "Stop the scan",
+                        menu=self.action_menu, icon_color=self.get_theme().red)
+        self.add_action('move_at', 'Move at doubleClicked', 'moving',
                         "Move to positions where you double clicked", checkable=True, menu=self.action_menu)
-        self.add_action('log', 'Show Log file', 'information2', menu=self.file_menu)
```
Ashwolaa, line 58: Missing an option to close the daq_scan!!!! Or give back the cross top right corner!

packages/pymodaq/src/pymodaq/extensions/scan/daq_scan.py
```diff
@@ -212,6 +211,39 @@ def setup_ui(self):
...
+        self.add_widget('dashboard_label', QtWidgets.QLabel('Dashboard:'),
+                        toolbar='dashboard')
+        self.add_action('show_dashboard', 'Show Dashboard', 'visibility',
+                        'Show/Hide the Dashboard window', checkable=True,
+                        icon_color=self.get_theme().green,
+                        icon_checked='visibility_off',
+                        icon_checked_color=self.get_theme().red,
+                        toolbar='dashboard')
+        self.preset_manager = PresetManager(self.dashboard, toolbar=self.get_toolbar('dashboard'))
+
```
Ashwolaa: Interesting choice, the idea is to directly reaccess the init phase inside the DAQ_scan? Should this be completed by a Configurator in that case? And if so, should the combinatino Preset/Configurator be merged through a single object?

packages/pymodaq/src/pymodaq/extensions/scan/daq_scan_ui.py
```diff
@@ -46,14 +46,14 @@ def __init__(self, parent):
...
-        self.add_action('start', 'Start Scan', 'run2', "Start the scan", menu=self.action_menu)
+        self.add_action('ini_positions', 'Init Positions', 'arrows_input', menu=self.action_menu)
+        self.add_action('start', 'Start Scan', 'motion_play', "Start the scan",
+                        menu=self.action_menu, icon_color=self.get_theme().green)
         self.add_action('start_batch', 'Start ScanBatches', 'run_all', "Start the batch of scans", menu=self.action_menu)
-        self.add_action('stop', 'Stop Scan', 'stop', "Stop the scan", menu=self.action_menu)
-        self.add_action('move_at', 'Move at doubleClicked', 'move_contour',
+        self.add_action('stop', 'Stop Scan', 'stop_circle', "Stop the scan",
+                        menu=self.action_menu, icon_color=self.get_theme().red)
+        self.add_action('move_at', 'Move at doubleClicked', 'moving',
```
Ashwolaa, line 56: No idea on how using this button

packages/pymodaq/src/pymodaq/extensions/scan/daq_scan_ui.py
```diff
@@ -46,14 +46,14 @@ def __init__(self, parent):
         self.connect_things()
 
     def setup_actions(self):
-        self.add_action('quit', 'Quit the module', 'close2', menu=self.file_menu)
-        self.add_action('ini_positions', 'Init Positions', '', menu=self.action_menu)
-        self.add_action('start', 'Start Scan', 'run2', "Start the scan", menu=self.action_menu)
+        self.add_action('ini_positions', 'Init Positions', 'arrows_input', menu=self.action_menu)
```
Ashwolaa, line 49: Not moving when pressing it

packages/pymodaq/src/pymodaq/extensions/scan/daq_scan.py
seb5g: My bad I forgot something. I did all the new development thinking at starting the application from the daq_scan directly. I suggest you try to start the daq_scan directly . From there you'll have all the sharedUI stuff activtated (and the close button). I will add stuff to have this also when you start the extension from the dashboard!

And concerning the preset choice within the extension, I agree that the configurator should be shipped together as well. But I still need to make those widgets and actions in sync with the main preset manager. I'll work on that and on your other suggestion and let you know when ready!

packages/pymodaq_gui/src/pymodaq_gui/managers/manager_base.py
seb5g: indeed it was supposed to be so before... I did it again for presets and configurators

packages/pymodaq/src/pymodaq/extensions/scan/daq_scan_ui.py
seb5g: its back with a proper SharedUI dealing with it

packages/pymodaq/src/pymodaq/extensions/scan/daq_scan.py
seb5g: I added also the configurator with everything correctly synchronised

packages/pymodaq/src/pymodaq/extensions/scan/daq_scan_ui.py
seb5g: this is after a scan you click on it to activate then by double clicking on a live plot, the actuators will go at those positions

packages/pymodaq/src/pymodaq/extensions/scan/daq_scan_ui.py
seb5g: it is now if the scanner is correct wrt the selected actuators

packages/pymodaq_gui/src/pymodaq_gui/managers/manager_base.py
```diff
@@ -411,7 +445,17 @@ def execute_entry_base(self, entry_path: Path = None, **kwargs):
...
+
+    @property
+    def entry_applied(self) -> bool:
+        return self._entry_applied
+
+    @entry_applied.setter
+    def entry_applied(self, applied: bool):
+        self._entry_applied = applied
+        if applied:
+            self.applied_entry.emit(self.entry_filename.stem)
```
Ashwolaa: When executing the load action for a new preet and pressing cancel, the configurator is still applied. Should not be the default

seb5g:
@Ashwolaa I'm good with this PR and I adressed your issues!

packages/pymodaq/src/pymodaq/dashboard.py
```diff
@@ -21,8 +21,10 @@
     QMessageBox,
 )
 import numpy as np
+from setuptools.command.bdist_egg import scan_module
```
Ashwolaa: remove?

Ashwolaa:
Two small things
- When applying a preset while the daq_scan is open, the close button disappears
- the configurator gets applied even when we press cancel (after trying to execute a preset), this should only occur after loading the preset (and this option should be checkable)

Otherwise looks good on usage! Nice!

packages/pymodaq/src/pymodaq/dashboard.py
seb5g: yes sure

seb5g:
> Two small things
> 
>     * When applying a preset while the daq_scan is open, the close button disappears
Could not reproduce this, what close button? Where? Do you mean the Dashboard disapears? You can still make it appear again using the display button but I'll make sure it stays as it was before  applying the preset
> 
>     * the configurator gets applied even when we press cancel (after trying to execute a preset), this should only occur after loading the preset (and this option should be checkable)
this is fixed. 
However, I planned this configurator 'default' entry to be run all the time after the preset is executed. The default configurator entry is therefore always created for all presets but could just be empty. It is to be used to set the starting point of your instruments. So I would prefer leaving it as is.
 
> 
> 
> Otherwise looks good on usage! Nice!
thx

Ashwolaa:
https://github.com/user-attachments/assets/ac8e483c-a043-4381-b1a2-7923fa91c3d0

seb5g:
that's the full toolbar that's missing in your case but it's not the case on my environment?? I'm using pyside6 and python 3.12.12 with those packages
pyleco==0.6.0.post1
-e git+https://github.com/PyMoDAQ/PyMoDAQ.git@c75df7326a45461477a8c5814b297aee439e441e#egg=pymodaq&subdirectory=packages\pymodaq
-e git+https://github.com/PyMoDAQ/PyMoDAQ.git@c75df7326a45461477a8c5814b297aee439e441e#egg=pymodaq_data&subdirectory=packages\pymodaq_data
-e git+https://github.com/PyMoDAQ/PyMoDAQ.git@c75df7326a45461477a8c5814b297aee439e441e#egg=pymodaq_gui&subdirectory=packages\pymodaq_gui
pymodaq_plugin_manager==1.3.8
pymodaq_plugins_mock==5.0.10
pymodaq_plugins_mockexamples==5.0.8
-e git+https://github.com/IumiShaping/pymodaq_plugins_optical_2D_shaping.git@5b069eb7fe4d7ea99567190edbc1ab53573ed59d#egg=pymodaq_plugins_optical_2D_shaping
-e git+https://github.com/PyMoDAQ/PyMoDAQ.git@c75df7326a45461477a8c5814b297aee439e441e#egg=pymodaq_utils&subdirectory=packages\pymodaq_utils
pyparsing==3.3.1
pyqtgraph==0.14.0
PySide6==6.10.1
PySide6_Addons==6.10.1
PySide6_Essentials==6.10.1
qt-material-icons==0.4.1
qt-themes==0.4.0
qtconsole==5.7.0
QtPy==2.4.3

Ashwolaa:
Python 3.13.10
It seems to happens both with Pyside6 and Pyqt6

PR #883, opened 2026-01-20
Refactor QApplication exec

### LINKED ISSUE #875: Calls to exec_ can be removed
It was useful with pyqt4 and PySide2, but pyqt5, pyqt6 and pyside6 supports `exec` (pyqt6 also remove `exec_`). Also they exit by themselves so any `sys.exit(app.exec_())` can be replaced by `app.exec()`, rendering the qt_utils.exec function obsolete

malik-irain:
Fixes #875 

* Replace all calls to exec_ with exec as exec_ is deprecated
* Wrap all of them with sys.exit to return the correct exit code to the OS
* Removed qt_utils.exec function as not needed anymore
* Fixes a problematic usage of `qt_utils.exec` in `pymodaq_gui.utils.widgets.push` where `PushButtonIcon` and `ActionMenu` implementation of `contextMenuEvent` where calling exec on the app instead of the menu

### DIFF
--- a/packages/pymodaq_gui/src/pymodaq_gui/qt_utils.py
+++ b/packages/pymodaq_gui/src/pymodaq_gui/qt_utils.py
@@ -82,6 +82,3 @@ def mkQApp(name: str):
     return app
 
 
-def exec():
-    app = mkQApp('a name')
-    return app.exec() if hasattr(app, 'exec') else app.exec()
--- a/packages/pymodaq_gui/src/pymodaq_gui/qt_utils.py
+++ b/packages/pymodaq_gui/src/pymodaq_gui/qt_utils.py
@@ -84,4 +84,4 @@ def mkQApp(name: str):
 
 def exec():
     app = mkQApp('a name')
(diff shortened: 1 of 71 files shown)


packages/pymodaq_gui/src/pymodaq_gui/qt_utils.py
```diff
@@ -84,4 +84,4 @@ def mkQApp(name: str):
 
 def exec():
     app = mkQApp('a name')
-    return app.exec() if hasattr(app, 'exec') else app.exec_()
+    return app.exec() if hasattr(app, 'exec') else app.exec()
```
seb5g: is it usefull or should we drop it?

seb5g (changes requested):

packages/pymodaq_gui/src/pymodaq_gui/qt_utils.py
malik-irain: I forgot to remove it!

seb5g (approved):

PR #890, opened 2026-01-27
Creation of a GlobalConfig to manipulate all configs

malik-irain:
With the goal to unify config under one object a GlobalConfig class where objects will live and eventually all be managed:
- Constructors of BaseConfig derivates should show a DeprecationWarning if used, reminding to use GlobalConfig registration mechanism
- Config objects are registered through a decorator (GlobalConfig.register) to put at the class level
- Config accessed through the GlobalConfig object keep the same keys but have a "prefix" corresponding to their config_name. So "pymodaq_utils.config.Config"
- Configs are now stored under .pymodaq/config (may be changed to configs through this PR)
- pymodaq_utils, pymodaq_data and pymodaq_gui now use the GlobalConfig object

There's still some work to do:
- [x] separate config for data and gui
- [x] register pymodaq (+ extensions) config(s)
- [x] use GlobalConfig in PyMoDAQ
- [x] check tree_toml works correctly especially for saving stuff
- [ ] write some tests


malik-irain:
A test in pymodaq_utils write a boolean in utils config style -> theme

PR #906, opened 2026-02-10
Feature/DataSaving_SingleWriteMultipleRead

Ashwolaa:
Currently, the way live plotter is managed during a DAQ_Scan is by making a new temporary file dedicated to the plotting. This is a good fallback solution but the better way would be to always have access to the entirety of the scan while it is being taken.

Among the last updates from the h5py library, one is the possibility to have the Single Write Multiple Read (SWMR) mode activated.
This mode allows to have multiple readers, from different threads or processes, able to directly acces an h5 file concurrently written in a dedicated thread/process.

In this PR, I propose to start benefiting from this new SWMR feature. The way it is currently implemented
**Note that it is only available with h5py not pytables!**

Among the next thing that we can look if this first feature works as intented is the possibility to develop extensions to do computations and plot it directly throughout the scan (for example transient, histograms, normalization, ...) with an access to all the instruments' data!

I am adding a roadmap which can be expanded according to additional suggestions.

- [x] SWMR implemented with h5py
- [x] Creating/Browsing files from daq_scan (now allows to select an existing file to add scan entry to it)
- [x] Adding a pause scan button 
- [x] Adding a QLED to indicate if file is accessed for daq_scan
- [x] Fix Timeout error (the actuators were disconnected at the end of the scan and entered into set_ini_positions without notification)
- [x] Add a settings to decide if actuators should go back to init when scan is stopped or finished
- [x] Switching easily from different hdf5_backend
- [x] Update h5browser to be compatible with SWMR mode

@rgeneaux @seb5g @malik-irain


seb5g:
First thought on the subject: 
* that goes back to how it was implemented before but indeed maybe the multiple read mode may solve the issue we had at the time. 
* does it mean you will always save the live acquisition?
* Often these are produced from ROI and are not meant to be saved
* if not saved, it means one would have to read/compute/display the data at each step while at the moment its read/display

malik-irain:
You put some tests about the pymodaq_data config object structure in pymodaq_utils and they fail, I suggest you move them in pymodaq_data

Ashwolaa:
> First thought on the subject:
> 
> * that goes back to how it was implemented before but indeed maybe the multiple read mode may solve the issue we had at the time.

Yes that is the main objective. This would make the data access much more fluid and easy in theory.
In practice, the second objective is to be able to live plot data with higher dimensions or with more complex schemes. For example, looking at a transient. 

> * does it mean you will always save the live acquisition?

What do you mean by this? The saving/writing would occur through the flusing mechanism. This is what make the data accessible to the readers. One can choose when writing occurs (every N steps).

> * Often these are produced from ROI and are not meant to be saved

That is a good point. Need to think on how dealing with this. One could still use the current implementation for it which is still the base one when not using h5py as backend.

> * if not saved, it means one would have to read/compute/display the data at each step while at the moment its read/display

Yep indeed. If we keep saving the ROI with h5py backend, one can then still use it directly too without the added computation cost.

Ashwolaa:
> You put some tests about the pymodaq_data config object structure in pymodaq_utils and they fail, I suggest you move them in pymodaq_data

Thanks, yes I finally understood it. I will create a config_test.py in pymodaq_data

Ashwolaa:
Ok I believe this is ready for a first round of review. 
Overall, the most impactful change for a random user is:
-  the access to the pause button during a scan. 
- can select an existing file to add additional scans to it (instead of always making a new file)

The backend selection follows Qt/logger and is now a list that updates according the available backend.

In theory, everything is now in placed for the SWMR approach. For this, select h5py backend and enable swmr. 
One can then either use a script / open a jupyter notebook / use a h5browser from a different process/thread and access the scan.


I think we can work a bit on the layout too. But it might be better to separate this into different PRs? What do you think?

seb5g:
I'll have a look to it! And yes let's do another PR for the layout. 

In the end, do you save all the live data (even from ROI?)?

Ashwolaa:
I have not changed what is being saved yet. So for now, this PR is really about making the hdf5 backend compatible with h5py (there were a few problems) and with the swmr mode.

packages/pymodaq_data/src/pymodaq_data/h5modules/data_saving.py
```diff
@@ -1003,6 +1018,19 @@ def add_data(self, where: Union[Node, str], data: DataToExport, indexes: Iterabl
                 self._data_saver.add_data(dwa_group, dwa, indexes=indexes,
                                           distribution=distribution)
 
+        # Enable SWMR after first data point created all structure
+        if (self._h5saver._swmr_mode and not self._swmr_activated
+                and self._h5saver.backend == 'h5py'):
```
seb5g: could you avoid calling the backend type here. The purpose of this is to have backends themselves handling their eventual difference

packages/pymodaq_data/src/pymodaq_data/h5modules/data_saving.py
```diff
@@ -959,6 +961,19 @@ def __init__(self, h5saver: H5SaverLowLevel, extended_shape: Tuple[int]):
         super().__init__(h5saver)
         self._data_saver = DataExtendedSaver(h5saver, extended_shape)
         self._nav_axis_saver = AxisSaverLoader(h5saver)
+        self._swmr_activated = False
+        self._flush_interval = 0
+        self._write_count = 0
+
+    def set_swmr_flush_interval(self, interval: int):
```
seb5g: why is this method here and not in the base class?

The DataToExportTimedSaver is not for periodic saving but a specific enlargeable saver for saving data as a function of the timestamp when data is added. Typically for the DaqLogger

seb5g: how is handled your flushing interval? I guess it should be in the add_data method of the base saver

packages/pymodaq_data/src/pymodaq_data/h5modules/saving.py
```diff
@@ -116,14 +118,27 @@ def init_file(self, file_name: Path, raw_group_name='RawData', new_file=False,
         if file_name is not None and isinstance(file_name, Path):
             self.h5_file_name = file_name.stem + ".h5"
             self.h5_file_path = file_name.parent
-            if not self.h5_file_path.joinpath(self.h5_file_name).is_file():
+            fullpath = self.h5_file_path.joinpath(self.h5_file_name)
+            if not fullpath.is_file():
                 new_file = True
+            elif swmr_mode and not new_file and self.backend == 'h5py':
```
seb5g: same remark about not checking in here what backend we use, it should be transparent to the user. It means there should be some method/attribute in the backend class allowing to know if swmr is available. Do you think it will prove to be too complicated?

packages/pymodaq_data/src/pymodaq_data/h5modules/saving.py
```diff
@@ -145,6 +160,26 @@ def save_file(self, filename=None):
             if str(file_path) != '':
                 super().save_file_as(filename)
 
+    def finalize_swmr(self, keep_open=False):
```
seb5g: this should probably also be in the backend and called in that object if necessary (but probably in some other too)

packages/pymodaq_data/src/pymodaq_data/resources/config_template.toml
```diff
@@ -1,6 +1,9 @@
 #this is the configuration file of the pymodaq_data package
 
 [data_saving]
+    swmr_enabled = false
```
seb5g: I guess it should be true by default? I don't see much reason not too (and therefore setting h5py as default)

seb5g: I almost never use h5py in real life but only in tests. Do you have experience on using it. Does it behave well and transparently (I know it should as this was the purpose of the backend stuff, but well I did this during the covid so it was a while back...)

seb5g (changes requested):
huge work! Could have a look at my comments ?

packages/pymodaq_data/src/pymodaq_data/h5modules/data_saving.py
Ashwolaa: Added a property is_swmr_capable in backends:py which checks if the backend matches the one with SWMR

packages/pymodaq_data/src/pymodaq_data/h5modules/saving.py
Ashwolaa: Same as above

packages/pymodaq_data/src/pymodaq_data/h5modules/saving.py
Ashwolaa: This has also be moved to the backend

packages/pymodaq_data/src/pymodaq_data/resources/config_template.toml
Ashwolaa: Changed to True

I mostly use h5py myself (not knowing about pytables before). I made a few tests which seemed to indicate the pytables was slightly faster than h5py. However, forums seem to indicate that it heavily depends on how you manage the cache in your data. I did not push much further but I did not see much of an impact testing the two in the daq_scan

Ashwolaa:
So I made a change with respect to the scan_done attribute. I find it clearer to have a dedicated boolean matrix which indicates what points have been taken. It was also a bit more cumbersome to use since it is not possible to modify the attrs when the swmr mode is active.
So in the last commits, I removed it and replaced it with a new dataset "ScanDone" which is just a matrix giving True/False if the data point has been taken in the scan.

Ashwolaa:
> So in the last commits, I removed it and replaced it with a new dataset "ScanDone" which is just a matrix giving True/False if the data point has been taken in the scan. 

Instead of a boolean, we could use a timestamp. This could be a useful feature to track stability issues during a scan.

seb5g:
I didn't remember I had this attrs back ten... The stuff with the timestamps can actually come in handy for a question that has been raised to me recently. In random mode how one can track in which order the scan points have been taken. That solves it.
But it also adds a 0D data array... do you think it make a big weight on top of the usual data?

Ashwolaa:
I think it should be fine.  I get roughly 4Mo for an array (100,100,100), 32 Mo for (200,200,200), ....
timestamps_matrix = np.random.randint(0, 2^32 - 1, 100^3, dtype=np.uint32)
np.save("timestamps_matrix.npy", timestamps_matrix)

We can just record the first time stamp when the scan starts and then get the delta. This gives us a window of 2^32/1000/3600 hours with ms precision (which seems way overkilled)

seb5g:
Hello,
I looked into it. I'm not sure of what I was doing with the options:

* load file
* open current file
* close current file

I feel there is some file status that is missing to know what's going on.
I found it on the bottom right bu then it says this:
image


red but say file is open??

Then I closed it but if I start a scan it just reopen it... I guess it's normal. Overall I'm lacking some guide on what's the impact of opening/closing. We should set the swmr mode as defualt in the config template as well as hdf5.


For the timestamps, they are not saved in a group making the arrays hanging. 
image

* it's not tidy
* axes are not grouped together.
* because of that clicking on the node doesn't plot it, in fact it triggers the error below:

image

I guess it's because you didn't use advanced saving mechanism to save it properly... To do it right I would first create a data group on the same level as detector called Timestamps or something and save the dwa or dte within using: 

```python
with DataToExportSaver(fname, metadata={'settings': plugin_config.to_xml_string()}) as h5saver:
    group = h5saver.add_data_group('/RawData', DataDim.Data2D, title='SLM Phases', group_name='SLM')
    h5saver.add_data(group, dte)
``` 
or 
```python
with DataSaverLoader(fname, new_file=new_file) as saver:
    group = saver.add_data_group(where, DataDim.Data2D, title=title, settings_as_xml=settings_str,
                                 group_name=group_name)
    saver.add_data(group, dwa)
```

Then be aware that the navigation axes are already saved in the nav_axis group. Because these axes are shared with all data nodes. I make it possible for the h5 browser to discover them in the hierarchy. So probably you should not save the data as simply as what I wrote above.

I found the culprits: 
```python
            # Pre-allocate the per-point acquisition time tracker (CARRAY: SWMR-compatible indexed writes)
            # Values are elapsed seconds since scan start; nan means the point has not been acquired yet.
            self._scan_time_array = self._h5saver.add_array(
                scan_node, 'ScanTime', data_saving.DataType.data,
                array_to_save=np.full(scan_shape, np.nan, dtype=np.float32),
                data_dimension=data_mod.DataDim['DataND'])
            # Tag all dimensions as navigation so h5browser renders with proper axes
            self._h5saver.set_attr(self._scan_time_array, 'nav_indexes',
                                   tuple(range(len(scan_shape))))
            self._h5saver.set_attr(self._scan_time_array, 'distribution', 'uniform')
            self._h5saver.set_attr(self._scan_time_array, 'units', 's')
            # Store scan axes alongside ScanDone so h5browser can label the dimensions
            axis_saver = data_saving.AxisSaverLoader(self._h5saver)
            for axis in nav_axes:
                axis_saver.add_axis(scan_node, axis)
```
 
as I said you should not do it "manually" like this. The best location to add stuff to be saved within a DAQScan is in here :
```python
self._module_and_data_saver: module_saving.ScanSaver = module_saving.ScanSaver(self)
```

This guy, will save data for each module it has registered. It will instantiate the nodes and then will transfer the h5file pointer to the modules for they are the one adding data into the nodes. I guess you should create a simple TimeModule and a TimeModuleSaver inhering from ModuleSaver. This is as close as the current way of properly saving things within a scan. Only the ScanSaver handle the navigation axes , and each module adds its data (0D, 1D, 2D...) intro the corresponding location in the data nodes. 

Hope it makes a bit of sense. If not, I'll do it later ;-)

And finally for the todo list, could you check that one can initiate the arrays with Nans? Aurore did some tests but not with the DAQScan!!!!

Are you happy I reviewed your code :-D ?

Ashwolaa:
Thanks for the feedback, yeah I went a bit wild without thinking too much on adding new elments to the h5. I was mostly happy to have my 2D matrix with the timestamps. 
I will give it a shot.

Ashwolaa:
Okay, so:
- I added in the config the possibility to choose how the arrays should be pre-filled. 
image

There is no difference in the browsing at all.
image
image


-  I tried to follow your approach by creating a new TimeModuleSaver (filled explicitely with np.nan but since we are relative in time, this could also be set to zero or follow the config file).

- The new structure looks like this. It is a bit overkilled imo but it stays consistent at least
image

- For the close/open/status, these were just some initial thoughts on interesting actions and visual indicators. They were essentially good tools for debugging (testing the different case scenarios). However, I assume we try to separate as much as possible the GUI part and the backend saving. I can remove them but they seem to behave as expected. I remove the auto closing of the file at the end of the scan which was a bit disturbing (and maybe not so smart also as a reader can lock it if not opened correctly).

Edit:
I changed the default label for the Viewer to index. I was a bit confused reading pixels when navigating on some actuators scan

packages/pymodaq/src/pymodaq/utils/h5modules/module_saving.py
```diff
@@ -459,6 +463,72 @@ def add_data(self, dte: DataToExport = None, indexes: Tuple[int] = None,
...
+        self._h5saver.set_attr(group, 'settings', ET.tostring(settings_xml))
+        return group
+
+    def initialize(self, extended_shape: Tuple[int]):
+        """Set up the extended saver; arrays are allocated on first add_time call."""
+        self._extended_shape = extended_shape
+        self._datatoexport_saver = DataToExportExtendedSaver(
+            self._h5saver, extended_shape, fill_value=np.nan)
+
+    def add_time(self, indexes: Tuple[int], elapsed_time: float):
```
seb5g: I'm wondering if it shoudl internally handle the timing also like having a method: start to cache the starttime then everytime add_time is called it would compute and add the elpased time. the startime method could also be added in the ScanSaver for commodity

packages/pymodaq_data/src/pymodaq_data/h5modules/backends.py
```diff
@@ -54,6 +54,8 @@
 if not (is_tables or is_h5py or is_h5pyd):
     logger.exception('No valid hdf5 backend has been installed, please install either pytables or h5py')
 
+SWMR_CAPABLE_BACKENDS = frozenset({'h5py'})
```
seb5g, line 57: what's a frozenset for?

packages/pymodaq_data/src/pymodaq_data/h5modules/backends.py
```diff
@@ -739,6 +746,26 @@ def reconcile_swmr_attrs(self):
                     actual_shape = node.node.shape
                     node.attrs['shape'] = actual_shape
 
+    def finalize_swmr(self, keep_open=False):
+        """End SWMR by closing the file, reopening in 'a' mode, and reconciling deferred attrs.
+
+        After SWMR mode, attrs['shape'] on EARRAY/VLARRAY nodes may be stale.
+        This method closes the file (ending SWMR), reopens it in append mode,
+        and updates all deferred attributes.
```
seb5g, line 772: what does it mean deferred attributes?

packages/pymodaq_data/src/pymodaq_data/h5modules/data_saving.py
```diff
@@ -678,9 +686,10 @@ class DataExtendedSaver(DataSaverLoader):
     """
     data_type = DataType['data']
 
-    def __init__(self, h5saver: H5SaverLowLevel, extended_shape: Tuple[int]):
+    def __init__(self, h5saver: H5SaverLowLevel, extended_shape: Tuple[int], fill_value=None):
         super().__init__(h5saver)
         self.extended_shape = extended_shape
+        self.fill_value = fill_value
```
seb5g, line 692: if its None, how doe it impact the array initialization? should you not then fallback to the config defined one?

packages/pymodaq_data/src/pymodaq_data/h5modules/data_saving.py
seb5g, line 1041: much better

packages/pymodaq_data/src/pymodaq_data/h5modules/data_saving.py
```diff
@@ -1020,16 +1033,12 @@ def add_data(self, where: Union[Node, str], data: DataToExport, indexes: Iterabl
...
             self._h5saver.flush()
             self._h5saver.enable_swmr()
             self._swmr_activated = True
 
-        # Periodic flush for SWMR readers
-        if self._swmr_activated and self._flush_interval > 0:
-            self._write_count += 1
-            if self._write_count % self._flush_interval == 0:
-                self._h5saver.flush()
+        self._h5saver.tick_flush()
```
seb5g, line 1041: what does it mean to flush data for readers? what's the impact ?

packages/pymodaq_gui/src/pymodaq_gui/h5modules/saving.py
```diff
@@ -143,6 +146,10 @@ class H5SaverBase(H5SaverLowLevel, ParameterManager):
         {'title': 'Saving dynamic', 'name': 'dynamic', 'type': 'list',
          'limits': config('data', 'data_saving', 'data_type', 'dynamic'),
          'value': config('data', 'data_saving', 'data_type', 'dynamic')[0]},
+        {'title': 'Fill value:', 'name': 'fill_value', 'type': 'list',
```
seb5g: you could use a dict as limits with:

```python
limits = {'0': 0, 'nan': np.nan}
```

how doe it articulate with the "dynamic" setting?

packages/pymodaq_gui/src/pymodaq_gui/h5modules/saving.py
```diff
@@ -530,13 +539,16 @@ def value_changed(self, param):
...
-            if not is_h5py and self.settings['swmr_options', 'enable_swmr']:
+            swmr_capable = new_backend in SWMR_CAPABLE_BACKENDS
+            self.settings.child('swmr_options').setOpts(visible=swmr_capable)
+            if not swmr_capable and self.settings['swmr_options', 'enable_swmr']:
                 self.settings.child('swmr_options', 'enable_swmr').setValue(False)
                 self.update_status('SWMR is only supported with h5py backend, disabling.')
             self.set_backend(new_backend)
 
+        elif param.name() == 'fill_value':
+            self.fill_value = np.nan if param.value() == 'nan' else float(param.value())
```
seb5g: using the dict above this would just update the fill_value

maybe having both a variable and the setting is too much:
maybe a fill value property would be better

```py
@property
def fill_value(self) -> Union[float, np.nan]:
     return self.settings[..., 'fill_value']

@fill_value.setter
def fill_value(self, value: Union[float, np.nan]):
     self.settings.child(..., 'fill_value').setValue(value)
```

seb5g (changes requested):

seb5g:
> Okay, so:
> 
>     * I added in the config the possibility to choose how the arrays should be pre-filled.
> 
perfect, it maybe should go just together with the dynamic setting?
> 
>     * I tried to follow your approach by creating a new TimeModuleSaver (filled explicitely with np.nan but since we are relative in time, this could also be set to zero or follow the config file).
> 
>     * The new structure looks like this. It is a bit overkilled imo but it stays consistent at least
overkilled but neat!!
> 
> 
>     * For the close/open/status, these were just some initial thoughts on interesting actions and visual indicators. They were essentially good tools for debugging (testing the different case scenarios). However, I assume we try to separate as much as possible the GUI part and the backend saving. I can remove them but they seem to behave as expected. I remove the auto closing of the file at the end of the scan which was a bit disturbing (and maybe not so smart also as a reader can lock it if not opened correctly).

probably in the GUI we just need the possiblilty to open an existing file and append scans there (maybe trigger an update of the dataset metadata filling that disapeared in some past versions (i just realized it). The actual opening or closing would be done under the hood automatically. Maybe keep some "hidden" action for debugging purpose
 
> Edit: I changed the default label for the Viewer to index. I was a bit confused reading pixels when navigating on some actuators scan

ok

packages/pymodaq/src/pymodaq/utils/h5modules/module_saving.py
Ashwolaa: I like it, I will do that

packages/pymodaq_data/src/pymodaq_data/h5modules/data_saving.py
Ashwolaa: If None, it is the self.fill_value of the h5saver which is read from the config but that means the fill_value could only be implemented in the h5saver instead.

packages/pymodaq_data/src/pymodaq_data/h5modules/data_saving.py
Ashwolaa: This is detailled here
https://docs.h5py.org/en/latest/swmr.html

On the writer side, when a dateset is flushed,  the data update is accessible to the readers. The reader dataset needs to call their refresh method to see the change.

packages/pymodaq_data/src/pymodaq_data/h5modules/backends.py
Ashwolaa: not much, it is a immutable set. It is a bit overkilled here as only h5py is currently adapted. But if pytables do it, we could add it.

packages/pymodaq_data/src/pymodaq_data/h5modules/backends.py
Ashwolaa: All the attributes that should/need to be changed. It is not possible to touch them as long as the file is open in swmr mode

packages/pymodaq/src/pymodaq/utils/h5modules/module_saving.py
Ashwolaa: Done, the time information is now limited to the TimeModuleSaver

packages/pymodaq_gui/src/pymodaq_gui/h5modules/saving.py
Ashwolaa: done

packages/pymodaq_gui/src/pymodaq_gui/h5modules/saving.py
Ashwolaa: nan does not exist with int type but would give 0, kind of a lie ^^

seb5g: I don't understand?!?

seb5g (approved):
This definitely is a big improvement. I especially like the pause and the open file on a live scan. 

I think we could operate better the scan flow: start, run, pause, stop by using a state machine. This would avoid the use of both while loop. I'm goign to work with Bernhard on those state machine and may implement them in there. What is actually interesting is that you can easily pause the scan to perform something else (optimisation for instance), then resume... (it should be doable right now but the while loops and processevents  are not very nice...)

Ashwolaa:
`This definitely is a big improvement. I especially like the pause and the open file on a live scan.`

Glad that you like it. I have something big that will be out soon which profits from it!
Overall, from my test case, the swmr has been fairly resilient overall. Accessing/Reading the file while the scan is ongoing or paused and resuming without problems.


`I think we could operate better the scan flow: start, run, pause, stop by using a state machine. This would avoid the use of both while loop. I'm goign to work with Bernhard on those state machine and may implement them in there. What is actually interesting is that you can easily pause the scan to perform something else (optimisation for instance), then resume... (it should be doable right now but the while loops and processevents are not very nice...)
`
I agree, this looks like a much proper way to operate it and could be generalized to multiple kind of measurements.

seb5g:
I just pushed a commit to correctly use the fill_value parameter: when using a type list with a dict as limits both .value() and setValue() refer to the values of the dict while the combobox displays the keys/strings. I will merge it as it was the only thing I found not working properly

PR #908, opened 2026-02-11
global pyproject

Ashwolaa:
Hey everyone,

so I have been struggling a lot using pytest and vscode.  The main reason comes from the pytest discovery which had a lot of errors and I was stuck passing by the terminal to test each package separately.
The errors mostly originated from tests sharing the same name between the different packages in pymodaq.

Trying to fix this, I discovered two things:
-- The "--import-mode=importlib" argument allows pytest to make the difference between files with same name.
-- We can give a global config to the different subpackages with this keyword attached to it

I propose to take this occasion to make a pyproject.toml that gathers the formatting/coverage/settings that we would like to be applied in the suite.

Here is a proposition of such file

### DIFF
--- a/packages/pymodaq/pyproject.toml
+++ b/packages/pymodaq/pyproject.toml
@@ -85,11 +85,3 @@ include = [
     "/src",
 ]
 
-[tool.coverage.run]
-omit = [
-    "*/examples/*",
-    "*/resources/*"
-]
-
-[tool.ruff]
-line-length = 120
--- a/packages/pymodaq_gui/pyproject.toml
+++ b/packages/pymodaq_gui/pyproject.toml
@@ -79,7 +79,3 @@ include = [
     "/src",
 ]
 
-[tool.coverage.run]
-omit = [
-    "*/examples/*",
-]
--- a/packages/pymodaq_utils/pyproject.toml
+++ b/packages/pymodaq_utils/pyproject.toml
@@ -74,7 +74,3 @@ include = [
     "/src",
 ]
 
-[tool.coverage.run]
-omit = [
-    "*/resources/*"
-]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,52 @@
+# Root configuration for development tooling
+# Package-specific build configs remain in packages/*/pyproject.toml
+
+[tool.pytest.ini_options]
+addopts = "--import-mode=importlib"
+testpaths = ["packages"]
+filterwarnings = [
+    "ignore::DeprecationWarning",
+    "ignore::PendingDeprecationWarning",
+]
+
+[tool.coverage.run]
+source = ["packages"]
+omit = [
+    "*/tests/*",
+    "*/examples/*",
+    "*/resources/*",
+]
+
+[tool.coverage.report]
+exclude_lines = [
+    "pragma: no cover",
+    "if TYPE_CHECKING:",
+    "if __name__ == .__main__.:",
+    "raise NotImplementedError",
+]
+
+[tool.ruff]
+line-length = 120
+target-version = "py39"
+
+[tool.ruff.lint]
+select = [
+    "E",      # pycodestyle errors
+    "W",      # pycodestyle warnings
+    "F",      # pyflakes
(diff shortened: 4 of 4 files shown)


seb5g:
how does it behave this main pyproject file? I thought pyproject could only be at the root of each package?
What app is using this pyproject and by the way how do you use ruff?

I'll let Malik handle this PR as he worked most on these stuff!

Ashwolaa:
So with this pyproject toml, my whole IDE reacts to it and indicate me all the things that Ruff consider as bad habbits (I see yellow lines everywhere now). I consider it as mostly useful for developers as it encourages to follow a specific layout and make sure that all packages follow it. 

With Ruff, you can install it with `pip install ruff`. 

It is pretty powerful, you have some straightforward commands. For example, to see how you should order all your imports you can use:
`ruff check --select I .`
and you can apply it with 
`ruff check --select I --fix .`

Similarly, all the unused import are refered to error F401, which can be removed with:
`ruff check --select F401 .`
In the context of pymodaq, if the import is required (for example `packages\pymodaq\src\pymodaq\__init__.py` it is suggested to use:
`import pymodaq_utils as pymodaq_utils` for Ruff to not consider it as unecessary.

I have not yet explored it too much but it looked fairly efficient to apply rapidly some defined practices over a whole library.
I went for a very fast pyproject.toml but I believe it is worth discussing as there are so many options.

seb5g:
> So with this pyproject toml, my whole IDE reacts to it and indicate me all the things that Ruff consider as bad habbits (I see yellow lines everywhere now). I consider it as mostly useful for developers as it encourages to follow a specific layout and make sure that all packages follow it.
> 
> With Ruff, you can install it with `pip install ruff`.
> 
> It is pretty powerful, you have some straightforward commands. For example, to see how you should order all your imports you can use: `ruff check --select I .` and you can apply it with `ruff check --select I --fix .`
> 
> Similarly, all the unused import are refered to error F401, which can be removed with: `ruff check --select F401 .` In the context of pymodaq, if the import is required (for example `packages\pymodaq\src\pymodaq\__init__.py` it is suggested to use: `import pymodaq_utils as pymodaq_utils` for Ruff to not consider it as unecessary.

I believe Benedikt showed me also a specific comment you can add to let it know that the import is needed something like: # noqa ...


> I have not yet explored it too much but it looked fairly efficient to apply rapidly some defined practices over a whole library. I went for a very fast pyproject.toml but I believe it is worth discussing as there are so many options.

Do you think you could write some guide lines (for code standard)(maybe for other code/project implementation) as you seem to know most about this? There are however a few thing I like and would like to keep: 

* the use of underscore in variables
* not automatically separating all arguments on a line (as sometimes it makes things less clear)
*  not automatically separating all keys of a dict on lines (especially for Parameters instanciation) as it makes things lengthy and also not that clear
* ... (we'll see for the rest)

thx

seb5g:
# noqa: F401 

this is for the linter/flake8 to ignore unused imports

malik-irain:
> how does it behave this main pyproject file? I thought pyproject could only be at the root of each package? What app is using this pyproject and by the way how do you use ruff?
> 
> I'll let Malik handle this PR as he worked most on these stuff!

It also allows to run pytest on the root of the repo and it will correctly triggers all tests. Also in pycharm I could start a specific test (it would launch all the package or file tests at once) and now it's possible

It works nicely even if I don't have ruff installed, as it only serves as a config file for it, so no issues on this side

One drawback is that doing `pip install .` at repository root now install `UNKNOWN-0.0.0`. It would be nice to find a way to forbid that!

Including the guidelines in the config would be nice, so that ruff usage don't mess with how we want to keep stuff

seb5g:
I would definitely like to  know more on how to use ruff stuff, @Ashwolaa could you do a webinar on the subject as we can't come eat pizza in your leaving room easily :-D

Ashwolaa:
haha I am no expert but I will try to find some time to look more into it

bernhardlang:
If that webinar hasn't happened yet, I'd be interested to join the club (pizza perhaps then in Saclay :-)

PR #910, opened 2026-02-11
Function to reduce units only if there is dimensionality

### LINKED ISSUE #909: move_rel don't work with angles
In the mock actuator, using a move rel on the Theta axis (with angle in °) leads to the following issue where it never go the wanted value:

Image

The issues comes from to_reduce_units that convert the ° value to radians and the unit to dimensionless
 
https://github.com/PyMoDAQ/PyMoDAQ/blob/3ed7b8dc27a2045e2812e4f6b0bf33762170c94d/packages/pymodaq_data/src/pymodaq_data/data.py#L963-L983

I suppose it also happens with other dimensionless units

malik-irain:
Fixes #909

It works, but I'm not sure if it should be limited to "degree/deg/°" units or kept as is?

### DIFF
--- a/packages/pymodaq_data/src/pymodaq_data/data.py
+++ b/packages/pymodaq_data/src/pymodaq_data/data.py
@@ -48,7 +48,9 @@ logger = set_logger(get_module_name(__file__))
 
 def dimensionless_aware_reduce_units(q: Type[Q_]) -> Type[Q_]:
     """
-    Take a quantity q and converts it to its reduced units, except if its dimensionless
+    Take a quantity q and converts it to its reduced units.
+    For dimensionless units it compact them
+
     Parameters
     ----------
     q: The quantity to be reduced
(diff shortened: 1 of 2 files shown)


seb5g (approved):
good catch!!and fix!

seb5g:
for your question, what could be without dimensionality but with a 'unit string': rpm, ..., steps... so I think it may be good to let your fix for all!

seb5g:
could you add a test for operation between quantity to check in the future this behaviour?

seb5g (approved):
add tests please!

seb5g:
@malik-irain could you wrap this PR up?

malik-irain:
I changed the function to returns `compact_units` for dimensionless so that `"° * °"` -> `"° ** 2"`. The only "issue" is multiplying compatible dimensionless units such as `"° * rad"` still returns `"° * rad"` (but that's OK)

(I'm fixing the issue with the tests)

PR #916, opened 2026-02-14
Feature/viewer0 d upgrade

Ashwolaa:
Currently the viewer0D resets the full dataset when a new entry is added. In this PR, we add a new action to decide if the x-axis should be reset or not. 
 -- If the icon is checked (default behavior), the data are being reset. 
 -- If the icon is unchecked, the new data gets filled with nan to be aligned with the other curves
The settings tree used here is taken from the last PR on pymodaq_plugins.
https://github.com/PyMoDAQ/pymodaq_plugins_mock/pull/38

A demo will make it clearer:
https://github.com/user-attachments/assets/b545f853-716a-4811-b581-408977d93215

N.B.: There is an ill behavior in the start up showing the wrong icon which should be fixed with #915


seb5g:
nice indeed! i accepted the Mock PR

PR #920, opened 2026-02-15
Feature/compact dock manager

Ashwolaa:
With the new layout, it becomes interesting to propose global actions mechanisms to all actuators/viewers and to factorize some of the common methods.

In this PR:
- Common methods, such as name, init, settings are factorized into global methods in ui_utils.py. This can later be expanded.
-  the simple dock used to store the toolbars is expanded into a new object that properly align the toolbars + add some global actions to it.

Here is how it looks like in practice:

https://github.com/user-attachments/assets/7e5b67a8-3a23-452f-933a-9533838c39e9

For now, the global actions chosen are:
- locking mechanism to disable actions
- show/hide widget
- grabbing / snapping

One can maybe dream of additional global actions such as removing or adding modules dynamically in the near future, who knows?

This PR makes use of #919


packages/pymodaq/src/pymodaq/control_modules/ui_utils.py
```diff
@@ -28,14 +31,91 @@ class ControlModuleUI(CustomApp):
...
+        """Add the module name label to the toolbar
+
+        Parameters
+        ----------
+        toolbar: QToolBar
+            The toolbar to add the widget to. If None, uses default toolbar.
+        use_label_class: bool
+            If True, uses LabelWithFont, otherwise uses QLabel with custom font
+        """
+        if use_label_class:
```
seb5g: why using both way?

seb5g:
that's quite neat!! I patched a few things I noticed but unrelated to your pr in fact...

The only question is about the LabelWithFont: why did you make on other way, it seems to be doing the same!

seb5g:
And I've been bothered by the unaligned names/buttons for a while and it magically disapeared :-D

Ashwolaa:
> that's quite neat!! I patched a few things I noticed but unrelated to your pr in fact...
> 
> The only question is about the LabelWithFont: why did you make on other way, it seems to be doing the same!

To be fair, these common methods are a legacy of a PR I was planning to homogenize/factorize most of the methods between the DAQs which went a bit out of hands. I need to be a bit more sequential. I simplified it to always use LabelWithFont.

> And I've been bothered by the unaligned names/buttons for a while and it magically disapeared :-D

Haha I feel you, it triggered me enough the first time I faced it to do something about it.

seb5g (approved):

PR #923, opened 2026-02-17
Scripting capabilities in PyMoDAQ

malik-irain:
First draft of what scripting in PyMoDAQ could look like. It automatically starts a leco coordinator at startup and when a preset is loaded, devices are all connected to the coordinator automatically. Afterwards one can use `Actuator` and `Device` from `pymodaq.scripting`. 

Most of the methods returns a result in the form of a Future. Basically, it is a result that will be filled later and its result can be accessed using the `result()` method. It allows to be async or sync at the same time, as calling `result()` will wait for the result to be available, but not calling it allow you to do different operations in parallel and wait (or not) for completion later by calling `result()`.

For actuators:
-  `get_actuator_value`, returns a Future containing the current value
- `move_*` methods, return a Future containing the *final* position. *Moving to the current position makes a Future that's never filled with the result, as nothing is done on PyMoDAQ side!*
- `stop_move`, returns a Future containing the final position after stopping

For detectors:
- `snap`, returns None if a grab is occurring, otherwise a Future containing the acquisition
-  `stop_grab` returns nothing
- `grab`, has two modes:
    - `keep=False`, by default, it doesn't give you access to the grabbed data in the script
    - `keep=True`, then the `grab` method will return a list (**not a Future!**) that will be filled until `stop_grab()` is called or `grab()` is called again.

For general devices:
- `get_settings`, returns a Future containing the xml tree representing the device settings. It can be manipulated and send back
- `set_settings`, takes the **modified** settings tree and update the settings

Modifying the settings tree is a bit tedious and is a temporary solution until, the Configurator gets LECOfied. Maybe not the best example but to modify a device from `Slave` to `Master`:

```python
...
tree = device.get_settings().result()
tree.find(".//controller_status").text = "str('Master')"
device.set_settings(tree)
...
```

For example, using the default preset, one can do that to achieve a simple scan:

**`simple_scan_script.py`** 
```python
from pymodaq_data import Q_
from pymodaq.scripting import Detector, Actuator

theta = Actuator('Theta')
det0d = Detector('Det0D')

acquisitions = []

base = theta.get_actuator_value().result()
target = base + Q_('90°') # Or a DataActuator(data=90, units='°')
step = Q_('10°')

count = (target - base)/step
count = int(count.quantities[0].m[0]) + 1

value = base 

for idx in range(1, count + 1):
    acquisitions.append((value, det0d.snap().result()))
    value = theta.move_abs(base + idx * step).result()


print(acquisitions)
```

To do:

- [x] Write documentation about usage
- [x] Write documentation about "LECOfication" of an element
- [ ] Write tests, also about the type of exchanged data
- [x] Use serializall for all data types
- [x] Refactor PyMoDAQ part with a "LECOMixin" or something mixin class for component that should be scriptable
- [x] Refactor and document scripting part


seb5g:
reading on the proposed lecoisation of the dashboard, do you think we should produce a mixin class to inherit from (and subclass its methods and commands) to provide a clear way to make it available within the leco protocol? we could use it for the daq_move, viewer, and extensions!

malik-irain:
Yes there should definitely be a mixin, it's actually not so hard to make something available and controllable through LECO and there's a quite a few redundant stuff. For now, i'm building a basic set of commands that work but I'll soon refactor into a better architecture.


I just made some dashboard features available for scripting:

```python

from pymodaq.scripting import Detector, Actuator, Dashboard

# Once a dashboard is started:
dashboard = Dashboard()


print(dashboard.get_presets().result())
# >>> ['default']
# Could make it so that it returns a future that completes when it's finished loading
dashboard.apply_preset('default')

print(dashboard.get_configurations().result())
# >>> ['default', 'my_custom_config']

# Same remark as dashboard.apply_preset
dashboard.apply_configuration('my_custom_config')

print(dashboard.get_devices().result()) # returns the names
# >>> {'actuators': ['Theta', 'Temperature', 'Power', 'Xaxis'], 'detectors': ['Det2D', 'Det0D', 'Det1D']}

devices = dashboard.get_scripting_devices()
print(devices)
# >>>  {'actuators': {'Theta': <Actuator object>, 'Temperature': <Actuator object>, 'Power': <Actuator object>, 'Xaxis': <Actuator object>}, 'detectors': {'Det2D': <Detector object>, 'Det0D': <Detector object>, 'Det1D': <Detector object>}}


#Then each device can be used
devices['actuators']['Theta'].move_abs('90°')
....
```

seb5g:
Very nice!

malik-irain:
It's not complete, but I think it's getting pretty much usable now, as everything works as expected, even the issue with moves to an already reached position

(But github seems to not want to show the last commits)

seb5g:
do you want me to do a review yet or to wait for leco refactorisation?

malik-irain:
More of a usage review  if possible to see if didn't miss something obvious in terms of features

packages/pymodaq/src/pymodaq/control_modules/utils.py
```diff
@@ -505,26 +506,18 @@ def connect_tcp_ip(self, params_state=None, client_type: str = "GRABBER") -> Non
...
-        port = self.settings["main_settings", "leco", "port"]
-        if host == '':
-            # take the localhost as default
-            host = 'localhost'
-        if port == '':
-            # take the default port as 12300
-            port = 12300
-        return (host, port)    
+    def get_leco_host_port(self) -> tuple[str, int]:
+        host = (self.settings["main_settings", "leco", "host"] or 'localhost')
```
seb5g, line 467: that's a boolean isn't it?

packages/pymodaq/src/pymodaq/control_modules/utils.py
```diff
@@ -505,26 +506,18 @@ def connect_tcp_ip(self, params_state=None, client_type: str = "GRABBER") -> Non
...
+        port = (self.settings["main_settings", "leco", "port"] or 12300)
```
seb5g, line 468: same!

packages/pymodaq/src/pymodaq/control_modules/daq_viewer.py
```diff
@@ -114,13 +114,14 @@ def __init__(
         parent: Optional[QtWidgets.QWidget] = None,
         title: str = "Testing",
         daq_type=config("pymodaq", "viewer", "daq_type"),
+        preset : Optional[str] = None,
```
seb5g: what is preset for?

packages/pymodaq/src/pymodaq/control_modules/daq_move.py
```diff
@@ -101,9 +101,8 @@ class DAQ_Move(ParameterControlModule):
     listener_class = MoveActorListener
     ui: Optional[DAQ_Move_UI_Base]
 
-    def __init__(
-        self, parent=None, title="DAQ Move", ui_identifier: Optional[str] = None, **kwargs
-    ) -> None:
+    def __init__(self, parent=None, title="DAQ Move", ui_identifier: Optional[str] = None,
+                 preset: Optional[str] = None, **kwargs) -> None:
```
seb5g: what is preset for?

packages/pymodaq/src/pymodaq/control_modules/utils.py
```diff
@@ -505,26 +506,18 @@ def connect_tcp_ip(self, params_state=None, client_type: str = "GRABBER") -> Non
...
     def get_leco_name(self) -> str:
-        name = self.settings["main_settings", "leco", "leco_name"]
-        if name == '':
-            # take the module name as alternative
-            name = self.settings["main_settings", "module_name"]
-        if name == '':
-            # a name is required, invent one
-            name = f"viewer_{randint(0, 10000)}"
-            name = self.settings.child("main_settings", "leco", "leco_name").setValue(name)
+        name = (self.settings["main_settings", "leco", "leco_name"] or
```
seb5g, line 459: isn't that a boolean?

packages/pymodaq/src/pymodaq/scripting/devices.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def grab(self, keep=False) -> Optional[list[DataToExport]]:
+        return self._leco_device_wrapper.grab(keep=keep)
+
+    def stop_grab(self) -> None:
+       return self._leco_device_wrapper.stop_grab()
+
+
+
+class Dashboard:
```
seb5g: should it not inherit from device also?

packages/pymodaq/src/pymodaq/scripting/utils.py
```diff
@@ -0,0 +1,431 @@
...
+        self._director.ask_rpc(method="stop_grab")
+
+
+
+    def get_actuator_value(self) -> Future[DataActuator]:
+        future = Future()
+        self._send_position_future = future
+
+        self.set_remote_name()
+        self._director.ask_rpc('get_actuator_value')
```
seb5g, line 188: method= to be similar to other methods

packages/pymodaq/src/pymodaq/scripting/utils.py
```diff
@@ -0,0 +1,431 @@
...
+        self._director.ask_rpc(method="move_home")
+
+        return future
+
+    def move_abs(self, value: DataActuator | int | float | str ) -> Future[DataActuator]:
+        future = Future()
+        self._move_done_future = future
+
+        value = value_to_data_actuator(value)
+        serialize = SerializableFactory().get_apply_serializer
```
seb5g, line 206: maybe call it serializer ?

packages/pymodaq/src/pymodaq/utils/leco/daq_move_LECODirector.py
```diff
@@ -146,7 +146,8 @@ def move_rel(self, position: DataActuator) -> None:
                 #position = [data.m_as(self.axis_unit) for data in position.quantities]
                 position = np.full(self.shape, position.value(self.axis_unit)).tolist()
 
-        self.controller.move_rel(position=position)
+        (self.controller.
```
seb5g: why skipping a line?

packages/pymodaq/src/pymodaq/utils/leco/leco_director.py
```diff
@@ -109,7 +109,7 @@ def start_timer(self) -> None:
 
     def check_actor_connection(self) -> None:
         try:
-            self.controller.ask_rpc("pong", timeout=0.1)
+            self.controller.ask_rpc("pong", timeout=500000)
```
seb5g: it is not on purpose this timeout? every 100ms it checks the connection is alive. I think I remembered that

packages/pymodaq/src/pymodaq/dashboard.py
```diff
@@ -252,6 +260,71 @@ def do_things_after_preset(self, preset_name: str):
         for menu in (self.overshoot_menu, self.roi_menu, self.remote_menu, self.extensions_menu):
             menu.setEnabled(True)
 
+        if self._scripted_preset_load:
+            self._scripted_preset_load = False
+            self._command_tcpip.emit(ThreadCommand(LECODashboardCommands.APPLIED_PRESET_DONE, True))
+
+    def get_leco_name(self) -> str:
```
seb5g, line 276: should be refactored as a mixin I guess

packages/pymodaq/src/pymodaq/dashboard.py
```diff
@@ -895,7 +969,7 @@ def add_move(
         QtWidgets.QApplication.processEvents()
 
         actuator_widgets.append(QtWidgets.QWidget())
-        mov_mod_tmp = DAQ_Move(actuator_widgets[-1], plug_name, ui_identifier=ui_identifier)
+        mov_mod_tmp = DAQ_Move(actuator_widgets[-1], plug_name, ui_identifier=ui_identifier, preset=self.preset_name)
```
seb5g: still don't get what the preset name is used for in the control module arguments (but I may have missed it :-)

seb5g (changes requested):
I looked at the files first and raised/asked things. For the usability, I'll do another round. It would be good to have a script exemple one could execute to check everything it as it should be!

Overall looks good!

seb5g:
> Yes there should definitely be a mixin, it's actually not so hard to make something available and controllable through LECO and there's a quite a few redundant stuff. For now, i'm building a basic set of commands that work but I'll soon refactor into a better architecture.
> 
> I just made some dashboard features available for scripting:
> 
> ```python
> from pymodaq.scripting import Detector, Actuator, Dashboard
> 
> # Once a dashboard is started:
> dashboard = Dashboard()
> 
> 
> print(dashboard.get_presets().result())
> # >>> ['default']
> # Could make it so that it returns a future that completes when it's finished loading
> dashboard.apply_preset('default')

Yes this should be a blocking call

> 
> print(dashboard.get_configurations().result())
> # >>> ['default', 'my_custom_config']
> 
> # Same remark as dashboard.apply_preset
> dashboard.apply_configuration('my_custom_config')
> 
Yes this should be a blocking call


> print(dashboard.get_devices().result()) # returns the names
> # >>> {'actuators': ['Theta', 'Temperature', 'Power', 'Xaxis'], 'detectors': ['Det2D', 'Det0D', 'Det1D']}
> 
> devices = dashboard.get_scripting_devices()
> print(devices)
> # >>>  {'actuators': {'Theta': <Actuator object>, 'Temperature': <Actuator object>, 'Power': <Actuator object>, 'Xaxis': <Actuator object>}, 'detectors': {'Det2D': <Detector object>, 'Det0D': <Detector object>, 'Det1D': <Detector object>}}
> 
> 
> #Then each device can be used
> devices['actuators']['Theta'].move_abs('90°')
> ....
> ```

seb5g:
tested and added your two examples in the PR together with some nice local saving mechanism for the grabbed data.

seb5g:
I tried the scripts your proposed here in the PR. They work fine if the names in the dashboard are ok (I had some differences) with your theta called Angle in my case (could you check what are the names on the actual default preset?)

I added an exemple of how to simply and properly save an enlargeable file out of the data from that moves and acquisitions.

Works pretty well I'm quite happy with that => good job!

bernhardlang:
This new feature looks very neat. It would most probably allow to implement the planned sequencer business much easier (PR #931). The following 60-liner of pseudo code implicitly assumes a couple of things/functionalities being present in the scripting environment. It would then implement an entire 1D step scan absorption measurement (even re-measuring dark and reference included), compared to an over 100 lines yaml config file already for a corresponding sequencer using a state machine, re-measuring not yet included, code around not even counted. The data_processor could be a DataMixer instance which sends data to display and to store to the PyMoDAQ engine. Looks like state machine and dedicated sequencer won't perhaps be necessary after all. 
```
def get_dark():
    wait_future([shutter_excitation.set_value(0), shutter_reference.set_value(0)])
    wait_future(data_processor.acquire_dark())
    
def get_refence():
    wait_future([shutter_excitation.set_value(0), shutter_reference.set_value(1)])
    wait_future(data_processor.acquire_reference())

def ignore_exceptions(operation):
    if type(operation) == list:
        for op in operations:
            ignore_exceptions(op)
    try:
        operation()
    except:
        pass

try:
    # get device instances from dashboard extension / preset / gui config dialog
    shutter_excitation, shutter_probe, camera, (delay_line, scanner) \
        = config([('Excitation shutter', 'Actuator'), ('Probe Shutter', 'Actuator'), 
                  ('Camera' , 'Detector1D'), ('Scan Type', 'Scanner1D')])
except:
    return # catch gui cancel

data_processor = AbsorptionProcessor()

try:
    data_processor.config() # config by gui
except:
    return # catch gui cancel
camera.data_ready.connect(data_processor.take_data) # or whatever preferred PyMoDAQ communication means

try:
    wait_future(camera.start_grabbing())
    delay_future = delay_line.set_value(scanner.init())
    while not scanner.done:
        if not data_processor.ready_for_absorption:
            if data_processor.need_dark:
                get_dark()
            if data_processor.need_reference:
                get_reference()
            shutter1_future = shutter_excitation.set_value(1)
            shutter2_future = shutter_reference.set_value(0)
        wait_future([shutter1_future, shutter2_future, delay_future])
        wait_future(data_processor.acquire_absorption())
        delay_future = delay_line.set_value(scanner.advance())

except DeviceError as e:
    gui.display_device_error(e)
except AbortError:
    gui.message_box("Aborted")

ignore_exceptions([camera.stop_grabbing,
                   lambda: shutter_excitation.set_value(0),
                   lambda: shutter_reference.set_value(0)])
```

packages/pymodaq/src/pymodaq/control_modules/utils.py
malik-irain: No that's the string, but if the setting is not set, None or '' (empty string), then it will act like False in this case. To be clearer, here is an example output:

```python
>>> 'a host' or 'localhost'
'a host'
>>> None or 'localhost'
'localhost'
>>> '' or 'localhost'
'localhost'
```

packages/pymodaq/src/pymodaq/control_modules/utils.py
malik-irain: same explanation!

packages/pymodaq/src/pymodaq/control_modules/daq_viewer.py
malik-irain: I wanted to prefix the leco names by preset, for example: 'default.det0d' but dot is not a valid naming character in  leco. I kept it but it is now useless

packages/pymodaq/src/pymodaq/control_modules/daq_move.py
malik-irain: same explanation as daq_move

packages/pymodaq/src/pymodaq/control_modules/utils.py
malik-irain: same explanation as localhost

packages/pymodaq/src/pymodaq/scripting/utils.py
malik-irain: yes! It's not the only one where it's missing!

packages/pymodaq/src/pymodaq/scripting/devices.py
malik-irain: It could, but Dashboard, Actuator and Dectector are classes around real LECOWrappers to expose only a selection of methods (because some LECO logic rely on names to find RPC calls). In the beginning I didn't want Actuator and Detector to inherit the same class.

It seemed a little bit more logical to separate as they work differently but I need to think a little bit more about this

packages/pymodaq/src/pymodaq/scripting/utils.py
malik-irain: But it is the serialize function in the end, it feel more natural to call serialize(value) than serializer(value)

packages/pymodaq/src/pymodaq/utils/leco/daq_move_LECODirector.py
malik-irain: I guess it's from auto reformating

packages/pymodaq/src/pymodaq/dashboard.py
malik-irain: Yes ;)

packages/pymodaq/src/pymodaq/utils/leco/leco_director.py
malik-irain: Default timeout triggers too many messages, so for readability over LECO messages on the network, I increased the value that I forgot to put back to default.

Also I think the implementation is missing something like disable the director when the actor stop responding but it's another topic

docs/src/developer_folder/make_component_leco_available.rst
```diff
@@ -0,0 +1,474 @@
+.. _leco_component_mixin:
+
+Making a PyMoDAQ Component Available Through LECO
+==================================================
+
+This page is a developer guide explaining how to expose a new PyMoDAQ component
```
seb5g: an existing pymodaq component

docs/src/developer_folder/make_component_leco_available.rst
```diff
@@ -0,0 +1,474 @@
...
+over the `LECO`_ network.  The ``DashBoard`` class is used as a running example
```
seb5g: over the pymodaq integrated `LECO`_ network.

docs/src/developer_folder/make_component_leco_available.rst
```diff
@@ -0,0 +1,474 @@
...
+
+
+Architecture
+~~~~~~~~~~~~~
+
+.. figure:: /image/component_leco/leco_scripting_architecture.svg
+   :alt: LECO scripting architecture in PyMoDAQ
+   :align: center
+   :width: 150%
+   :class: no-max-width-figure
```
seb5g: the figure is too wide and is cropped by sphinx
It is ok to follow using the numbers  but I'm not sure it provides much understanding. Maybe a second image with just the outer methods would be best to start with then show the inner mechanisms?!?

docs/src/developer_folder/make_component_leco_available.rst
```diff
@@ -0,0 +1,474 @@
...
+   PyMoDAQ component (the Actor side) over the network.
+
+Part 1 – LECO-Enabling a Component (PyMoDAQ Side)
+--------------------------------------------------
+
+The process has five steps:
+
+1. Inherit from :class:`~pymodaq.utils.leco.pymodaq_listener.LECOComponentMixin`
+   and implement its abstract methods.
+2. Define the command and RPC-method name enumerations.
```
seb5g, line 44: do you think some declarations could be simplified by using directly the enums? 
Do you think there could be something a bit more evolved than an enum storing commands in and out? For instance the LECOMyComponentCommands.GET_INFO is associated with LECOMyComponentCommands.SEND_INFO so a short class could define the in and out commands and emit the correct signal with a given attribute:

```python
class BaseCommand:
    in: str
    out: str

    def __init__(self, leco_mixin):
        self._mixin = leco_mixin

    def process_command(self, attribute: Any):
        self._leco_commands_signal.emit(self.out, info))



class Info(BaseCommand):
    in = 'GET_INFO'
    out = 'SEND_INFO'

```

would that simplify things or overcomplexify them?

docs/src/developer_folder/make_component_leco_available.rst
```diff
@@ -0,0 +1,474 @@
...
+    are actually aliases for ``ActorListener``
+
+Step 2 – Define Command Enumerations
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Two sets of names are needed:
+
+* **RPC method names** (strings exchanged over the network) – defined in
+  :mod:`pymodaq.utils.leco.rpc_method_definitions` as ``StrEnum`` subclasses.
+* **ThreadCommand names** (used internally to cross the Qt thread boundary) –
```
seb5g, line 112: are these the union of the others? could we not simplify them also?

docs/src/developer_folder/make_component_leco_available.rst
```diff
@@ -0,0 +1,474 @@
...
+  defined in :mod:`pymodaq.utils.leco.pymodaq_listener`.
+
+However, they are generally the same names.
+
+For each new component you need one enum for incoming commands (Director →
+Actor, i.e. what your component *receives*) and one for outgoing commands
+(Actor → Director, i.e. what your component *sends back*):
```
seb5g: adds that it seems to add in complexity but that in fact it helps to understand who is doing what...

docs/src/developer_folder/make_component_leco_available.rst
```diff
@@ -0,0 +1,474 @@
...
+           super().__init__(name, handler_class=MyActorHandler, **kwargs)
+
+**Dashboard example** – ``get_devices``,
+``get_configurations``, ``apply_configuration``, ``get_presets``,
+``apply_preset`` were added to ``ActorHandler``.  Each simply emits the matching
+``LECODashboardCommands`` on ``cmd_signal``:
+
+.. code-block:: python
+
+   def get_devices(self):
```
seb5g, line 191: I have the feeling these are just extraboilerplate to pass the rpc message to a threadcommand. One could also probably simplify it or automatize it?

docs/src/developer_folder/make_component_leco_available.rst
```diff
@@ -0,0 +1,474 @@
...
+
+``ActorListener.queue_command`` is the outgoing channel: it receives
+``ThreadCommand`` objects emitted on ``_leco_commands_signal`` and translates
+them into ``ask_rpc`` calls sent to the Director.  Add ``elif`` branches for
+your new commands:
+
+.. code-block:: python
+
+   # in your MyActorListener (or patched into ActorListener)
+   def queue_command(self, command: ThreadCommand) -> None:
```
seb5g, line 213: could also be automatized? bundled into the object dealing with Getinfo/send _info

docs/src/developer_folder/make_component_leco_available.rst
```diff
@@ -0,0 +1,474 @@
...
+           info = self.compute_info()
+           self._leco_commands_signal.emit(
+               ThreadCommand(LECOMyComponentCommands.SEND_INFO, info)
+           )
+
+**Dashboard example**:
+
+.. code-block:: python
+
+   def process_leco_commands(self, status: ThreadCommand) -> None:
```
seb5g, line 289: seeing this structure I'm almost sure we should bundle things into a dedicated object, a factory could then just provide the right one. should discuss this

docs/src/developer_folder/make_component_leco_available.rst
```diff
@@ -0,0 +1,474 @@
...
+           # 2 – register Director-side callbacks
+           self._listener.register_rpc_method(self.something_done)
+           # for binary replies use register_binary_rpc_method:
+           # self._listener.register_binary_rpc_method(self.some_data, accept_binary_input=True)
+
+           # 3 – get a communicator and a Director
+           self._communicator = self._listener.get_communicator()
+           self._director = Director(actor=actor_name,
+                                     communicator=self._communicator, **kwargs)
+           atexit.register(self.close)
```
seb5g: what is doing this?

seb5g: make sure close is called before garbage collected?

docs/src/developer_folder/make_component_leco_available.rst
```diff
@@ -0,0 +1,474 @@
...
+           except (InvalidStateError, AttributeError):
+               pass
+
+Binary replies (e.g. a ``DataToExport``) require ``register_binary_rpc_method``
+and deserialization:
+
+.. code-block:: python
+
+   # registration
+   self._listener.register_binary_rpc_method(self.set_data, accept_binary_input=True)
```
seb5g, line 516: why is the name different AND have an accept_binary_input=True argument seems superfluous

docs/src/developer_folder/make_component_leco_available.rst
```diff
@@ -0,0 +1,474 @@
...
+   # connect to a running Dashboard and load a preset
+   dashboard = Dashboard()
+   dashboard.apply_preset('default').result()
+
+   # get all loaded modules
+   devices = dashboard.get_scripting_devices()
+   theta  = devices['actuators']['Angle']
+   camera = devices['detectors']['Camera']
+
+   # blocking move, then async snaps
```
seb5g: here it's a sync snap

docs/src/conf.py
```diff
@@ -176,7 +176,8 @@
 #
 # html_theme_options = {}
 html_css_files = [
-    'css/overflow.css'
+    'css/overflow.css',
```
seb5g, line 179: what is in the css?

packages/pymodaq/src/pymodaq/scripting/utils.py
seb5g: ok

seb5g (changes requested):
We have to discuss about refactorising more by myabe using some factory pattern

docs/src/developer_folder/make_component_leco_available.rst
malik-irain: When I was implementing, the PyMoDAQ part was actually something I need because it's not obvious what component communicate with each other

For the size, it is wide but is not cropped on my side:

image

Anyway, I put it back on normal size for now

docs/src/developer_folder/make_component_leco_available.rst
malik-irain: Yes, all the Enum are similar in the end, the is one set for RPC/Network communication and one set for thread communication with same names and same values, this could also be refactored

docs/src/developer_folder/make_component_leco_available.rst
malik-irain: I agree about the feeling, but I think all these methods need to exist because of pydantic and the RPC library used by pyleco that find stuff by name

docs/src/developer_folder/make_component_leco_available.rst
malik-irain: Yes, so that component are unregistered on LECO network

docs/src/developer_folder/make_component_leco_available.rst
malik-irain: I agree, but this is from pyleco

docs/src/conf.py
malik-irain: For the big image

docs/src/developer_folder/make_component_leco_available.rst
malik-irain: thanks!

malik-irain:
I kept the figure in the doc but resized it and took into account the review (except for refactoring around the commands). So the doc is up to date with the current version!

If you need to merge it @seb5g it should be good enough for some alpha

seb5g (approved):

PR #926, opened 2026-02-20
Feature/action_LED

Ashwolaa:
Hey little PR,

- Feature:
 1 new parameter_type:
   - Action + LED which combines a push button with a LED. The value of the LED can then be connected to the success of the action. Overall, this allows to merge two parameter items into one. Such as probing/testing modules in module_managers.

- Factorization:
   - Using a common method for building nested menus

- Hotfix:
   - Fixing error in setup_dock when parent is a dockarea


packages/pymodaq_gui/src/pymodaq_gui/parameter/pymodaq_ptypes/action_led.py
seb5g, line 1: how do you think it will be saved/loaded from the xml string? Could you check? My pyqtgraph PR on serialization would definitely solve this problem but they are not in a hurry to answer...

seb5g (changes requested):

packages/pymodaq_gui/src/pymodaq_gui/parameter/pymodaq_ptypes/action_led.py
Ashwolaa: This seems to behave fairly well. Label is the text of the action and the value of the LED is defined at the end. No problem on my side when trying to save/edit/load it on the parameter_ex.

`
<anactionled type="action_led" title="Action + LED" visible="1" removable="0" readonly="0" label="Test">1</anactionled>`

Ashwolaa: https://github.com/user-attachments/assets/f176b96c-e072-491f-8da3-285854556300

seb5g (approved):

PR #929, opened 2026-02-21
Feature/modules manager

Ashwolaa:
This PR focuses on the modules_manager layout to make it more compact and organized with respect to the data. 
This tree structure could then be reused in other GUI with parameter trees such as in the H5Browser/Datamixer

Using ActionLED from #926

https://github.com/user-attachments/assets/7bbb368b-a9e5-48a4-bc92-bf45314bb2b0


seb5g (approved):
it is indeed more compact. wondering if this will cause troubles in the extensions... i saw you took care of modifying them also but ...

PR #930, opened 2026-02-22
using dev keyword for testing

Ashwolaa:
During the development of the datasaving, I looked a bit on how the packages were tested on github.

Currently, we need to give a list of the libraries that we want to test for each package. This feels a bit cumbersome as they need to be repeated for each package and when we have multiple libraries (such as hdf5 backend or qt), they may all need to be listed explicitely.

For these different backends, we could also test
`pip install -e "packages/pymodaq_data[dev]"`

which allows the full installation of the different backends.

In this case, we only need to focus on listing the relevant packages in the dev part in the project.optional-dependencies (at least this is how I understand the dev part).

I put it as a PR but this is more about triggering discussions on the subject with an easily trackable example


malik-irain:
You can provide the -d option to the install-packages.py script, so it is better to write:

```bash
python install-packages.py -d -u pymodaq_data
```
instead of:

```bash
python install-packages.py -u pymodaq_data
pip install -e "packages/pymodaq_data[dev]"
```

as it will install dev dependencies for pymodaq_data and pymodaq_utils

seb5g:
@malik-irain could you check the PR?

malik-irain:
the `--qt-api` option doesn't exist:
```bash
$ pytest  --qt-api=pyqt6
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: --qt-api=pyqt6

```

And giving QT_API variable through environment isn't always respected by pytest. The PYTEST_QT_API variable seems to work on linux and windows but I have an issue (locally) with linux and pyqt5.

So can you try to add `PYTEST_QT_API=${{ matrix.qt-backend }}` in  the environment of the github action (see [here](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-variables)) and try on your fork if it works?

Ashwolaa:
I did what you suggested and I ran it, it seems to go through all the qt-backends
https://github.com/Ashwolaa/PyMoDAQ/actions/runs/22716636555/job/65869461175

malik-irain:
For me this is good, is there other points you would like to add @Ashwolaa ?

Ashwolaa:
nope, fine by me thanks for the checking!

PR #931, opened 2026-02-25
first state machine implementation with simple test

bernhardlang:
On top of what we discussed, tests/sequencer/test_sequencer.py tests the state machine and its factory in the context of a very basic implementation of a sequencer together with some dummy devices and glue code to simulate the operations to be controlled by the dashboard. A delay line is added to the definition file step-scan.yml so that the two types of actuators, "change a continuous value" and "perform a 'binary' action" can be treated / tested separately if that turns out to be useful / needed. 

The preamble of the definition file (until the first line containing 'states:') is not yet used, for future use to configure the sequencer. The actions therein are taken care of by the rudimentary sequencer. The state machine only connects states by transitions.

Before going further it might be useful to discuss format and content of the yaml configuration file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -131,3 +131,4 @@ venv.bak/
 
 
 **/README.rst
+*~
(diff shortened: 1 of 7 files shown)


seb5g:
I think the PR #923 may be very nice for the purpose of your sequencer! For the rest of the discussion I think it is best to wait for your coming in Toulouse!

PR #932, opened 2026-02-25
Export repr() of non-printable characters in XML

astroloba:
For some obscure reasons, certain devices (like Keithley 2600 series) generate VISA resource names contain a null character (at least when using `pyvisa-py`):
`'USB0::1510::9782::1234567\x00::0::INSTR'`

If the resource name is stored as a list in the parameter tree of the instrument plugin, then this non-printable character causes an error when editing/loading a dashboard preset (but not when saving it):
```
  File "/usr/lib/python3.12/xml/etree/ElementTree.py", line 1204, in parse
    tree.parse(source, parser)
  File "/usr/lib/python3.12/xml/etree/ElementTree.py", line 569, in parse
    self._root = parser._parse_whole(source)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
xml.etree.ElementTree.ParseError: not well-formed (invalid token): line 1, column 6121

```

This PR aims to make XML export/import more robust with such non-printable characters. Instead of exporting `str('literal string contents with maybe non-printable characters')` it instead exports `str(representation of string contents)`, which will, like in the Python console, automatically generate quotes and escape characters as needed.

This change was tested with success with a device such as described above.

### DIFF
--- a/packages/pymodaq_gui/src/pymodaq_gui/parameter/ioxml.py
+++ b/packages/pymodaq_gui/src/pymodaq_gui/parameter/ioxml.py
@@ -92,7 +92,7 @@ def add_text_to_elt(elt, param):
         text = str([param.value().red(), param.value().green(), param.value().blue(), param.value().alpha()])
     elif param_type == 'list':
         if isinstance(param.value(), str):
-            text = "str('{}')".format(param.value())
+            text = "str({})".format(repr(param.value()))
         elif isinstance(param.value(), int):
             text = 'int({})'.format(param.value())
         elif isinstance(param.value(), float):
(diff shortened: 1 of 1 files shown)


Ashwolaa:
I would take the occation to switch to a f-string which is more condensed and easier to read so:

```
if isinstance(value, str):
    text = f"str({value!r})"
elif isinstance(value, int):
    text = f"int({value})"
```

seb5g:
I agree with @Ashwolaa , @astroloba could you remove the .format string to use fstrings?

astroloba:
Done! I have also performed a small factorization to avoid calling param.value() every time. If you have any trouble with this, you can discard commit https://github.com/PyMoDAQ/PyMoDAQ/pull/932/commits/81881c56d8ee9605a17bbaaf6b7cbb2c4db65f90.

seb5g (approved):
perfect!

PR #937, opened 2026-03-01
using first element of the list as default backend

seb5g:
adding raising an error if unknown backend or wrong type


malik-irain:
Tests don't pass!

PR #938, opened 2026-03-01
Feature/overshoot manager

seb5g:
Building from the ManagerBase, I propose to refactor entirely the overshoot mechanism and use the ManagerBase mechanisms.

This new Overshooter object allow to create a scalable Parameter from the selection of a 0D data (detector, roi, or actuator). 

image

This Parameter will allow to define a condition to emit an overshoot signal:

image

Once triggered, everytime the chosen data is emitted (grab_done or current_value), its value is compared to the set point in the parameter either above or below. If the condition is met, a given Configurator entry will be executed:
image

there is therefore a tight link between the configurator and the overshooter:

image

The result of a given overshoot is therefore a Configurator Entry. So far one could set module settings, move actuators, wait or init. Related to the use with the Overshooter, new special configurator subentries have been developed:

image

* stop: stops a given control module
* stop_all: stops all control modules
* stop extension: stop a given extension (if opened)

The CustomExt interface has been changed to add a **stop** mandatory method in order to stop any action related to the control modules that is called by the special configurator subentry **stop_extension**.

Overall it allowed me to remove from the DashBoard everything related to overshoot, just like I did with the preset manager and configurator.

Next PR on the same principle will be related to ROI and remote manager. 

For the Overshooter the execute action is checkable allowing an overall connection/disconnection while the individual triggers can fine tweak the overshoots:
image

Next PR on the same principle will be related to ROI and remote manager. 

@Ashwolaa did you somehow worked on a syncWidget to sync ROI object on the view and the ROI ParameterTree representation in the ROIManager?


Ashwolaa:
`@Ashwolaa did you somehow worked on a syncWidget to sync ROI object on the view and the ROI ParameterTree representation in the ROIManager?`

Nope but that should be well suited I guess. Synchronization with parametertree should be fine in theory

seb5g:
> Nope but that should be well suited I guess. Synchronization with parametertree should be fine in theory

Ok I'll look into it before I start on the roi manager refactorisation

packages/pymodaq/src/pymodaq/utils/managers/overshoot/utils.py
```diff
@@ -0,0 +1,116 @@
...
+
+def find_last_index(list_children: list = [], name_prefix='', format_string='02.0f'):
+    # Custom function to find last available index
+    child_indexes = ([int(par.name()[len(name_prefix) + 1:]) for par in list_children if name_prefix in par.name()])
+    if child_indexes == []:
+        newindex = 0
+    else:
+        newindex = max(child_indexes) + 1
+    return f'{newindex:{format_string}}'
+
```
Ashwolaa, line 26: move this as a utils.py in the parent folder? Same syntax as preset

Ashwolaa:
Mostly, on UX. 
- We get a pop up when we execute it (while it is already toggled) with same syntax. I guess, it should only happen when we toggle it?
- At this point, it could maybe nice to have a button/shortcut to open the selected configurator from the overshoot or to make a new one
- Stop Module can take false as value? I guess the name could be Activate? with False = Stop and True = Snap (aka query value)?
- I think we could add some additional actions in the overshoot not directly connected to modules but more on the user side such as making a sound / making the screen flash / sending a mail or a text ^^
- double clicking on the data in the list could forward it into the preset
- changing overshoot preset breaks the add mechanism (no more entries available). It requires to press update data to get the expected behavior. I guess this should be done automatically and should be implemented as a function in the manager_base.py from which the others can inherit.
- From this, it would be interesting to add a logical mechanism. Currently, the overshoot is set through the OR operation but we could add the AND with others to make it even more customizable. I guess, this is the kind of operation that will be used afterwards in the statemachine? This could also be done in another PR.
- There are some synchronization problem, with the available configurators. I am limited to select default. I am not quite sure on the flow of actions to get there. But restarting the dashboard/making new preset does not change it.
image

- Instead of having a dedicated trigger entry, one could use a boolean approach directly on the parent settings with a checkbox or a LED to activate it or not. Overshoot 1 : [ x ] 

Looks good otherwise, imo the splash screen with the full unreadable line is a bit overkilled. It works fine for loading the preset but one could maybe think of having an extra widget next to the splash instead (or no splash...? ^^)

seb5g:
@malik-irain this is ready for final review. However there are some seg faults in the tests... if you could have a look on the way!

malik-irain:
Seems like replacing `_execute_entry` with `execute_entry` in the dashboard test fixed the segfault issue

seb5g:
cool!!

PR #955, opened 2026-03-10
Rationalize the origin/names of the data produced in the viewer/actuator and display in modules manager

seb5g:
I updated a recent PR that broke the way the data are to be used within the ModuleManager

the origin is kept as from the title of the control module

the name is built using slashes (like files) see below
image

using simple strings parameter (allowing double click and copy paste (important for the datamixer model equation)

the title of the setting is the origin, the value is the name and the name is the full_name

### DIFF
--- a/packages/pymodaq/tests/utils/managers/modules_manager_test.py
+++ b/packages/pymodaq/tests/utils/managers/modules_manager_test.py
@@ -66,9 +66,9 @@ def manager(detectors, actuators):
 # Helpers to build controlled DataToExport for tree-building tests
 # ---------------------------------------------------------------------------
 
-def make_raw_dte(det_title: str, channel_name: str = 'CH0') -> DataToExport:
+def make_raw_dte(det_title: str, dwa_name: str = 'DWA_NANE') -> DataToExport:
     """DTE with a single raw 1D channel from a detector."""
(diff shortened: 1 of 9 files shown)


seb5g:
did you touch at the actuators by the way @Ashwolaa

Ashwolaa:
I don't remember so. I think only the widget to display their positions when one wants to test them.

packages/pymodaq/tests/utils/managers/modules_manager_test.py
```diff
@@ -319,34 +316,7 @@ def test_tree_cleared_on_repopulate(self, manager):
...
-            if dwa.origin != 'Det1':
-                dte.append(dwa)
-
-        with patch.object(manager, 'grab_data', return_value=dte):
-            manager.get_det_data_list()
-
-        ch_param = manager.settings.child('probe_data').children()[0].children()[0]
-        roi_groups = [p for p in ch_param.children() if p.type() == 'group']
-        assert {g.name() for g in roi_groups} == {'ROI_00', 'ROI_01'}
+        pass
```
Ashwolaa: why keep the test if no assert?

Ashwolaa:
good for me

Ashwolaa (approved):

PR #961, opened 2026-03-19
Bugfix/data dtype saving

seb5g:
porting PR #960 to dev


seb5g:
@Ashwolaa Be aware of this change. I checked it doesn't change the behaviour of the fill_value to init scanner arrays. It can however save a lot of space on disk if used properly!

PR #970, opened 2026-03-26
added icon color to QAction.set_icon

bernhardlang:


packages/pymodaq_gui/src/pymodaq_gui/managers/action_manager.py
```diff
@@ -65,13 +65,14 @@ def clicked(self):
...
-    def set_icon(self, icon_name: str = None):
+    def set_icon(self, icon_name: str = None,
+                 icon_color: Union[QtGui.QColor, bytes, str]=None):
         if icon_name is None:
             if self.isChecked():
                 icon_name = self.icon_checked
             else:
                 icon_name = self.icon_unchecked
-        self.setIcon(create_icon(icon_name))
+        self.setIcon(create_icon(icon_name, icon_color))
```
seb5g, line 75: we have an issue to solve as if icon_name is None, one get the icon_name from self.icon_checked or unchecked which are MaterialIcon and not the name of the icon. I'll do a commit in this issue so that create_icon can apply coloring if icon_name is an instance of a MaterialIcon

seb5g (approved):

Issue #973, opened 2026-03-27
Bug : bad Configurator widgets synchronisation


daarklink:
@Ashwolaa 

When several Configurator widgets are loaded, there are poorly synchronized with each other.  In this case, the expected Configurator is "tb" but the one displayed is "default". Only the toolbar in the Configurator window is correct. The value in memory is actually "tb", and the widget also returns the correct value when querying currentText. There is also a synchronization issue when changing the combo box entry, as it only updates on the second attempt.

A pull request will be coming with fixes for cases in "manager_base", please take a deeper look at the synchronization issues in "widget_sync".

https://github.com/user-attachments/assets/bdefb0f6-1243-4058-9ad3-8f00378d1337

Ashwolaa:
It is a bit hard for me to answer from your example as I have no information on how things are getting synced together here. Neither how you propagate the information.

You are syncing both Preset and Configurator here, changing the values on both sequentially?

Synchronizing combobox is a bit tricky as one of the property depends on the other. Usually, it is easier to just setup the items from the WidgetSync value and only use the BIDIRECTIONAL on the selection. 

I will look into it but can you maybe direct me towards your implementation to understand how you currently do it? or show a smaller example where you observe similar failing? 


In the mean time, I published a branch (https://github.com/Ashwolaa/PyMoDAQ/tree/hotfix/synchronization_comboboxes) in which comboboxes are synchronized either from the widget_sync or from a combobox (check 2_dict_sync_example.py). For this, I just modified the set_items of combo.py
This decomposes cleanly the signal into first updating the elements and then selecting the chosen index

```
  def set_items(self, items: list[str], item_selected: str = None):
      self.clear()
      self.addItems(items)
      self.items_changed.emit(items)
      if item_selected is not None:
          self.setCurrentText(item_selected)
```

daarklink:
@Ashwolaa 

I'm experiencing the issue again when using an external_toolbar : selecting a Preset entry from outside the configurator does not update the configurator's entries. The two widgets are only synchronized from the configurator itself.

Outside the Configurator, the preset_filename method (@preset_filename.setter) in the Configurator class is never called.

Ashwolaa:
For the configurator and how it should react with a change in preset,  you should check directly with @seb5g as I did not do the implementation. Is the preset synchronized in the toolbar but not the configurator? 
[...]

PR #974, opened 2026-03-27
fix bad synchronization between widgets

daarklink:
#973 

Update the widget sync value instead of relying on the combo box currentText. Previously, one widget (in the Configurator window) was updated and the others were synced from it. This fix resolves the issue for the launcher, but does not address the root cause in the widget synchronization logic.

### DIFF
--- a/packages/pymodaq_gui/src/pymodaq_gui/managers/manager_base.py
+++ b/packages/pymodaq_gui/src/pymodaq_gui/managers/manager_base.py
@@ -193,9 +193,7 @@ class ManagerBase(CustomExt):
 
     @entry.setter
     def entry(self, entry_name: str):
-        value = self.entries_sync.value
-        value['current'] = entry_name
-        self.entries_sync.set_value(value)
+        self.entries_sync.set_value({**self.entries_sync.value, 'current': entry_name})
 
     @property
     def entry_filepath(self) -> Path:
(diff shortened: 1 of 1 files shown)


seb5g:
@Ashwolaa do you reckon this is the way we should update the value? It seems to propagate well in this case!

Ashwolaa:
yes this is better, acting directly on the sync should be prefered whenever possible as its value contains the dict with all the information

seb5g:
is there a way to update only part of the dict in one line of code?

Ashwolaa:
Can you be more specific? This is how I would do.
sync.value = {**sync.value, 'the_key_to_change': the_new_value}

In the propagation, there will be a cross check on each widget to see if the value associated to the key has changed.

Envoyé : samedi 28 mars 2026 20:29
Cc : Ashwola ***@***.***>; Mention ***@***.***>
Objet : Re: [PyMoDAQ/PyMoDAQ] fix bad synchronization between widgets (PR #974)

[https://avatars.githubusercontent.com/u/6609955?s=20&v=4]seb5g left a comment (PyMoDAQ/PyMoDAQ#974)<https://github.com/PyMoDAQ/PyMoDAQ/pull/974?email_source=notifications&email_token=AMFLHBI7G73HFSOYPCEVHA34TARZDA5CNFSNUABFM5UWIORPF5TWS5BNNB2WEL2JONZXKZKDN5WW2ZLOOQXTIMJUHA3DMOJUGE32M4TFMFZW63VHNVSW45DJN5XKKZLWMVXHJNLQOJPWG33NNVSW45C7N5YGK3S7MNWGSY3L#issuecomment-4148669417>

is there a way to update only part of the dict in one line of code?

PR #975, opened 2026-03-29
Adding a None check before disconnect

Ashwolaa:
Fix AttributeError bug when the _leco_client (ActorListener) is None

```
(pymodaq_3.13) D:\Work\PyMoDAQ>dashboard
  File "D:\Work\PyMoDAQ\packages\pymodaq\src\pymodaq\control_modules\daq_move.py", line 188, in process_ui_cmds
    self.init_hardware(cmd.attribute[0])
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "D:\Work\PyMoDAQ\packages\pymodaq\src\pymodaq\control_modules\daq_move.py", line 439, in init_hardware
    self.connect_leco(False)
    ~~~~~~~~~~~~~~~~~^^^^^^^
  File "D:\Work\PyMoDAQ\packages\pymodaq\src\pymodaq\control_modules\daq_move.py", line 844, in connect_leco
    super().connect_leco(connect)
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "D:\Work\PyMoDAQ\packages\pymodaq\src\pymodaq\utils\leco\pymodaq_listener.py", line 483, in connect_leco
    self._leco_commands_signal.disconnect(self._leco_client.queue_command)
                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'queue_command'
```

### DIFF
--- a/packages/pymodaq/src/pymodaq/utils/leco/pymodaq_listener.py
+++ b/packages/pymodaq/src/pymodaq/utils/leco/pymodaq_listener.py
@@ -479,10 +479,11 @@ class LECOComponentMixin:
             self._leco_client.start_listen()
         else:
             self._leco_commands_signal.emit(ThreadCommand(LECOCommands.QUIT, ))
-            try:
-                self._leco_commands_signal.disconnect(self._leco_client.queue_command)
-            except TypeError:
-                pass  # already disconnected
+            if self._leco_client is not None:
+                try:
+                    self._leco_commands_signal.disconnect(self._leco_client.queue_command)
+                except TypeError:
+                    pass  # already disconnected
 
     def get_leco_name(self) -> str:
         """Return the LECO component name used to register on the network.
--- a/packages/pymodaq/src/pymodaq/utils/leco/pymodaq_listener.py
+++ b/packages/pymodaq/src/pymodaq/utils/leco/pymodaq_listener.py
@@ -461,8 +461,8 @@ class LECOComponentMixin:
         sent from the Qt thread.
 
         When *connect* is ``False``, a :attr:`~LECOCommands.QUIT` command is
-        emitted to stop the listener gracefully, and the signal/slot connection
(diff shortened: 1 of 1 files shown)


malik-irain:
It shouldn't happen, how did you get the issue?
Calling `connect_leco(True)` creates `self._leco_client` and connects `self._leco_client.cmd_signal` if it doesn't exist. So you should never be in the case where when disconnecting  `self._leco_client` is `None`.

Also, if you disconnect ` self._leco_client.cmd_signal` and never set ` self._leco_client` to `None` again, the signal is never reconnected even when leco is reconnected

Ashwolaa:
Good question, I noticed it when I was trying to fix a threading issue with DAQ_move #976

But I am thinking that it could be due to an ill installation I cannot reproduce it clearly now that I uninstalled pyleco and went with the install_packages.py.

Let me revert it from #976 and pause this.

malik-irain:
Do you remember what version of pyleco was installed? You can find it in `.pymodaq/environments/<env_name>/` If we are lucky, one of the file may correspond to a state of the environment before using the script

Ashwolaa:
I check it and there is no detectable change in the environment except my commit history inside pymodaq.
Th env was pyleco==0.6.0.post1
Maybe it is a relicat of some of the work I made on the actor part on another branch.. not clear

seb5g:
what do we do with this then?

Ashwolaa:
Not sure, I guess we can close it. I am trying to fix the segfaults in #996 but no success so far and using this syntaxt does not seem to help.
My guess is that the (re)execution of preset is now too fast for LECO to keep track so there might be some conflict because of this.

seb5g:
ok I'm closing it

Issue #922, opened 2026-02-17
[BUG] daq_move -> move_abs is not called when asking deplacements smaller than 1e-8.


bernhardlang:
### Describe the bug

In a (delay-line) move derived from the templates, small values do not lead to a call of move_abs or move_rel. It seems like moving in steps of picoseconds or femtoseconds is not forseen, even with _eplison set to < 1e-15. The behavior is reproduced on daq_move_Mock from pymodaq_plugins_mock.

### Expected Behavior

When the difference between the current value and the value asked for is larger than _epsilon, move_abs or move_rel should always be called.

### Steps To Reproduce

- run daq_move_Mock.py
- enter a number which is less apart than 1e-8 from the current value
- press the arrow button
- nothing happens
- enter a number which is further apart than 1e-8 from the current value
- press the arrow button
- the current value changes 
This is independent of the Epsilon value set in the move's parameters.

### Reference(s)

```markdown

```

### Environment

```markdown
- OS: Debian 12
- Pymodaq: 5.1.3
- python env: 3.11.3
```

seb5g:
@bernhardlang could you check the PR #979 fixed your issue?

bernhardlang:
Running my move plugin or the  daq_move_Mock.py from the plugins_mock repo with the PyMoDAQ-dev branch installed, I get an exception which is probably related to PR #981 (since both exception and PR mention status bar). Looks like that PR breaks the daq_move interface. Which commit should I check out to test #922?

Traceback (most recent call last):
  File "/home/bernhard/comp/pymodaq/pymodaq_plugins_mock/src/pymodaq_plugins_mock/daq_move_plugins/daq_move_Mock.py", line 126, in <module>
    main(__file__)
  File "/home/bernhard/comp/pymodaq/PyMoDAQ-dev/packages/pymodaq/src/pymodaq/control_modules/move_utility_classes.py", line 203, in main
    prog = DAQ_Move(widget, title=title, actuator=act)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/bernhard/comp/pymodaq/PyMoDAQ-dev/packages/pymodaq/src/pymodaq/control_modules/daq_move.py", line 132, in __init__
    self.ui = DAQ_Move_UI(parent, title)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #979, opened 2026-03-30
Bugfix/actuator value comparison

seb5g:


seb5g:
fixes #922 

so far initial move_abs was checking the curretn value and target were equal to make a move. This simple equal: `self._current_value == value` was ok only within the regular float value usual comparison minimal value so femto thing were always equal.

Now the initial check is:  `value.equal_to(self._current_value, self.epsilon)` with the method equal_to is introduced in this PR. It does a real comparison within epsilon to assert if two data object are equal

PR #987, opened 2026-04-02
add the FORCE_QT_API environment variable to force the use of the chosen backend

seb5g:
…

otherwise qtpi use the one already present in sys.modules in whatever order it chose

### DIFF
--- a/packages/pymodaq/tests/conftest.py
+++ b/packages/pymodaq/tests/conftest.py
@@ -1,0 +1,16 @@
+import pytest
+from qtpy import QtWidgets
+
+
+@pytest.fixture
+def qapp():
+    app = QtWidgets.QApplication.instance()
+    if app is None:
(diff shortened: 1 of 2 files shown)


seb5g:
@malik-irain all the tests fails with pyqt5... seems like qtbot in this case is not active... could you have a look please?

malik-irain:
It won't be possible because forcing the backend is done from the config, which is always pyqt5 in the test cases, but we also force it to be something else with an environment variable `PYTEST_QT_API`.

I believe the only way to ensure a qt api is if pymodaq_gui (or this part of code) is imported first in any packages that uses qt.

seb5g:
it started to bug all around on my code for some unknown reasons... I reinstalled an env and it seems to be better. 

How could we make sure pymodaq_gui is called first ? in pymodaq init?

PR #999, opened 2026-04-06
Bugfix/custom app,  setup UI and toolbars management/visibility

seb5g:
This PR is a kind of big modification with: 

* the names of the methods used to produce custom apps and extension
* better management of  the toolbars associated with those customapps
* automatic creation of the View menu in the shared ui to show hide toolbars

The goal is to make those constructor methods clearer, they are now named as follow and `setup_ui` method is calling them in this order (setup_actions and setup_menus have been reversed to allow menus and toolbar creation before creating actions that could be added automatically to created menus and toolbars):

```python
* setup_docks_and_widgets()  # because we don't always use docks but also regular widgets
* setup_menus_and_toolbars(self.menubar)  # create menus and toolbars and add them to main_window
* setup_actions()  # create and add actions to various menus and toolbars
* connect_things()  # connect actions and other signals
* do_things_after_ui_setup()  # post setup ui stuff if needed
```

In order to have a clear and simple view of how it is built and what's the result, you can run the example:

*pymodaq_gui.examples.shared_ui.py*

There should still be backcompatibility with existing customapp


seb5g:
@bernhardlang as you just started with customapps and extensions do you reckon it make things clearer?

packages/pymodaq_gui/src/pymodaq_gui/examples/function_plotter.py
```diff
@@ -60,7 +62,7 @@ def __init__(self, dockarea):
 
         self.ind_plot = 0
 
-    def setup_docks(self):
+    def setup_docks_and_widgets(self):
```
Ashwolaa, line 75: comment in init is longer good

packages/pymodaq/src/pymodaq/extensions/scan/daq_scan.py
```diff
@@ -183,26 +185,7 @@ def __init__(self, dockarea: gutils.DockArea = None, dashboard: DashBoard = None
...
-        self.settings.child('plot_options', 'plot_0d').setValue(
-            dict(all_items=data0D_names, selected=data0D_names))
-        self.settings.child('plot_options', 'plot_1d').setValue(
-            dict(all_items=data1D_names, selected=data1D_names))
-
-
-    def setup_docks(self):
-        self.mainwindow.removeToolBar(self.toolbar)  # hides it
-
+    def do_things_after_ui_setup(self):
```
Ashwolaa: I guess this one is slightly confusing as DAQ_Scan is also a custom app similarly as DAQ_Scan_ui so the methods'name are repeated in both the ui and non-ui part.
Maybe this should be replaced by a method in the ui with the arg self.

self.finalize_ui(self)
or the do_things_after_ui_setup should accept some args

packages/pymodaq/src/pymodaq/dashboard.py
```diff
@@ -492,33 +508,35 @@ def setup_actions(self):
...
-            if file.suffix == ".xml":
-                self.add_action(
-                    self.get_action_from_file(file, ManagerEnums.remote),
-                    file.stem,
-                    "",
-                    auto_toolbar=False,
-                )
+        # self.add_action("save_roi", "Save ROIs as a file", "", auto_toolbar=False,
+        #                 menu='roi')
+        # self.add_action("modify_roi", "Modify ROI file", "", auto_toolbar=False,
```
Ashwolaa, line 509: Keeping all ROIs as comment for now to ease future implementation? Could be moved to a "junk" folder (not necessarily much prettier) but it has the advantage of keeping it stored as a block

packages/pymodaq_gui/src/pymodaq_gui/examples/shared_ui.py
```diff
@@ -0,0 +1,71 @@
+from qtpy import QtWidgets
```
Ashwolaa, line 1: Pretty cool example, I wonder if we could add a title over/below each menu to indicate more clearly the section it refers too. 
This could ease the navigation and one could have a checkbox to accept it or not

packages/pymodaq_gui/src/pymodaq_gui/utils/custom_app.py
```diff
@@ -161,30 +159,107 @@ def setup_docks(self):
...
+                           icon_name = '',
+                           auto_menu=True
+                           ) -> QtWidgets.QMenu:
+        if menu is None:
+            menu = self.menubar
+        return self.add_menu(self.__class__.__name__.lower(),
+                             self.__class__.__name__,
+                             menu,
+                             icon_name=icon_name,
+                             auto_menu=auto_menu)
```
Ashwolaa: a bit dry, I get the spirit but could use some docstrings. It is also not used throughout the code except here specifically. This seems a bit overkilled just to help people make their menu/toolbar. I would think it makes more sense to just add it as kwargs in the init of the CustomApp()

packages/pymodaq_gui/src/pymodaq_gui/utils/custom_app.py
```diff
@@ -161,30 +159,107 @@ def setup_docks(self):
...
+        """Get the default menu
+
+        Reimplemented from ActionManager
+        """
+        if self._menu is None:
+            if len(self.menus_names) == 0:
+                self._menu = self.create_object_toolbar()
+            else:
+                self._menu = self.menus[0]
+        return self._menu
```
Ashwolaa: I would stick to ActionManager. The main difference is the create_object_toolbar which could be implemented in the init if args are specified

packages/pymodaq_gui/src/pymodaq_gui/examples/function_plotter.py
seb5g: done

packages/pymodaq_gui/src/pymodaq_gui/utils/custom_app.py
```diff
@@ -161,30 +159,107 @@ def setup_docks(self):
...
+        >>>file_toolbar = self.add_toolbar('file_toolbar', 'File', self.mainwindow)
 
-        >>>file_menu.addSeparator()
-        >>>self.affect_to('quit', file_menu)
 
         See Also
         --------
         pymodaq.utils.managers.action_manager.ActionManager
+        """
+        self.setup_menu(menubar)
```
Ashwolaa: why keep setup_menu method here?

packages/pymodaq_gui/src/pymodaq_gui/utils/custom_app.py
```diff
@@ -161,30 +159,107 @@ def setup_docks(self):
...
+        return self._menu
+
+    def setup_menu(self, menubar: QtWidgets.QMenuBar = None):
+        """ Deprecated, use `setup_menus_and_toolbars`
+
         """
         pass
 
+    def setup_actions(self):
+        """Method where to create actions to be subclassed. Mandatory
```
Ashwolaa: Is it really mandatory for people to have it?

Ashwolaa: Docstring is off in CustomApp

    * setup_actions: mandatory, see :class:`pymodaq.utils.managers.action_manager.ActionManager`
    * value_changed: non mandatory, see :class:`pymodaq.utils.managers.parameter_manager.ParameterManager`
    * child_added: non mandatory, see :class:`pymodaq.utils.managers.parameter_manager.ParameterManager`
    * param_deleted: non mandatory, see :class:`pymodaq.utils.managers.parameter_manager.ParameterManager`
    * setup_docks: mandatory
    * setup_menu: non mandatory
    * connect_things: mandatory

Ashwolaa:
Looks nice overall, I am on the edge of using it directly in my Tpx3Cam GUI now!

packages/pymodaq/src/pymodaq/extensions/scan/daq_scan.py
seb5g: I moved all ui related things into the ui following your advice to pass self as argument.

packages/pymodaq/src/pymodaq/dashboard.py
seb5g: this is indeed a momentary action before I code the new roi_manager. I prefer to keep them as is to not forget them later... and storing them as a block is not that good as it will lack the methods that calls them...

packages/pymodaq_gui/src/pymodaq_gui/examples/shared_ui.py
seb5g: I'm not sure I understand what you mean here!!

packages/pymodaq_gui/src/pymodaq_gui/utils/custom_app.py
seb5g: did it using kwargs in init

packages/pymodaq_gui/src/pymodaq_gui/utils/custom_app.py
seb5g: done

packages/pymodaq_gui/src/pymodaq_gui/utils/custom_app.py
seb5g: for backcompatibility. if setup_menus_and_toolbars is not reimplemented, this one will be triggered and setup_menu will be called. Otherwise it would never be called  (because setup_ui is no more calling it) except if specifically called in the init

seb5g: I added the comment: for back-compatibility

packages/pymodaq_gui/src/pymodaq_gui/utils/custom_app.py
seb5g: not really no

seb5g: but I thought it would be the heart of an app with widgets and connections but you're right none of them are mandatory

seb5g: I cleaned the mandatory thing and added some comments on backcompatibility

seb5g:
alright these last modification should be ok then!

Ashwolaa:
I just discovered the toolbutton style mode with Toolbar that might be worth looking at for pymodaq and seems slightly relevant to this PR (but could be implemented afterwards though).
https://doc.qt.io/qt-6/qt.html#ToolButtonStyle-enum

This looks pretty neat for some new user that discover for the first time pymodaq to easily display the title of the actions. It makes it much clearer at what things does. This could be a checkable action list and could be added as a config_entry in pymodaq_gui. 

image

Ashwolaa (approved):

runrunm:
Toolbars are not efficient at displaying text.

Following https://github.com/PyMoDAQ/PyMoDAQ/issues/1002 if all toolbar elements are listed (with their name) in a dedicated `experiment_name` menu (or submenu) and additionally if all elements have a clear tooltip and a dedicated icon, text in the toolbar should not be needed.

PR #1001, opened 2026-04-07
Fix configurator executed multiple times on preset loading

daarklink:
Previously, the *defaul*t configurator was always loaded before the one provided to the *load_dashboard_with_preset* method in the Dashboard. This caused unnecessary data loading from a file and introduced a minor visual glitch, as the *default* configurator was still visible on screen after the UI loaded, before being replaced by the one passed to the method.

### DIFF
--- a/packages/pymodaq/src/pymodaq/dashboard.py
+++ b/packages/pymodaq/src/pymodaq/dashboard.py
@@ -1426,7 +1426,7 @@ def load_dashboard_with_preset(preset_name: str,
     -------
 
     """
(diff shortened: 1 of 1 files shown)


seb5g:
Why closing it?

daarklink:
Closing this PR as I forgot to push all changes to the fix branch. After pushing the remaining local changes, I realized the fix works on my *feature/launcher_extension* branch but not in isolation. I deleted the branch and recreated it, applying the fix from scratch, but it still does not work. I will investigate further later. Closing this PR to avoid merging the fix into *dev* as it breaks the dashboard.

PR #1003, opened 2026-04-08
Fix configurator executed multiple times on preset loading

daarklink:
Previously, the default configurator was always loaded before the one provided to the load_dashboard_with_preset method in the Dashboard. This caused unnecessary data loading from a file and introduced a minor visual glitch, as the default configurator was still visible on screen after the UI loaded, before being replaced by the one passed to the method.

### DIFF
--- a/packages/pymodaq/src/pymodaq/dashboard.py
+++ b/packages/pymodaq/src/pymodaq/dashboard.py
@@ -143,6 +143,7 @@ class DashBoard(CustomApp, LECOComponentMixin):
     """
     Main class initializing a DashBoard interface to display det and move modules and logger"""
     status_signal = Signal(str)
+    new_preset_created = Signal()
     config_changed = QtCore.Signal()
     # will be emitted when the user changed anything in the configuration files (emitted from SharedUI)
     # included in CustomExt by default but Dashboard is special with that respect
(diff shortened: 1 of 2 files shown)


malik-irain:
I don't get why so many stuff are changed, something like the following should be enough to solve the problem addressed:
 
```python
def load_dashboard_with_preset(preset_name: str,
                               extension_name: str = None,
                               configuration_name: str = 'default')  -> tuple[DashBoard, 'CustomExt', SharedUI]:
    """
    from pymodaq.utils.config import get_set_preset_path
    shared_ui, dashboard = create_load_dashboard()

    preset_path = get_set_preset_path().joinpath(f'{preset_name}.xml')
    preset_name = preset_path.stem
    extension = None

    if preset_name in dashboard.preset_manager.entries:
        dashboard.preset_manager.entry = preset_name
        if configuration_name is not None:
            dashboard.configurator.entry = configuration_name
        dashboard.preset_manager.execute_entry(preset_path)
        ...
    ...
  ```
 + the `self.configurator._execute_entry` that should be a `self.configurator.execute_entry` in Dashboard method `do_things_after_ui_setup`  
  The rest should be addressed with the opened issue #973

malik-irain:
It's simpler to use #1005 than to handle conflict and name changes here

PR #1005, opened 2026-04-09
Fix sync bug in + renamed manager standalone + restore '-p' for preset/exPeriment loading in cli

malik-irain:
Three fixes in this PR:

1. The way the configurator was updated when its associated experiment was changed made its toolbar take `default` entry instead of `self.entry` --> _fixed_
2. Configurator and Experiment managers app where named PresetManager --> _modified_
3. Restored `-p` for ex**P**eriment loading (in order to keep a one letter option). It could also be `-x` but it could be though as "e**x**tension". Or it could be `-e|--experiment` and `-x|--extension`


malik-irain:
@seb5g What's your opinion on the third point?

seb5g:
let's use -x for loading experiments

PR #1006, opened 2026-04-09
Feature/custom menu+persistent

Ashwolaa:
Little PR following #999 

- It changes the add_menu of ActionManager to accept custom menu instead of always creating a new one. In that case, the title of the given menu is taken as default. If a title is given, it replaces the menu's one
- Adding StickyMenu as a new menu type dedicated to keeping the menu open when checking/unchecking action. 
**sticky_menu_demo.py** lists the different use case.


Ashwolaa:
And by the way, it is now applied on the shared_ui.py toolbars

seb5g (commented):
I'm fine with this StickyMenu it's a good idea!

Also I'm ok with both the parent_menu and the menu (it behaves more like toolbar like this). I also wanted to do something similar but refrain so far as if one use the keyword menu in old code (or other customapps/extensions) it will behave badly. So please check that at least the extensions are fine in their setup_menus_and_toolbars methods. I think it will be the only likely place where add_menu has been used!

seb5g (approved):

PR #1007, opened 2026-04-10
Use a style proxy for checkboxes in dark themes

### LINKED ISSUE #892: Darktheme checkbox doesn't show properly
### Describe the bug

Using a dark theme produce checkboxes whose contour is not contrasted with respect to the background color:

Image

### Expected Behavior

We should clearly see the empty box

### Steps To Reproduce

Using the dev branch with a dark theme such as github-dark

### Reference(s)

```markdown

```

### Environment

```markdown
- OS: windows
- Pymodaq: dev branch
```

### Additional context

_No response_

malik-irain:
Fixes #892 
Applying CSS to a checkbox removes the box and the check mark. The check mark could be restored using an icon, but as the CSS only takes the path and can't modify an icon, it appeared black.

So I found that using a proxy and do the drawing inside was best way to have it without using an icon outside of MaterialIcons


seb5g:
I think this will do the job nicely! However we have to tweak a bit the drawing as it appears somewhat strange at the moment:

image

and maybe use a grey color and not white

seb5g:
and a bit more roundy!

malik-irain:
I don't have the issue, can you tell me if it is fixed now (I made the rectangle a little bit smaller).

For the rounder part, I can't really do more because it starts looking bad on Linux. Here is a radius of 2:
image

which is nice, but when I go higher with a radius of 3 it looks dis-formed: 
image
 
 Until it looks totally circular at 6

malik-irain:
Ok, I used material icons for both but it seems to be not rounded at all

malik-irain:
If you retry this, and it don't look good, can you try changing the size to 18:

```python
   def drawPrimitive(self, element:QtWidgets.QStyle.PrimitiveElement, option: QtWidgets.QStyleOption,
                      painter: QtGui.QPainter, widget: Union[QtWidgets.QWidget, None] = None):
        if element == QStyle.PrimitiveElement.PE_IndicatorCheckBox:
            size = QSize(20, 20) # <- Try with QSize(18, 18)
```

seb5g:
it is indeed better with size 18x18 ! I'm committing it and merging the PR!!

PR #1008, opened 2026-04-10
Fix configurator sync

### LINKED ISSUE #973: Bug : bad Configurator widgets synchronisation

malik-irain:
Fixes #973 

I don't know if its the intended way to make the sync widgets work but it solves the issue


Ashwolaa:
yes going with the sync is always recommended whenever possible

PR #1009, opened 2026-04-11
SharedUI improvements

runrunm:
This PR is for UX/UI improvements and clarifying code in shared_ui.py.

Changes:
- Actions order of apparition within menus
- Removal of `Restart` button from Runtime toolbar (still accessible from `File` menu) so that only `Quit` button is displayed in the toolbar (most used button).
- Set Runtime toolbar to unmovable (a user is not likely to move around the close button)

Some work on actions order of apparition within menus is needed, for example `Logs` and `Preferences` actions should always be at the bottom of the `Tools` menu and separated from the rest with a separator.


seb5g:
I did the modifications as I wanted to use your modifications in some other dev. I included the PR from @Ashwolaa on add_menu that should specify a parent_menu and an eventual custom menu (sticky)

PR #1011, opened 2026-04-12
Readability: SIM101

Ashwolaa:
Removing duplicate isinstance which can be simplified by checking against tuple.
https://docs.astral.sh/ruff/rules/duplicate-isinstance-call/

- applying SIM101
- simplifying _check_data_type

### DIFF
--- a/packages/pymodaq/src/pymodaq/utils/leco/pymodaq_listener.py
+++ b/packages/pymodaq/src/pymodaq/utils/leco/pymodaq_listener.py
@@ -478,7 +478,6 @@ class LECOComponentMixin:
             self._leco_commands_signal.connect(self._leco_client.queue_command)
             self._leco_client.start_listen()
         else:
-            self._leco_client.stop_listen()
             self._leco_commands_signal.emit(ThreadCommand(LECOCommands.QUIT, ))
             try:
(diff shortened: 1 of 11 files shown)


packages/pymodaq/src/pymodaq/utils/leco/pymodaq_listener.py
```diff
@@ -478,6 +478,7 @@ def connect_leco(self, connect: bool) -> None:
             self._leco_commands_signal.connect(self._leco_client.queue_command)
             self._leco_client.start_listen()
         else:
+            self._leco_client.stop_listen()
```
seb5g: this is unrelated?? but should probably be patched @malik-irain could you look at this

seb5g (changes requested):
overall good. Only one thing about leco that should be treated elsewhere

packages/pymodaq/src/pymodaq/utils/leco/pymodaq_listener.py
Ashwolaa: indeed my bad

Ashwolaa:
I also added a class method in h5saverlowlevel to create an instance directly from filepath

seb5g (approved):

PR #1012, opened 2026-04-12
Ux/configurator renaming in state manager

seb5g:
I renamed in this PR the Configurator to StateManager as it is managing state of the experiment setup. Configurator was a bit misleading wrt the preferences/configurations

Also removed the QLabel associated with the Managers and make them obvious by using a dedicated icon to open/describe  them:

image

However, I'm not sur the one I picked for the StateManager is the best... @runrunm any suggestions?

### DIFF
--- a/docs/src/api/api_utility_modules/api_managers.rst
+++ b/docs/src/api/api_utility_modules/api_managers.rst
@@ -77,12 +77,12 @@ API of the various managers, special classes easing the experimental orchestrati
 
 .. autosummary::
     experiment.experiment_manager::ExperimentManager
-    configurator.configurator::Configurator
+    state.state_manager::StateManager
 
 .. currentmodule:: pymodaq.utils.experiment.experiment_manager
 
 .. autoclass:: ExperimentManager
 
-.. py:currentmodule:: pymodaq.utils.managers.configurator.configurator
+.. py:currentmodule:: pymodaq.utils.managers.state.state_manager
 
-.. autoclass:: Configurator
+.. autoclass:: StateManager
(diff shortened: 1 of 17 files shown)


Ashwolaa:
I like the icons! not sure for the name though I think Configurator Manager carries more meaning but I see the issues with the config files. However, now that Preferences is used I think it is less problematic.
Good idea to remove the labels. I think we can add a togglable action to show the title of the actions for new comers or tutorial.

In my mind StateManager would actually refer to the current status of the plugins (initialized/grabbing/...)

runrunm:
Nice 👍 I was wondering how I could give a different icon to the different managers haha

I took note last time of using "Settings" or "Experiment settings" instead of "Configurator manager". Wouldn't it work better?
Using the settings wheel icon would then make sense. Also, we are not using "Settings" for PyMoDAQ (we are using "Preferences") so "Settings" could actually be dedicated to experiments by default (no need to specify "Experiment" in "Experiment settings").

Regarding the settings wheel icons in the dashboard modules I think the users can make the difference with the `Experiment settings` as it is clearly separated from the window toolbar and these modules settings are not available from the `Tools` menu. If you think it can bring confusion I think the "discover_tune" icon would work (the one actually used for the Configurator manager).

I have not used the configurator manager really much so far so maybe there is a good reason not to name it `Settings` or `Experiment settings`. For example does "manager" need to be in the name of the tool? Is it relevant to the user to know that `Experiment manager`, `Configurator manager`, `Overshoot manager` are actually managers ?

I am saying this always with the goal of making PyMoDAQ more user-friendly, "Settings" is by default understood by users, which is not the case of "Configurator" or "State" or even "Manager" :)

Ashwolaa:
how about the build/build_circle now that the others do not use it?
There are also construction and manufacturing that are not so bad too

seb5g:
My Problem with Settings is that is in fact only part of what the StateManager does. It is also used to set particular values of the actuators or to init them. Think about beamshutters used in a complex experiment of light absorption where you have one state of those to measure the withe light, another to measure the background and a third state to do the real measurement.  This is mostly why I called it StateManager as the name can be thought of for both the instrument settings and the instruments values/positions/ init state ... In the light of this, finally I guess discover_tune is not that bad (even if used at the moment in the actuator UI to display other options (relative moves, homing...)

Issue #1010, opened 2026-04-12
Python version


Ashwolaa:
<img width="859" height="539" alt="Image" src="https://github.com/user-attachments/assets/474089d2-69a1-4ec0-95c0-40863afa7686" />

We reached end support for Python<3.10.
Should we change the python version to only accept >=3.10 from now on?

Following this, I think we can adress more directly some of the PEPs requirements which sometimes rely on newer pythonic version.

seb5g:
I'm fine with this! What kind of PEP requirements ?

Ashwolaa:
I am going through some of the ruff annotations.

Among the interesting ones, I see these two:
- Getting rid of Union to use | operator  https://docs.astral.sh/ruff/rules/non-pep604-annotation-union/
- removing Tuple/Dict/... to use the direct type tuple/dict/... https://docs.astral.sh/ruff/rules/non-pep585-annotation/

seb5g:
yes ok. 
Beware that using | for Union is not working for class like 'DAQ_Move' where the '' are because DAQ_Move (or other classes) are imported only within if TYPE_CHECKING clause (to circonvient circular import or other things)

PR #1013, opened 2026-04-13
Updating python versioning

Ashwolaa:
Adressing  #1010
- Changing python base version to 3.10 and extending it to 3.13. This now matches also the github workflows' tests.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -31,11 +31,15 @@ Requirements
 --------
 
 Pymodaq aims at following `SPEC0`_.
+.. _SPEC0 :https://scientific-python.org/specs/spec-0000/
+
+
 In practice, that means we will only test python versions still considered `alive`_ and that have been bugfixed for at least a year. Other python versions may still work but we cannot guarantee it. 
 For now, this reflects into support in: Python 3.10+ - 3.13
-We usually recommend to use the python distribution that left or will soon leave bugfix status: python 3.12-3.13
-.. _SPEC0 :https://scientific-python.org/specs/spec-0000/
 .. _alive:https://devguide.python.org/versions/
+
+
+We usually recommend to use the python distribution that left or will soon leave bugfix status: python 3.12-3.13
 
 Published under the MIT FREE SOFTWARE LICENSE
 
--- a/README.rst
+++ b/README.rst
@@ -31,12 +31,12 @@ Requirements
 --------
 
 Pymodaq aims at following `SPEC0`_.
-.. _SPEC0 :https://scientific-python.org/specs/spec-0000/
+.. _SPEC0: https://scientific-python.org/specs/spec-0000/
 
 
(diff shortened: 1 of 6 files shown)


seb5g:
You said:

in the dependencies of pymodaq_utils, there is:
"strenum; python_version<"3.11"",

Good reasons to keep?

Yes because people may still use python 3.10 for a while. What's the recommended version in the doc at the moment? we should say 3.12!

Ashwolaa:
> You said:
> 
> in the dependencies of pymodaq_utils, there is: "strenum; python_version<"3.11"",
> 
> Good reasons to keep?
> 
> Yes because people may still use python 3.10 for a while. What's the recommended version in the doc at the moment? we should say 3.12!

I added some text in the README.rst with reference to SPEC0 and devguide. 
- According to SPEC0 (followed py pyqtgraph), the python's version we should care about are 3.11-3.14.
- According to devguide, the last "alive" version is 3.10 (which will read end of life at the end of 2026)
I can adapt the text accordingly. But maybe we should start adding tests for 3.14? or is it too premature?

Why do we need a tpl? I edited both but I am not quite sure if I did it the way it is intended.

seb5g (approved):

seb5g:
I guess tpl is generated during building of the docs by sphinx but is not included in gitignore (could you delete it and add it to gitignore?)

Ashwolaa:
Ok I found it in the github actions. I guess it is @malik-irain that cleaned up this.
.github\workflows\compile-readme-templates.yml

I'll try to only touch the .tpl then.

seb5g:
I see, indeed one should not remove it then... ;-)

malik-irain:
> > You said:
> > in the dependencies of pymodaq_utils, there is: "strenum; python_version<"3.11"",
> > Good reasons to keep?
> > Yes because people may still use python 3.10 for a while. What's the recommended version in the doc at the moment? we should say 3.12!
> 
> I added some text in the README.rst with reference to SPEC0 and devguide.
> 
>     * According to SPEC0 (followed py pyqtgraph), the python's version we should care about are 3.11-3.14.
> 
>     * According to devguide, the last "alive" version is 3.10 (which will read end of life at the end of 2026)
>       I can adapt the text accordingly. But maybe we should start adding tests for 3.14? or is it too premature?
> 
> 
> Why do we need a tpl? I edited both but I am not quite sure if I did it the way it is intended.

This specific tpl is not really needed, but it is useful for other READMEs that get compiled to automatically update tests badges with the branch number and potentially other git specific informations (but I see they fail ...)

malik-irain:
Also, the test matrix in github action should be updated to reflect that change !

seb5g:
@Ashwolaa could you update the matrix in github actions as suggested by @malik-irain ?

Ashwolaa:
For now I was thinking to stick to 3.10-3.13 or are you ok to go with 3.14?

PR #1018, opened 2026-04-13
Feature/python editor scripting

seb5g:
This PR adds an App to edit python code using the Monaco Editor from VSCode adapted to be used with Qt with integrated syntax highlighting and code completion!!

It is wrapped with file management to load, save, create new file and display them.

image

To be used next to run code, especially code to script data acquisition!!


Ashwolaa:
WoW! This was exactly what I was dreaming off and that I couldn't manage to do myself

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
```diff
@@ -0,0 +1,183 @@
...
+class MonacoApp(CustomApp):
+    def __init__(self, parent: QtWidgets.QWidget = None,):
+        super().__init__(parent, create_app_toolbar=False)
+
+        self.monaco_widget: Monaco = None
+        self._save_path = config('data', 'data_saving', 'h5file', 'save_path')
+
+        self._files: list[Path] = []
+        self._file_widgets: dict[Path, FileWidget] = OrderedDict([])
+
```
Ashwolaa: Why not use QTabWidget directly?

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
```diff
@@ -0,0 +1,183 @@
...
+    def save_file_as(self):
+
+        file_path = select_file(start_path=self._save_path, save=True, ext='py',
+                                force_save_extension=True)
+        if file_path:
+            with open(file_path, 'w', encoding="utf-8", newline='') as f:
+                f.write(self.monaco_widget.get_text())
+            self.remove_file(self._current_file)
+            self.add_file(file_path)
```
Ashwolaa: I would imagine more a current_file property

self.current_file = file_path #immediatly clean the text widget and try to load the file.

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
```diff
@@ -0,0 +1,183 @@
...
+        action_widget = FileWidgetAction(file_path)
+        self.add_action(str(file_path), str(file_path), toolbar='files', menu='files',
+                        action=action_widget)
+        widget = action_widget.defaultWidget()
+
+        self.filenames_group.addButton(widget.file_name_button)
+        widget.file_name_button.setChecked(True)
+        widget.file_name_button.toggled.connect(self._create_lambda_file_slot(file_path))
+        widget.close_button.clicked.connect(lambda: self.remove_file(file_path))
+        self._file_widgets[file_path] = widget
```
Ashwolaa: the QTabWidget does all of it for free with possible reordering between the tabs.

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
```diff
@@ -0,0 +1,183 @@
...
+        elif len(self._file_widgets) > 1:
+            self._current_file = list(self._file_widgets.keys())[1]
+        else:
+            self._current_file = None
+            self.monaco_widget.set_text('')
+
+        self._file_widgets.pop(file_path)
+        self._files.remove(file_path)
+        if self._current_file is not None:
+            self._file_widgets[self._current_file].file_name_button.setChecked(True)
```
Ashwolaa: By connecting the qtabWidget.currentChanged, you can immediatly obtain the next "key" and associate it to the self.current_file.

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
```diff
@@ -0,0 +1,183 @@
...
+        with tempfile.NamedTemporaryFile(mode='w', suffix='.py', delete=False) as f:
+            self.add_file(Path(f.name))
+
+    def load_file(self):
+        file_path = select_file(start_path=self._save_path, save=False, ext='py')
+        if file_path:
+            self.add_file(file_path)
+
+    def save_file_as(self):
+
```
Ashwolaa, line 119: there should be a distinction between 
save_file_as and a save_file when the file already exists ctrl + S only updates it, when it is not it opens a dialog.
The base_folder should not be the temp file too but maybe where the script runs as default
ctrl + shift + s for mandatory save_as

Ashwolaa:
https://github.com/user-attachments/assets/d2886dd4-983c-466c-8966-1d65468918c7

Not as good as what is promised in https://github.com/bec-project/qtmonaco/tree/main
Are there some packages to install?

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
seb5g: Well it does not do the collapsible things a toolbar does when you shrink the window. With a tabwidget the width of the window is set to the total width of the tab

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
seb5g: I want it to collapse the FileWidgets just like a toolbar can do if the window size is smaller than the size of the toolbar

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
seb5g: well no here if you rename the file, it is discarded (remove widget with name) and another widget is created and then file loaded in the editor

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
seb5g: reordering is possible it's a toolbar!

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
seb5g: same comment as above

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
seb5g: and actually the page area of the qtabwidget is useless to me as I have only on editor widget

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
seb5g: ok I'll buy that, i added a save file to update the saving

seb5g: should we autosave ? for instance if you do a modification then move from a tab to another and come back you'll lose your modifications...

seb5g: I implemented this in the display_file_text method just before updating the current file attribute

seb5g:
> Not as good as what is promised in https://github.com/bec-project/qtmonaco/tree/main Are there some packages to install?

I don't know, they say there is so far not much features enabled. Anyhow in other solution we would need to do it by hand. For me I wanted mainly an editor to load scripts and run them. Having syntax highlighting is already a good thing. And for development one can still rely on doing it in usual IDE

seb5g:
actually about the qtabwidget: it would be ok to use if one has an editor widget in each tab. I don't know why but I went with a single editor and update of it depending the selected file. Maybe it's not a good idea...

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
Ashwolaa: yes I was mostly talking about the QTabBar of the QTabWidget.

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
Ashwolaa: You can drag qactions from one place to another? With the QTabBar you have this behavior and you can circulate through them with standard IDE shortcuts. ctrl+tab / ctrl +shift+tab does the switch naturally.
But I guess it is really a matter of use. I just think using tab just make it much closer to what you would expect from a fake IDE. You also have access to much more direct signal. I do not really buy this collapsible argument. How short do you want it to be? You can always make it very narrow
alt text
alt text
vs 
alt text

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
```diff
@@ -0,0 +1,190 @@
...
+    @property
+    def local_files_path(self) -> Path:
+        return get_set_local_dir(user=True).joinpath('monaco_files')
+
+    def quit_fun(self):
+        super().quit_fun()
+        with open(self.local_files_path, 'wb') as f:
+            pickle.dump(self._files, f)
+
```
Ashwolaa, line 160: why not using human readable? like JSON?

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
Ashwolaa: It mostly depends how you build your property right?
You can alway add a save_current action if the file has a modified flag?

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
Ashwolaa: From this perspective, yes you have to autosave. I am not so confortable with this single widget in this regard. I don't know how heavy this it. One could also just keep the text in memory instead of reading it everytime.

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
Ashwolaa: image
vs 
image
from
image

Ashwolaa:
At the end, it is also a matter of preferences. In my mind, I would have start with a QTabWidget and an editor in each tab but I do not have much clue on the memory cost when you have 10 widgets opened. I did some of it with gigantic tables (1000*1000) and as long as you are not rendering the widget not going onto this tab it is ok. But I do not know much on this for qt-monaco. On my computer, the autocomplete is a bit slow (not very slow) but sufficiently to be noticeable.

I feel like QTabBar is the good compromise for it if you want to stick to single qt-monaco widget. 

I am wondering about the CustomApp too. It makes sense to benefit from the ActionManager, but I would have maybe decomposed first into a ScriptEditorWidget that manages the connection between files vs actions/tabs

seb5g:
you convinced me, here is an implementation using QtabWidget (much simpler indeed) and multiple editors (could not see slowness). When shrinking the window the Qtabbar shows lateral arraws, I would prefer a dropdown menu to display the tabs that are not on screen (like pycharm)

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
```diff
@@ -0,0 +1,177 @@
...
+        self._save_path = file_path.parent
+
+        monaco_widget = self.create_monaco_widget(file_path)
+        tab_index = self.tab_widget.addTab(monaco_widget, file_path.name)
+        self.tab_widget.setTabToolTip(tab_index, str(file_path))
+        self.tab_widget.setCurrentIndex(tab_index)
+
+        if display:
+            self.display_file_text(file_path, monaco_widget)
+
```
Ashwolaa, line 143: give focus to the newly opened tab

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
```diff
@@ -0,0 +1,177 @@
...
+    def close_editor(self, tab_index: int):
+        monaco_editor = self.current_editor
+        messagebox(text=f'Are you sure you want to close the file {monaco_editor.file_path.name}')
+        self.tab_widget.removeTab(tab_index)
+
+    def create_file(self):
+        with tempfile.NamedTemporaryFile(mode='w', suffix='.py', delete=False) as f:
+            self.add_file(Path(f.name))
+
```
Ashwolaa, line 109: I do not like this too much. In my use, when I create a new file I immediatly get a name tmpfile.py in my temp. 
When you want to save it, you start immediatly in the temp folder.
In practice, it would be better to just have a "Untitled" / "New" name that would ask you to save when leaving. I think using a flag could maybe do (like current_tab.is_untitled) or something in this line

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
```diff
@@ -0,0 +1,177 @@
...
+    monaco_app = MonacoApp(win)
+
+    shared_ui = SharedUI(win)
+    shared_ui.affect_application(monaco_app)
+
+    qapp.exec_()
+
+if __name__ == "__main__":
+    main()
```
Ashwolaa, line 177: self.tab_widget.setElideMode(Qt.TextElideMode.ElideRight)
        self.tab_widget.setMovable(True)

Ashwolaa, line 177: use dialog instead

Ashwolaa, line 177: Looking at notepad++
New / Open / Save / Save_as / Save a copy as / Close

Ashwolaa, line 177: why not json?

Ashwolaa, line 177: logger?

Ashwolaa, line 177: no need for enumerate if unused

Ashwolaa, line 177: You can check QFileSystemWatcher, this monitor if the files have been changed. So the display of the tab could be updated to show that.
- vscode shows a filled circle
- np++ changes the color of the text
- I have seen with * also

One could also store the self._value of monaco when a file is opened/saved. A direct comparison between the actual value and the stored one would do

Ashwolaa: I don't know how far you want to go with this tool but if it runs some python scripts. One could add an action to test compilation
```
def is_python_syntax_valid(file_path):
    try:
        with open(file_path, "r", encoding="utf-8") as f:
            compile(f.read(), file_path, "exec")
        return True
    except SyntaxError as e:
        print(f"Syntax error: {e}")
        return False
    except Exception as e:
        print(f"Error: {e}")
        return False
```

Ashwolaa: My PR tool got fucked up it seems. This was in 
`connect_actions`

Ashwolaa: This was in
`        messagebox(text=f'Are you sure you want to close the file {monaco_editor.file_path.name}')
`

Ashwolaa:
```
        with open(self.local_files_path, 'wb') as f:
            pickle.dump(self.files, f)
```

Ashwolaa: `                for ind, file in enumerate(_files):
`

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
seb5g: it has focus with:

self.tab_widget.setCurrentIndex(tab_index) 

in add_file,

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
seb5g: done

seb5g: yeah of course!

seb5g: I modified save_as with save a copy as

ad for compilation test, I'll actually add it into the pymodaq extension I will buid on top of the editor dedicated to script execution. I'll include your method there!

seb5g: no need for and pickle handle directly Path object

seb5g: ??

seb5g: yep

Ashwolaa: yeah but human readable is best!

Ashwolaa:
```
        try:
            with open(self.local_files_path, 'rb') as f:
                _files = pickle.load(f)
                for ind, file in enumerate(_files):
                    self.add_file(file, display=True)
        except FileNotFoundError as e:
            pass
```
            
            It was for the except

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
Ashwolaa: I meant the monaco widget inside

seb5g:
@Ashwolaa The PR is done with all your suggestions and others. I included the syntax check and the file execution using subprocess. I'm printing below the editor some info on the subprocess execution. It seems to work nicely!!

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
```diff
@@ -0,0 +1,393 @@
...
+from pymodaq_gui.messenger import dialog
+from pymodaq_utils.config import get_set_local_dir, GlobalConfig
+
+
+
+
+from pymodaq_gui.utils import CustomApp
+from pymodaq_gui.utils.shared_ui import SharedUI, MenuNames
+from pymodaq_gui.utils.file_io import select_file
+
```
Ashwolaa, line 21: - remove unused imports
- format import, I can do that 
- MenuNames should be MenuToolbarNames

packages/pymodaq_gui/pyproject.toml
```diff
@@ -56,6 +56,11 @@ dev = [
     "qt_material_icons>=0.5.0"
 ]
```
Ashwolaa, line 65: shouldn't qtmonaco be  in here too?

Ashwolaa:
Did you do all the development with pyside6?

I have these issues when running it with pyqt5/6

  File "C:\Users\Constant\miniconda3\envs\pymodaq_3.13\Lib\site-packages\qtpy\QtWebEngineWidgets.py", line 28, in <module>
    from PyQt5.QtWebEngineWidgets import (
    ...<5 lines>...
    )
ModuleNotFoundError: No module named 'PyQt5.QtWebEngineWidgets'

The above exception was the direct cause of the following exception:

  File "d:\Work\PyMoDAQ\packages\pymodaq_gui\src\pymodaq_gui\editor\monaco.py", line 10, in <module>
    from qtmonaco import Monaco
  File "C:\Users\Constant\miniconda3\envs\pymodaq_3.13\Lib\site-packages\qtmonaco\__init__.py", line 1, in <module>
    from .monaco import Monaco
  File "C:\Users\Constant\miniconda3\envs\pymodaq_3.13\Lib\site-packages\qtmonaco\monaco.py", line 6, in <module>
    from qtpy.QtWebEngineWidgets import QWebEngineView
  File "C:\Users\Constant\miniconda3\envs\pymodaq_3.13\Lib\site-packages\qtpy\QtWebEngineWidgets.py", line 36, in <module>
    raise QtModuleNotInstalledError(
    ...<2 lines>...
    ) from error
qtpy.QtModuleNotInstalledError: The QtWebEngineWidgets module was not found. It must be installed separately as PyQtWebEngine.

seb5g:
I did it only with pyside6. I think QtWebengine may be in a separate package. In pyside6 it is in the addons but addons are installed by default with pyside6.

See this on pypi:
image

packages/pymodaq_gui/pyproject.toml
seb5g: it's been added in an optional dependency you'll have to install pymodaq_gui[editor]

don't know if it's a good idea but I feel we may have some trouble of dependencies with this editor thing

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
seb5g: done

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
```diff
@@ -0,0 +1,388 @@
...
+            self.tab_widget.removeTab(tab_index)
+
+    def remove_file(self, file_path: Path):
+        """ remove the editor corresponding to the specified file Path """
+        tab_index = self.files_path.index(file_path)
+        self.close_editor(tab_index)
+
+    def create_file(self):
+        """ Create a new file and display it in a new tab """
+        current_directory = self.current_editor.file_path.parent
```
Ashwolaa: current_editor has to be initialized if it has not been opened on start up.  Adda dummy tab if there is none in memory?

  File "D:\Work\PyMoDAQ\packages\pymodaq_gui\src\pymodaq_gui\editor\monaco.py", line 278, in create_file
    current_directory = self.current_editor.file_path.parent
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'file_path'

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
Ashwolaa: Weitrd syntax, I get an error with it
  File "D:\Work\PyMoDAQ\packages\pymodaq_gui\src\pymodaq_gui\editor\monaco.py", line 280, in create_file
    with current_directory.joinpath('untitled.py') as f:
         ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
TypeError: 'WindowsPath' object does not support the context manager protocol

give the path directly instead?
 self.add_file(current_directory.joinpath('untitled.py'))

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
```diff
@@ -0,0 +1,388 @@
...
+        if file_path:
+            self.add_file(file_path)
+
+    def save_file(self):
+        """ Save the current editor content in the corresponding file Path """
+        self._save_file(self.current_file)
+
+    def save_copy_file_as(self, file_path: Path = None):
+        """ Make a copy of the curent editor content in a new file Path """
+        if file_path is None:
```
Ashwolaa, line 351: should be False, the signal is sending a boolean. It never goes through the selection

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
Ashwolaa: I would also add a counter to it, untitled1.py /n new1.py 
I think we have some utility functions to get nearest integer

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
```diff
@@ -0,0 +1,388 @@
...
+        proc = subprocess.Popen([sys.executable, str(file_path)], stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE, text=True)
+        out, err = proc.communicate()
+        if err:
+            self.redirect.write('Error in subprocess:')
+            self.redirect.write(err)
+        if out:
+            self.redirect.write('Output of subprocess:')
+            self.redirect.write(out)
+
```
Ashwolaa, line 121: Could you add some tests for this?
Testing with random files is always stuck at Starting subprocess
Worth adding a completion message too?

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
```diff
@@ -0,0 +1,388 @@
...
+        self.connect_action('load', lambda: self.load_file())
+        self.connect_action('new', self.create_file)
+        self.connect_action('save', self.save_file)
+        self.connect_action('save_copy_as', self.save_copy_file_as)
+
+        self.connect_action('run_file', lambda: self.run_file())
+
+        self.tab_widget.tabCloseRequested.connect(self.close_editor)
+        self.tab_widget.tabBarClicked.connect(self.highlight_tab)
+
```
Ashwolaa, line 249: should be with currentChanged too

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
seb5g: create an untitled file if no files in memory

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
seb5g: indeed I know send a signal with no argument using a lambda. I'm also implementing the save_as

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
seb5g: image

when expecting error and ouput it's showing it, but when it's working and doing what is expected I don't know what could be added

seb5g: and the ouput and error is only printed when the subprocess is ended. So if your start a qtloop it will not print anything. Do you have something else in mind?

packages/pymodaq_gui/src/pymodaq_gui/editor/monaco.py
seb5g: isn't tabBarClicked enough? or do you mean if someone change it programmatically? I added it

Ashwolaa: I am usually navigating with keyboard shortcut not detected with tabBarClicked

seb5g:
@malik-irain all the tests but on linux pass. Could you try to run them locally on your computer then check why it's not working on the github machines (something related to usage of gpu on chromium in the QtWebEgine...),

malik-irain:
Does it work?

malik-irain:
Ok it seems the issue now comes from pyqt5 not having `QWebEnginePage` in `QtWebEningeCore` even through qtpy.

PR #1019, opened 2026-04-14
Readability/spacing and trailing commas

Ashwolaa:
Second attemp with the Ruff formater.
This PR aims at cleaning up files with respect to comma styles and space issues
The list of checks and fixes that have been applied is the following.
COM
E202
E211
E221
E222
E241
E251
E261
E271
E272

https://docs.astral.sh/ruff/linter/#rule-selection


malik-irain:
For most of it I agree with the uniformization of syntax even though I don't like final trailing commas

Ashwolaa:
It is a bit confusing but I noticed there was a distinction between:
```
my_dict = {a:5,
           b:7,
           c:8}
```

```
my_dict = {a:5,
           b:7,
           c:8,
}
```

In the first case, no comma. However if the closing bracket is on the next line, it will add it.

seb5g:
@Ashwolaa could you resolve those conflicts before we merge your pr?

Ashwolaa:
should be good

PR #1028, opened 2026-04-24
Feature/scan manager

seb5g:
This PR introduces two things concerning the DAQ_Scan extension.

* A Scan Manager to quickly configure, save and recall a file describing everything needed for a scan, given a certain preset! This is a way to preconfigure various scans and start them later on, for instance in the future Sequencer...

* A modification of the ScannerFactory object interface to allow more customization:
  * one can separate the positions variable  (historically as floats in a ndarray) from the DataActuator actually sent to the DAQ_Move. This is particularly important to still plot live things in a scan while using DataActuator with higher dimensionality (for instance in my under developpement 2D beamShaping extension). As a example, I'm performing a holography SLM calibration applying in half the SLM a 0 grey value and in the other half a grey value scanned from 0 to 256. I'm therfore able to plot things as a function of the grey scale (bundled in the positions variable) while actually sending 2D DataActuator to the SLM
  * More intersting is a processing of the raw data similarly to the DataMixer but integrated in the scanner factory. Usually, one select the data to be plotted live from the settings. But all data can now be processed within a reimplement custom ScannerBase object (part of the scanner factory) to produce processed data. One can benefit from the ScannerBase params to tailor the processing. This is a bit less convenient than the DataMixer but is much more integrated within the DAQ_Scan


seb5g:
@bernhardlang @rgeneaux you may want to have a look at the last point in the description that could heavily interest you. I guess the subject of custom ScannerBase classes could also be the subject of a tutorial...

seb5g:
Here is how a custom ScannerBase for the calibration looks like:

```python
@ScannerFactory.register()
class Scan1DCalibration(Scan1DBase):
    """ Defines a linear scan between start and stop values with steps of length defined in the step setting"""

    scan_subtype = 'BeamShapingCalibration'
    do_process_data = True

    params = [
        {'title': 'Grey Start:', 'name': 'start', 'type': 'float', 'value': 0},
        {'title': 'Grey Stop:', 'name': 'stop', 'type': 'float', 'value': 255},
        {'title': 'Grey Step:', 'name': 'step', 'type': 'float', 'value': 1}
        ]

    n_axes = 1
    distribution = DataDistribution.uniform

    def __init__(self, actuators: List['DAQ_Move'] = None, display_units=True, **_ignored):
        super().__init__(actuators=actuators, display_units=display_units)

    def process_data(self, dte: DataToExport) -> DataToExport:
        """ Process Acquired data if the boolean class attribute *do_process_data* is set to True

        Here we suppose the dte contains a 1D Horizontal ROI presenting fringes
        """
        ind_lineout = None
        for ind, name in enumerate(dte.get_names()):
            if 'Hlineout' in name:
                ind_lineout = ind
                break


        if ind_lineout is not None:
            dwa_camera_1d = dte[ind_lineout]
        else:
            dwa_camera_1d = dte.get_data_from_dim(DataDim.Data2D)[0].sum(0)
        dwa_camera_1d.create_missing_axes()
        dwa_ft = dwa_camera_1d.ft(axis_units='')
        index_max = np.argmax(np.abs(dwa_ft.isig[dwa_ft.size//2+10::][0]))

        phase = np.angle(dwa_ft.isig[index_max + dwa_ft.size//2+10][0][0])

        dte = DataToExport('Process')
        dte.append(dwa_camera_1d)
        dte.append(DataCalculated('Phase', data=[np.atleast_1d(phase)], labels=['Phase'], origin='Scanner'))
        return dte

    def data_actuator_at(self, scan_index: int, axis_index=0) -> DataActuator:
        """ Get the DataActuator specified with two indexes (will be used to be sent to the actuators)

        This allows to have a difference between self.positions (simple numbers used to describe the scan, usually the
        actuator values) and the real DataActuator sent to the Actuator (that could be 0D, 1D, 2D Data, see BeamShaping
        plugin)

        To be reimplemented if needed

        """
        shaper_shape = get_slm_size()
        data = np.zeros(shaper_shape)
        data[:, shaper_shape[1] // 2:] = self.positions[scan_index, axis_index]

        return DataActuator(self.actuators[0].title, data=[data],
                            units = self.actuators[0].units)

    @property
    def grey_scale(self) -> np.ndarray:
        return mutils.linspace_step(self.settings['start'], self.settings['stop'],
                                          self.settings['step'])

    def set_scan(self):
        self.get_info_from_positions(self.grey_scale)

        self.n_axes = 1

        self.n_steps = len(self.positions)

    def evaluate_steps(self) -> int:
        n_steps = int(np.abs((self.settings['stop'] - self.settings['start']) / self.settings['step']) + 1)
        return n_steps

    def get_nav_axes(self) -> List[Axis]:

        return [Axis(label=f'{self.actuators[0].title}',
                     units=f'{self.actuators[0].units}',
                     data=self.grey_scale)]

    def get_scan_shape(self) -> Tuple[int]:
        return len(self.positions),

    def get_indexes_from_scan_index(self, scan_index: int) -> Tuple[int]:
        """To be reimplemented. Calculations of indexes within the scan"""
        return (scan_index,)

```

bernhardlang:
Who is calling process_data? In case that's possible, wouldn't it be sensible to pass the current scan position or index as argument to this method? Or is that anyway accessible via some property of ScanBase / Scan1DBase?

seb5g:
> Who is calling process_data? In case that's possible, wouldn't it be sensible to pass the current scan position or index as argument to this method? Or is that anyway accessible via some property of ScanBase / Scan1DBase?

The DAQ_Scan extension is, whenever it grabbed raw data, it saves them, then check if they should be processed in the Scanner implementation and then export them for plotting/saving (if options are correct).

I just added this scan index feature in https://github.com/PyMoDAQ/PyMoDAQ/pull/1034 

you'll have in your ScanBase implementation a property called *current_scan_index* reflecting it

PR #1029, opened 2026-04-25
Hotfix/config+backend

Ashwolaa:
### Problem

When PyMoDAQ started with a fallback Qt backend (e.g. falling back from pyqt6 to pyside6 because pyqt6 was not installed), it saved only the currently available backends to the user config. On subsequent runs, the user config overrode the full system list, making newly installed backends invisible to PyMoDAQ.                                                                                                                                                                 

###Changes

pymodaq_gui/__init__.py — set_and_check_qt_backend_or_die: on fallback, reorder the full backend list (working backend first) instead of pruning it to only installed entries. The saved config now always retains all known backends.
                                                                                                                                                                   parameter/pymodaq_ptypes/list.py — ListParameterItem gains an unavailable opt. Items listed in unavailable are shown greyed-out and non-selectable in the combo   box, with a "Not installed" tooltip. The state is re-applied on every limitsChanged (repopulation) and optsChanged.

utils/widgets/tree_toml.py — TreeFromToml._post_process_params: after the parameter tree is built, walks it recursively alongside each config's template. For every list param named backend, expands the displayed limits to the full template list (so all known backends are always visible) and marks non-installed ones as unavailable. Covers Qt, HDF5 and plotting backends with no special-casing.                                                                                  
  
examples/parameter_ex.py — adds a demo entry in the Lists section showing the unavailable opt in action.                                                         

image

### DIFF
--- a/packages/pymodaq_gui/src/pymodaq_gui/utils/widgets/tree_toml.py
+++ b/packages/pymodaq_gui/src/pymodaq_gui/utils/widgets/tree_toml.py
@@ -4,6 +4,7 @@ Created the 19/10/2023
 
 @author: Sebastien Weber
 """
+import contextlib
 import sys
 from typing import Union, Any
 import datetime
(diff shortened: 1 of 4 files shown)


seb5g (approved):

seb5g:
seems quite nice, does it work ok when the value is a dict?

seb5g (changes requested):
there is actually an issue on the plotting backends, qt is available obviously (at least one) but is greyed in the plotting options??

packages/pymodaq_gui/src/pymodaq_gui/utils/widgets/tree_toml.py
```diff
@@ -41,6 +42,58 @@ def __init__(self, config: Config = None, capitalize=True, start_path: Union[str
 
         self._cached_config_changes = {}
         self.dialog = None
+        self._post_process_params()
+
+    def _post_process_params(self):
+        """For every list param named 'backend': expand limits from the template (so the full
```
seb5g, line 48: all backends are not directly related to a package... for instance the plotting backends are matplotlib and qt. qt is for any Qt python backend but is not a package. if we call it qtpy then it would work again but this is strange (but easy fix)...

seb5g (changes requested):

packages/pymodaq_gui/src/pymodaq_gui/utils/widgets/tree_toml.py
Ashwolaa: Ah yes, I was not carefulI on it. I added a mapping for these cases where backend differ from packages. Otherwise, I can change the config to say qtpy

Ashwolaa:
> seems quite nice, does it work ok when the value is a dict?

The unavailable compares directly from the combo's itemText so the setting up is the same as before.
For now, it only accepts list arguments in it. Is that your question?

packages/pymodaq_gui/src/pymodaq_gui/utils/widgets/tree_toml.py
seb5g: that's fine like this, one will have to recall this is here if other similar cases happen

seb5g (approved):

seb5g:
> > seems quite nice, does it work ok when the value is a dict?
> 
> The unavailable compares directly from the combo's itemText so the setting up is the same as before. For now, it only accepts list arguments in it. Is that your question?

No my question is related to the fact that a list parameter can accept both a list of str or a dict as limits. But as you say, because you get itemText from the combo, it should work whatever the value of the Parameter object

PR #1032, opened 2026-04-27
hotfix/install_python_3.14

Ashwolaa:
# Context:
Trying to install pymodaq_gui with pip install -e . with python 3.14 leads to a timeout error.

The root cause seems to be 
`run_git("tag", "--merged", "HEAD").splitlines()`
which runs for every `get_tag_containing` call.

# In this PR:
- adding merged as new arg in get_tag_containing to avoid recomputing it for each commit.
- updating the run_git to have timeout argument

### DIFF
--- a/git_describe.py
+++ b/git_describe.py
@@ -35,14 +35,16 @@ PRE_RANK_ORDER = { # Pre-release order mapping
 }
 
 
-def run_git(*args) -> str:
+def run_git(*args, timeout: int = 30) -> str:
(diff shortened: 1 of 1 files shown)


malik-irain:
I couldn't reproduce the issue, but you can just add the timeout and nothing else, it will be caught and print `FALLBACK_VERSION`.

You don't need for precomputed `merged` as if it fails, `merged` will be empty, so there won't be any  the candidates tag will also be empty, leading to a fallback to `FALLBACK_VERSION`.

PR #1036, opened 2026-04-28
Patch/data filtering and extra data in daq scan

seb5g:
This PR fix a few things in the DAQ_SCAN

* make sure data are saved before moving to next step
* properly filter data to be saved depending if the DWA has a do_save extra argument, if it is True and the H5saver settings are taken into account
* one can add some extra data to be saved into the h5file (on top of the one from the DAQ_Viewers). This is used here to save the result of the processed data from the scanner custom process_data method.

### DIFF
--- a/packages/pymodaq/src/pymodaq/utils/h5modules/module_saving.py
+++ b/packages/pymodaq/src/pymodaq/utils/h5modules/module_saving.py
@@ -12,12 +12,14 @@ import xml.etree.ElementTree as ET
 
 
 import numpy as np
+
+from pymodaq_data import DataDim, DataWithAxes
 from pymodaq_utils.logger import set_logger, get_module_name
 from pymodaq_utils.enums import BaseEnum
 from pymodaq_utils.abstract import ABCMeta, abstract_attribute, abstractmethod
 from pymodaq_utils.utils import capitalize
-from pymodaq_data.data import Axis, DataToExport, DataDistribution, DataRaw
-from pymodaq_data.h5modules.saving import H5SaverLowLevel
+from pymodaq_data.data import Axis, DataToExport, DataDistribution, DataRaw, DataSource
+from pymodaq_gui.h5modules.saving import H5SaverBase
 from pymodaq_data.h5modules.backends import GROUP, Node
 from pymodaq_data.h5modules.data_saving import (
     DataToExportSaver, DataToExportEnlargeableSaver,
@@ -45,7 +47,7 @@ class ModuleSaver(metaclass=ABCMeta):
     """Abstract base class to save info and data from main modules (DAQScan, DAQViewer, DAQMove, ...)"""
     group_type: GroupModuleType = abstract_attribute()
     _module = abstract_attribute()
-    _h5saver: H5SaverLowLevel = abstract_attribute()
+    _h5saver: H5SaverBase = abstract_attribute()
     _module_group: GROUP = abstract_attribute()
     main_module = True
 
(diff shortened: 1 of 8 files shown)


packages/pymodaq/src/pymodaq/utils/h5modules/module_saving.py
```diff
@@ -12,12 +12,14 @@
...
 from pymodaq_utils.enums import BaseEnum
 from pymodaq_utils.abstract import ABCMeta, abstract_attribute, abstractmethod
 from pymodaq_utils.utils import capitalize
-from pymodaq_data.data import Axis, DataToExport, DataDistribution, DataRaw
-from pymodaq_data.h5modules.saving import H5SaverLowLevel
+from pymodaq_data.data import Axis, DataToExport, DataDistribution, DataRaw, DataSource
+from pymodaq_gui.h5modules.saving import H5SaverBase
 from pymodaq_data.h5modules.backends import GROUP, Node
 from pymodaq_data.h5modules.data_saving import (
     DataToExportSaver, DataToExportEnlargeableSaver,
```
Ashwolaa, line 23: keep ordering either by library or by name
utils / data / gui
data / utils / gui

packages/pymodaq/src/pymodaq/utils/h5modules/module_saving.py
```diff
@@ -189,6 +191,33 @@ def _add_module(self, where: Union[Node, str] = None, metadata={}) -> Node:
...
+        """
+        dte_filtered = DataToExport('Filtered')
+        for dwa in dte:
+            flag = True
+            if 'do_save' in dwa.extra_attributes:  # this is the main filter
+                flag = flag and dwa.do_save
+            else:
+                flag = flag and self.filter_data_wrt_settings(dwa)
+            if flag:
+                dte_filtered.append(dwa)
```
Ashwolaa: No need for the flag variable here. It is always compared to True, something like this would do

    if 'do_save' in dwa.extra_attributes:
        if dwa.do_save:
            dte_filtered.append(dwa)
    elif self.filter_data_wrt_settings(dwa):
        dte_filtered.append(dwa)

packages/pymodaq/src/pymodaq/utils/h5modules/module_saving.py
```diff
@@ -189,6 +191,33 @@ def _add_module(self, where: Union[Node, str] = None, metadata={}) -> Node:
...
+        return dte_filtered
+
+    def filter_data_wrt_settings(self, dwa: DataWithAxes) -> bool:
+        """ Check if this DataWithAxes should be saved depending on the H5Saver settings
+
+        Return True if it should be saved
+        """
+        flag = True
+        if not self.h5saver.settings['save_2D']:  # exclude 2D data and above
+            flag = flag and not (dwa.dim == DataDim.Data2D or dwa.dim == DataDim.DataND)
```
Ashwolaa, line 215: Ok for now, but I cannot remember the interest to have this option? What is the point of it?

packages/pymodaq/src/pymodaq/utils/h5modules/module_saving.py
seb5g: utils / data / gui !!
done

packages/pymodaq/src/pymodaq/utils/h5modules/module_saving.py
seb5g: done

packages/pymodaq/src/pymodaq/utils/h5modules/module_saving.py
seb5g: to limit the size of the H5file. Because if you save a lot of images you may result with something so huge... So you can just check this when doing basic stuff and save them when performing important scans...

PR #1037, opened 2026-04-29
Feature/icon animations

Ashwolaa:
Little try to add animations to some icons in pymodaq. The functions are based on the PR#297 done on [qtawesome](https://github.com/spyder-ide/qtawesome/pull/297)

Unfortunately, it is a bit trickier than anticipated to apply as I do not see how to properly make use of the CharIconPainter in our current implementation with MaterialIcons (or I need to figure how). So for now, it looks like this:

https://github.com/user-attachments/assets/341ebd78-66b2-47a8-969a-149adf260d52

### DIFF
--- a/packages/pymodaq_gui/src/pymodaq_gui/examples/animated_actions.py
+++ b/packages/pymodaq_gui/src/pymodaq_gui/examples/animated_actions.py
@@ -1,0 +1,210 @@
+"""
+Animation Demo — PyMoDAQ.
+
+Three-tab showcase of AnimatedIconEngine:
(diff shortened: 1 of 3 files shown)


seb5g:
what do you think of this as a potential use case?

Ashwolaa:
I am thinking as usecase mostly to indicate either problems on the hardware or software processing (scan in progress / processing_data / loading preset / status).  
For example, I like the blinking pause button to highlight interrup in the scan / slow spin when loading presets / and maybe a heartbeat on the status of the dashboard / Travel X during a scan progress

PR #1039, opened 2026-05-06
Automatically switch configs when incompatible

malik-irain:
This PR introduces:
- Automatic config change when an incompatible change is made
- User warnings about it
- A best effort to keep custom config values (if they still exist in the template)


seb5g:
I tried a few things but somehow few things seems to work?
First the user config folder renaming ask for permission?? Is it because it is open?
Then I don't have messages appearing.

How do you propose to test this. Should I change:
* a config entry value? Type?

malik-irain:
> First the user config folder renaming ask for permission?? Is it because it is open?

A little bit strange, if the system folder is already created it shouldn't ask for permission to create a new one


>  How do you propose to test this. Should I change:
>
>    a config entry value? Type?

What I generaly did was changing `n_iter` type in pymodaq's config_template

malik-irain:
@seb5g should work now, but there is an "issue" when removing entries from template, it detects it as incompatible config

seb5g:
@malik-irain in the actual dev branch, the plugin configuration are not registered automatically in the GlobalConfig object. Was it not the case before? We should incorporate the registration mechanism within the init of the BaseConfig method so that all configs that are imported will be automatically registered!

malik-irain:
Found out why, that's because we don't want Cache (that are configs object) to be registered (and it doesn't work)

seb5g:
il ya  des erreurs au niveau des tests de utils. Il ne trouve pas une fixture!

malik-irain:
should be all good now

PR #1041, opened 2026-05-06
Standardization of a quit mechanism

malik-irain:


packages/pymodaq/src/pymodaq/extensions/custom_ext.py
```diff
@@ -112,11 +112,12 @@ def modules_manager(self) -> ModulesManager:
         return self._modules_manager
 
     def exit_runner_thread(self, duration : int = 5000):
```
seb5g: do you know why we have this method? Is it an attempt in facilitating the multithreading? Should we add similarly a runner and runner_thread creation method and use it in our extensions?

Something like:

```python

def create_runner(self, runner_obj: Any):
    #stuff to create runner_thread

def connect_runner(self):
    # signal/slots connection

def start_runner_thread(self):
    ...

def exit_runner_thread(self):
    ...
```  

maybe in a Mixin?

packages/pymodaq/src/pymodaq/extensions/custom_ext.py
```diff
@@ -53,10 +53,10 @@ def stop(self):
         """
         raise NotImplementedError
 
-    def quit_fun(self):
```
seb5g, line 56: add a quit_fun method for backcompatibility assing a deprecation message and calling quit

packages/pymodaq/src/pymodaq/utils/shared_ui.py
```diff
@@ -54,8 +54,8 @@ def start_plugin_manager(self):
         widget = QtWidgets.QWidget()
         self.win_plug_manager.setCentralWidget(widget)
         self.plugin_manager = PluginManager(widget)
-        self.plugin_manager.quit_signal.connect(self.quit_fun)
-        self.plugin_manager.restart_signal.connect(self.restart_fun)
```
seb5g, line 58: same create restar_fun with deprecation message for backcompat

packages/pymodaq/src/pymodaq/dashboard.py
```diff
@@ -476,12 +476,6 @@ def setup_menus_and_toolbars(self, menubar: QtWidgets.QMenuBar = None):
         self.add_menu('overshooter', 'Overshooter', MenuToolbarNames.TOOLS, icon_name=Overshooter.icon_name)
         self.get_menu('overshooter').setEnabled(False)
 
-        # self.roi_menu = self.add_menu('roi', 'ROI', auto_menu=False)
```
seb5g, line 476: please leave the comments so far until I finished with Manager creation: ROI and remote, then I'll remove the comments

packages/pymodaq_gui/src/pymodaq_gui/utils/shared_ui.py
```diff
@@ -117,17 +130,22 @@ def __init__(self, widget: Union[QtWidgets.QWidget, DockArea],
...
 
         self.setup_ui()
+
+        self._close_filter = CloseEventFilter(self.quit)
+        self.mainwindow.installEventFilter(self._close_filter)
+
         self.mainwindow.setVisible(show)
 
-    def affect_application(self, app: CustomApp):
+    def affect_application(self, app: QuittableApp):
```
seb5g, line 142: if app is not a CustomApp the following lines won't work as a QuittableApp has no menus attribute and toolbars attributes, title and so on... 
In fact I'm wondering why we have `if isinstance(app, CustomApp):` on line153

packages/pymodaq_gui/src/pymodaq_gui/utils/shared_ui.py
```diff
@@ -236,33 +254,29 @@ def setup_actions(self):
...
+        quit
         """
         try:
-            if hasattr(self._main_application, 'quit_fun'):
-                self._main_application.quit_fun()
-                QtWidgets.QApplication.processEvents()
-
-            if hasattr(self, "mainwindow"):
-                self.mainwindow.close()
+            self._main_application.quit()
```
seb5g, line 273: should call the super?

seb5g (changes requested):
Good refactorisation, one could go beyond with the threading stuff (or maybe in another PR)

Some small chenges to be done

packages/pymodaq/src/pymodaq/extensions/custom_ext.py
malik-irain: QThreads should be subclassed in PyMoDAQ to include this termination method if possible

malik-irain: For the separation in a mixin I'm not really convinced as these functions are already one-liners for threads (QThread() +  to create, thread_start.start() to start, etc. obviously there's some more stuff to do but it's really linked to each specific extension)

I feel the mixin would add complexity following who does what

packages/pymodaq/src/pymodaq/extensions/custom_ext.py
malik-irain: I'll add it in QuitableApp

packages/pymodaq_gui/src/pymodaq_gui/utils/shared_ui.py
malik-irain: affect_application can be used on daq_move.ui, daq_viewer.ui, etc. stuff that are not CustomApp but have some similarities with it, that's why it accept something else that's not a CustomApp, originaly the type was "Any".

QuittableApp are effectively classes that have at least a title, I can keep the CustomApp check and add title and toolbars attributes in QuittableApp MixIn

PR #1042, opened 2026-05-06
Quick fix for the name of the averaged data

rbeignon:
Avoid the averaged data to look like: PL_multiply_add_divide_multiply...

### DIFF
--- a/packages/pymodaq_data/src/pymodaq_data/data.py
+++ b/packages/pymodaq_data/src/pymodaq_data/data.py
@@ -1046,6 +1046,7 @@ class DataBase(DataLowLevel, NDArrayOperatorsMixin):
         if isinstance(other, DataBase) and len(other) == len(self) and isinstance(weight, numbers.Number):
             averaged_data = (other * weight + self) / (weight + 1)
             averaged_data.name = other.name
+            averaged_data.add_extra_attribute(**{'averaged': True, 'n_averaged': weight + 1})
             return averaged_data
         else:
             raise TypeError(f'Could not average a {other.__class__.__name__} or a {self.__class__.__name__} '
--- a/packages/pymodaq_data/src/pymodaq_data/data.py
+++ b/packages/pymodaq_data/src/pymodaq_data/data.py
@@ -1044,7 +1044,9 @@ class DataBase(DataLowLevel, NDArrayOperatorsMixin):
         DataBase: the averaged DataBase object
         """
         if isinstance(other, DataBase) and len(other) == len(self) and isinstance(weight, numbers.Number):
-            return (other * weight + self) / (weight + 1)
+            averaged_data = (other * weight + self) / (weight + 1)
+            averaged_data.name = other.name
+            return averaged_data
         else:
             raise TypeError(f'Could not average a {other.__class__.__name__} or a {self.__class__.__name__} '
                             f'of a different length')


packages/pymodaq_data/src/pymodaq_data/data.py
```diff
@@ -1044,7 +1044,9 @@ def average(self, other: 'DataBase', weight: int) -> 'DataBase':
         DataBase: the averaged DataBase object
         """
         if isinstance(other, DataBase) and len(other) == len(self) and isinstance(weight, numbers.Number):
-            return (other * weight + self) / (weight + 1)
+            averaged_data = (other * weight + self) / (weight + 1)
+            averaged_data.name = other.name
```
seb5g, line 1048: why not something like:

```
averaged_data.name = f'averaged_{name}'
```
to at least keep track that this is an averaged version of the initial data

Ashwolaa: Even 
`averaged_data.name = f'averaged_{name}_{weight+1}'`

to know the amount of frame used

rbeignon: Personnally I'm not so sure I want the name to change with N_average, I feel like it rather belongs in the parameters. I don't always use the same averaging numbers, I think having all my data named the same would be better for data treatement after

Ashwolaa: I only had in mind display purposes on the daq_viewer here to know at which acquisition state you are since this naming issue mostly occurs during live_averaging.

I agree that it should not change the saving structure and it indeed might change its title but I am not sure that is will change its name though.

What do you think @seb5g ? The add_data in DataSaverLoader seems to carry the data.name in its title.

seb5g:
That depends what averaging we are talking about:

* in the DAQ_Scan: no incidence as the averaged data are not saved 
* in the DAQ_Viewer: if the produced raw data is done with averaging, then indeed the data.name will be used in the saving mechanism (in the title attribute). We could also add in the averaging method an extra_attribute storing the number of object that have been saved. This will be saved as metadata automatically. Something like:
 
```python
if isinstance(other, DataBase) and len(other) == len(self) and isinstance(weight, numbers.Number):
    return (other * weight + self) / (weight + 1)
    averaged_data = (other * weight + self) / (weight + 1)
    averaged_data.name = other.name
    averaged_data.add_extra_attribute(**{'averaged': True, 'n_averaged': weight+1})
```

rbeignon:
Yes I like it this way ! with return averaged_data at the end of course: 

```python
if isinstance(other, DataBase) and len(other) == len(self) and isinstance(weight, numbers.Number):
            averaged_data = (other * weight + self) / (weight + 1)
            averaged_data.name = other.name
            averaged_data.add_extra_attribute(**{'averaged': True, 'n_averaged': weight + 1})
            return averaged_data
```

I tested with this it works well

seb5g:
To go beyond and answer some suggestions from @Ashwolaa , maybe the dataViewer could check for these attributes and print them in the viewer title?

@rbeignon could you commit the proposed modification?

seb5g:
#1045 is implementing this on the dev branch as it should be. @rbeignon make sure to use the dev branch to propose new features. Branch 5.1.x only accepts bugfix!

I'm closing this PR , you can continue to keep your code as is but the feature will only be available on dev or future 5.2.x

rbeignon:
Ok thanks !

PR #1044, opened 2026-05-11
Reseting the live averaging between acquisitions

rbeignon:
I use the live acquisition and whant to reset the measurement sometimes. I would like to have a way to "forget" the averaged data. I think that could be done when clicking on the stop (red square) button. So far I need to stop the measurement, un-click live aveaging, acquire data, click live averaging, and finally start accumulating data.

I implemented a way to do it but it is not very clean. I struggled a lot because when I click on the stop button while the detector is grabbing, an acquisition is done and averaged after whatever function is triggered by the stop button. There might be a better way of handling it but I didn't want to change the way things are done.

### DIFF
--- a/packages/pymodaq/src/pymodaq/control_modules/daq_viewer.py
+++ b/packages/pymodaq/src/pymodaq/control_modules/daq_viewer.py
@@ -164,6 +164,7 @@ class DAQ_Viewer(ParameterControlModule):
                                           module_saving.DetectorExtendedSaver] = None
         self._h5saver_continuous: Optional[H5Saver] = None
         self._ind_continuous_grab = 0
+        self.reset_averaging = [False, "None"]
 
         self.settings.child('main_settings', 'DAQ_type').setValue(self.daq_type.name)
         self._detectors: List[str] = [det_dict['name'] for det_dict in DET_TYPES[self.daq_type.name]]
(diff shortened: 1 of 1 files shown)


Ashwolaa:
Why do you need to stop the measurements? Pressing twice on live_averaging does not do the trick?
I agree that it could reset by default when pressing stop though.

This can be maybe added by default in the stop method of the daq_viewer. You can try this:

```
    def stop(self):
        """ Stop the current continuous grabbing """
        self.update_status(f'{self._title}: Stop Grab')
        self.command_hardware.emit(ThreadCommand(ControlToHardwareViewer.STOP_GRAB))
        self._grabing = False
        self._ind_continuous_grab = 0
```

seb5g:
Indeed what @Ashwolaa say shoulf work fine. In case you want the checkbox enabled during the live, make sure to tick the tight options in the preferences/configuration. 

Also make sure to propose new features from an uptodate dev branch!
 cheers
Seb

seb5g:
@rbeignon could you tell if what suggested @Ashwolaa answer your issue?

rbeignon:
Ah yes actually it works fine with this simple change!

seb5g:
@rbeignon would you then propose this change on the dev branch after you've resync your fork! I'm closing this one

seb5g:
new PR is: https://github.com/PyMoDAQ/PyMoDAQ/pull/1095

PR #1045, opened 2026-05-12
Feature/data averaging metadata

seb5g:
This is an implementation of #1042 for the dev branch

It adds two extra attributes in the dwa objects to store and inform about the level of averaging. Implemented in the average method. This is used in the viewer title description and stored in the h5files

It also fixes the weight in the DAQ_Viewer averaging


seb5g:
@rbeignon @https://github.com/rbeignon

This dev should be done on the dev branch while you did it on the 5.1.x branch that only accepts bugfixes. This was not really a bug and moreover we now introduce metadata to incorporate averaging information.

Make sure to use the dev branch if you want to propose new features.

Ashwolaa:
I would be tempted to add n_averaged directly as attribute of data and making averaged a property so something like this. Then, we do not need the hasattr / getattr as all data share the attributes. What do you think?

```
    @property
    def averaged(self) -> bool:
        """True if this data has been through at least one averaging step."""
        return self.n_averaged > 0
```


and in average

```
        if isinstance(other, DataBase) and len(other) == len(self) and isinstance(weight, numbers.Number):
            averaged_data = (other * weight + self) / (weight + 1)
            averaged_data.name = self.name
            averaged_data.n_averaged = weight + 1
            return averaged_data
```

seb5g:
yeah it is intrinsic of the data as usually it would be n_averaged=1, I'm just afraid sometimes you will compute average not using the average method and forget to adjust manually the n_averaged attribute, so the metadata would be wrong... 

Also just adding it as a property or regular attribute won't add it in the h5file metadata, we should then still use the extra attribute

Ashwolaa:
> yeah it is intrinsic of the data as usually it would be n_averaged=1, I'm just afraid sometimes you will compute average not using the average method and forget to adjust manually the n_averaged attribute, so the metadata would be wrong...

That's a fair concern, but it's inherent to any metadata system  if you bypass the method, you own the metadata.

> Also just adding it as a property or regular attribute won't add it in the h5file metadata, we should then still use the extra attribute

yeah but that is easily solved by adding it into the add_data method of the DataSaverLoader?
` metadata['n_averaged'] = data.n_averaged`

seb5g:
Actually I modified the code to both have them as properties but also as extra_attributes!

If you're fine with that, we should merge the PR!

PR #1046, opened 2026-05-13
Feature/cleaning/control module factorization

malik-irain:


seb5g:
Why did you closed it?

malik-irain:
The real one is still opened, this is a port of @Ashwolaa branch on my fork to investigate segfaults without putting all the trial and error part on github and triggering github actions email to each contributor each time I try something, but I forgot to change the target when opening the PR on my fork!

Also doing so he doesn't have to resolve conflict each time something changes ;)

PR #1047, opened 2026-05-13
Feature/ressource manager utils

Ashwolaa:
## Context
Many plugins rely on pyvisa / pyserial or other wrappers to find the correct adress on which to connect the instrument.
Overall, this usually leads to use a method within the plugin that will list the available ressources before connecting to the element in the list associated to the instrument.
```

            rm = pyvisa.ResourceManager()
            info = dict(rm.list_resources_info())

            from serial.tools.list_ports import comports
            return list(comports())
```
The issue is that, when you have many plugins installed, it is likely that you will recompute this for each plugin. This approach is redundant and very inefficient. 

## In this PR:

I suggest to move to a cache approach for it that can be used/retrieved by all plugins. These methods would be stored in pymodaq_utils and can be accessed by anyone within their plugins.

For now, I propose to have cache resources for pyvisa and pyserial. This would go as:

```
      # VISA-based plugin (Newport, Thorlabs, PI, ...)
      from pymodaq_utils.hardware.visa import list_serial_resources
      ports = list_serial_resources()  # ['ASRL/dev/ttyUSB0::INSTR', ...]

      # pyserial-based plugin (Arduino, Ocean Optics, ...)
      from pymodaq_utils.hardware.serial_ports import list_resources
      ports = list_resources()  # ['/dev/ttyUSB0', 'COM3', ...]
```

The first call will populate the cache that can then be accessed later on by any plugin. 

In practice, this permits to simplify the loading of connected instruments and gives commodity functions that can be used for most plugins.

Finally, it is particularly easy to extend this approach to other libraries that are often used within pymodaq's plugins.


seb5g:
I think this would be better in a dedicated plugin, well dedicated to communication features or instrument templating  like the one I'm using for camera... I don't want to put any communication related feature into pymodaq _utils. Why not having pymodaq_plugins_utils for utilities related to instruments?

seb5g:
It could be even restricted to visa stuff or serial stuff

Ashwolaa:
Yeah that would be a better location. I did not think too much about where to put that so pymodaq_utils seem like the default location but another folder could do things well too.
For now, VISA and serial will likely cover most cases so I am perfectly fine with it.


Envoyé à partir de Outlook pour Android<https://aka.ms/AAb9ysg>

seb5g:
alright then please push your code into: https://github.com/PyMoDAQ/pymodaq_plugins_utils
this is a plugin I just created. I will also incorporate into it my camera layout

Ashwolaa:
I removed the files and updated the docs

seb5g (approved):

PR #1058, opened 2026-06-09
in the underlying data stored in the model a space was included in th…

seb5g:
…e string

the combobox didn't want to be changed
also the value is now saved without explicitly pressing enter to set the new text of the combo!


malik-irain:
Issue with linting in this PR

`packages/pymodaq/src/pymodaq/utils/managers/state/utils.py::L223`
```python
suffix = f' {entry.setting.parameter.opts.get('suffix')}'
```
has two set of nested single quotes. One should be replaced with double quotes!

seb5g:
yes I bumped into it also. I patched it and will merge it soon

PR #1059, opened 2026-06-09
Feature/roi param sync

seb5g:
Creation of an object used to synchronise a ROI (color, position, size...) and its representation as a Parameter. Makes connection, signal and slots much easier to handle.

This is using the DictSync objects from pymodaq_gui together with the validator to handle the object emitted using ROI signals!

### DIFF
--- a/packages/pymodaq_gui/src/pymodaq_gui/managers/roi_manager.py
+++ b/packages/pymodaq_gui/src/pymodaq_gui/managers/roi_manager.py
@@ -1,9 +1,9 @@
 
 import os
 import sys
-from typing import List, TYPE_CHECKING, Tuple, Union
+from typing import List
 
-from qtpy import QtCore, QtGui, QtWidgets
+from qtpy import QtCore, QtWidgets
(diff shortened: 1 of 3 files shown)


seb5g:
@Ashwolaa could you have a look? There is an exemple at the bottom of the script for linear and 2D ROI. I will build on top of this to write a new ROIManager

Ashwolaa:
Do you want me to check the ROI sync specifically?

seb5g:
It's not for a former review but more for a check if I used correctly the sync features. It can also wait for when the New roi manager will be ready

Ashwolaa:
Ok I will wait a bit then.
Maybe as a side thing, you can decide when you do the binding the source of truth by picking either widget/sync.
`bind_properties(roi, property_map=roi_property_map, init_from='widget') `first (sync ← roi's current state)
 then
 ` bind_parameter(roi_parameter, property_map=parameter_property_map, init_from='sync')` (parameter ← that state).

seb5g:
closing this one as superseeded/included in #1100

PR #1061, opened 2026-06-10
Feature/timeout during scan

Ashwolaa:
Timeout during a scan can be quite cumbersome, especially if it only occured from only one instrument (maybe only for one datapoint).

In this PR, I propose:
-  to add a tickbox in the scan panel to decide if the scan should be stopped or not (by default, it does as before)
- add some information in the logger of the modul_manager to inform the user which instruments undewent a timeout.

In the long run, one could decide on this Timeout value/behavior per instrument instead to make it a bit more flexible.


packages/pymodaq/src/pymodaq/utils/managers/modules/modules_manager.py
```diff
@@ -559,8 +563,12 @@ def move_actuators(self, dte_act: DataToExport, mode='abs', polling=True) -> Dat
 
                 QtWidgets.QApplication.processEvents()  # mandatory for the det_done_flag boolean to be modified in the corresponding method
                 if time.perf_counter() - tzero > self.actuator_timeout / 1000:  # timeout in seconds
-                    self.timeout_signal.emit(True)
-                    logger.error('Timeout Fired during waiting for actuators to be moved')
+                    received_names = self.move_done_positions.get_names()
```
seb5g: you're using names here and not origin to take into account the muti-axes stuff. Don't you think you should do just the same for the viewers? It would be more similar and also make sure any evolution on the viewer side is taken into account! Is there an issue with ROI also?

seb5g (commented):
could you adress my point?

also is the timeout signature of the modules manager used elsewhere? Other extensions?

packages/pymodaq/src/pymodaq/utils/managers/modules/modules_manager.py
Ashwolaa: I'll check

Ashwolaa:
> could you adress my point?
>
> also is the timeout signature of the modules manager used elsewhere? Other extensions?

Nope, I did not find it elsewhere.

> you're using names here and not origin to take into account the muti-axes stuff. Don't you think you should do just the same for the viewers? It would be more similar and also make sure any evolution on the viewer side is taken into account! Is there an issue with ROI also?

I switched to origin for both daq_viewer and daq_move. This makes more sense as the timeout is related to the DAQ only , answering did it manage or not its actions?

Adding additional infos, such as missing channels should come from the daq_viewer itself. It could be interesting to work a bit more in expanding the timeout info but from the DAQ themselves. The info could then be gathered then transmitted by the modules_manager. No problem for the ROIs at first glance, everything is managed from the grab_done_signal which contains it if I understood it well.

seb5g (approved):

Issue #1064, opened 2026-06-12
[BUG] h5py and swmr mode for enlargeable data and optimizers extensions


seb5g:
### Describe the bug

When doing an optimization. The 
```python
 self.enlargeable_saver.add_data('/RawData', dwa_data, axis_values=actuators_values)
```
command is calling the reading of attributes. This is working fine with pytables but somehow with h5py it's not. I tried to deactivate the swmr mode in the preferences but to no avail. Restarting the DashBoard with pytables as a backend work fine.

### Expected Behavior

I want the enlargeable and live plot associated with it to be transparent whatever the backend I'm using

### Steps To Reproduce

1. create/load an Experiment
2. Open the Bayesian extension
3. configure the default model
4. start the optimization

the optimization happens but the live plotting and continuous saving fails

### Reference(s)

```markdown
pymodaq.utils.config.deep_update
```

### Environment

```markdown
- OS: windos 10
- Pymodaq: dev (5.2.x)
- python env: 3.12.*
```

seb5g:
recent work on saving data #1143 produced a saving/loading pattern using threads and properly closing the file to fix this type of issues. 

If interested or having similar problems look at the example: pymodaq_gui.examples.saving_loading

PR #1069, opened 2026-06-12
Hotfix/h5py for enlargeable data

Ashwolaa:
Fixing #1064 

To be honest, I am a bit lost into it. 
What I seem to have understood is:
 - CLASS is not automatic with h5py. This PR patches it up with a new method `_infer_class` when CLASS is not found.
 - I found a TypeError: No conversion path for dtype: dtype('<U11') in array.attrs['TITLE'] = title. This is now fixed by enforcing that 
data.name =  'probed_data'

I can run the optimizer with it without clear issues but the first part seems quite patchy.


seb5g (approved):

seb5g:
I accepted it as it's working but this is weird indeed that the CLASS that seems to be a basic stuff is not present with h5py??? is there an equivalent?

Issue #576, opened 2025-03-25
Units of 0D data


bernhardlang:
0D data are displayed with ##Samples as x-axis. This seems to be hard-coded. Though, a series of 0D data, displayed in a 2D graph, may yet carry a specific unit. For example time when the acquisition of single items happens upon external trigger events, or, say, the position of an X-ray grating shared by different experiments hosted at the same beam line. Whenever such parameter is not under direct control of PyMoDAQ, data may be of a "quasi-1D" type, just coming in point by point like 0D, instead of an entire set at a time as is the case with "true" 1D data.

seb5g has recently pointed out a solution in a similar case, collecting 0D data in the plugin and sending it when appropriate to PyMoDAQ as 1D data. That probably make sense when the range of the x-axis is known beforehand (example X-ray grating). The scan cannot take place in a DAQ scan -> take care of it in the plugin.

In case that the x-values associated with the 0D (y) samples are not known beforehand or at least their range, the collecting and shifting (once the number of items increases beyond a displayable limit) has to do the same as what is already implemented in a 0D viewer. Re-implementing it in the plugin a) feels like violating the DRY principle and b) may lead to issues when past items are shipped to be re-displayed but should not be stored again.

A solution within the existing framework could probably be to export x and y values in parallel as two 0D data items. Then, a specialized plot would be needed which combines these values for plotting them on a "running plot" as implemented in the 0D viewer, for monotonically increasing (or decreasing) x-values, that is to say. Otherwise a scatter plot simply accumulating the items could do the job.

Or did I oversee something already being present?

PR #1083, opened 2026-06-18
0DViewer: using Timestamps as axis and using scatter

seb5g:
In this PR, I implement the possibility to plot 0D data using the timestamps the data where taken at. We keep the history but provide the possibility to switch between samples or timestamps.

The PR also expose a new action/icon to plot the curves using continuous lines or scatters

How to test it: start a DashBoard with default Experiment and grab/snap the 0D Viewer. Try to trigger the timer icon to switch between samples and timestamps and the scatter icon to change the brush


seb5g:
the last commit also adds the possibility to plot second channel as a function of the first!

all this should fix #576

Ashwolaa:
I like the idea, I was actually looking to have this for some of my instruments instead of just having acquisition indexes

seb5g:
I think they're could be another  improvement by using a specialized data axis provided in pyqtgraph

bernhardlang:
I tested the way you propose and didn't encounter any problems. I'll do further testing in a custom app I've got to whip up until August.

An even more general approach to the matter could be to permit joining any two related 0D data sets in an x-y plot, or even plot n-1 sets against the first one or against any single out of the n. But I'm not sure whether that would be too specialized to incorporate it into the general package.

seb5g:
Actually I thought about this as well and in favor of plotting any second and more waveforms as a function of the first. It is up to the developer to order the data in the plugin to benefit from this.

seb5g:
In the end I implemented this:

* plot n-1 sets against the first one
* using DateAxisItem when using timestamps. This axis change the tick labels depending on the zoom level (ms, seconds, hours, days, year)! Quite neat!

I think i'm done with the PR so maybe do another round on this and if you don't see any obvious bug, I'll merge it. We can come back to it later if needed!

PR #1085, opened 2026-06-19
Tabular scan save & load fix

### LINKED ISSUE #1084: Tabular scan save & load from txt not working
It currently refuses to save because of format (expecting a float but is given a string). Similar issue when trying to load

malik-irain:
Fixes #1084


seb5g:
I also corrected the display for the sparse mode. For this I now pass as an argument to the ScannerImplementations, the settings argument

PR #1086, opened 2026-06-20
Feature/Adding_ScanInfo_to_daq_viewer_plugins

Ashwolaa:
# Expose scan context to detector plugins via `ScanInfo`

## Context

During a `DAQ_Scan` run, `DAQ_Viewer` plugins have no officially supported way to know
which scan step is currently executing or where PyMoDAQ is writing its HDF5 file. This
matters for any plugin that writes its own files alongside PyMoDAQ's `.h5` (vendor cameras
with proprietary binary formats, time-stamped detectors, etc.) and wants those files
organized to mirror the scan structure PyMoDAQ already uses (`<Dataset_dir>/Scan001/...`),
rather than inventing independent session bookkeeping.

The problem is that a plugin's `self.parent` is the hardware-worker `QObject` running on
its own thread — it only exposes `status_sig`, not the GUI-side `DAQ_Viewer` that owns the
`H5Saver`. There was no clean existing channel to reach this information from inside a
plugin.

## Strategy

The scan context already exists inside `DAQScanAcquisition` (`self.ind_scan`,
`self.ind_average`) and is derivable at the scan-node creation site (`scan_node.name`,
`h5saver.settings['current_h5_file']`). It simply was never threaded through to the plugin.

Rather than adding loose keyword arguments (fragile, hard to document, hard to extend),
this PR bundles the context into a single typed dataclass:

```python
@dataclass
class ScanInfo:
    ind_scan: int        # zero-based linear step index
    ind_average: int     # zero-based averaging pass (0 when scan_average == 1)
    scan_node_name: str  # HDF5 group name, e.g. 'Scan001'
    h5_file_path: str    # absolute path to the HDF5 file being written
```

A `ScanInfo` instance is constructed once per step inside `DAQScanAcquisition` and
forwarded as a single `scan_info=` kwarg through the existing chain:

```
DAQScanAcquisition  →  ModulesManager.grab_data(**kwargs)
  →  mod.command_hardware.emit(ThreadCommand(SINGLE, kwargs))
  →  DAQ_Viewer_Hardware.single(**kwargs)
  →  plugin.grab_data(Naverage, live=False, **kwargs)
```

`ModulesManager.grab_data` already accepts and forwards arbitrary `**kwargs`, so no change
was needed there. All existing plugin signatures (`grab_data(self, Naverage=1, **kwargs)`)
already absorb unknown kwargs, making this **purely additive — no existing plugin breaks**.

Outside a scan (live view, manual snap) the kwarg is simply absent, so plugins guard with
`kwargs.get('scan_info')`.

## Usage in a plugin

```python
def grab_data(self, Naverage=1, **kwargs):
    scan_info = kwargs.get('scan_info')  # None when not in a scan
    if scan_info is not None:
        out_dir = Path(scan_info.h5_file_path).parent / scan_info.scan_node_name
        out_dir.mkdir(parents=True, exist_ok=True)
        # write proprietary file aligned to PyMoDAQ's layout, e.g.:
        # out_dir / f"frame_{scan_info.ind_scan:05d}.bin"
```

This lets plugin authors align their own file tree exactly to PyMoDAQ's without any
guesswork.

## Files changed

- `extensions/scan/daq_scan.py` — `ScanInfo` dataclass; `DAQScanAcquisition` accepts
  `scan_node_name` / `h5_file_path`; per-step grab passes a populated `ScanInfo`
- `extensions/scan/__init__.py` — re-exports `ScanInfo` for
  `from pymodaq.extensions.scan import ScanInfo`
- `control_modules/viewer_utility_classes.py` — `DAQ_Viewer_base.grab_data` docstring
  documents the new kwarg


seb5g:
This is a good idea to pass things through however what you said could be true for any extension using the instruments plugins. Could you think of a more general way to expose information about the h5file structure without being so specific about the scan?

 It could be a Caller object and each caller would know what it should expose and bundle in a dataclass, in particular I think h5file path, current node are the two info shared by any extension (even the info within a DAQViewer in continuous saving mode) in a CallerBase object that could be subclassed by ExtensionCallers

Finally what other info should be exposed would depend on the extension and what you will need from within the plugin. Do you plan to save data from within the plugin into the current h5 file? I guess no, then what do you want to do?

seb5g:
and also any new feature should be targeted to dev branch!

Ashwolaa:
> This is a good idea to pass things through however what you said could be true for any extension using the instruments plugins. Could you think of a more general way to expose information about the h5file structure without being so specific about the scan?
> 
> It could be a Caller object and each caller would know what it should expose and bundle in a dataclass, in particular I think h5file path, current node are the two info shared by any extension (even the info within a DAQViewer in continuous saving mode) in a CallerBase object that could be subclassed by ExtensionCallers
> 
> Finally what other info should be exposed would depend on the extension and what you will need from within the plugin. Do you plan to save data from within the plugin into the current h5 file? I guess no, then what do you want to do?

I agree, this should be generalized. I will look into it.

In my use case, I am using pymodaq to launch the scan (and store most of the parameters and other instruments) and I have a dummy daq_viewer that just launches the saving mechanism of my instrument through ZMQ. This daq_viewer did not know about the pymodaq saving structure and I wanted it to share this information to my instrument acquisition software to write the file in the same folder with the same filename formatting approach.

> and also any new feature should be targeted to dev branch!

Indeed, I did not pay attention ^^

Ashwolaa:
Is this what you had in mind? I made a Caller class that is either used and given by the Extension itself (DAQ_Scan, ...) when asking a grab_data or directly taken from the module's GUI for the direct save (when there is no Extension above)

seb5g:
Yes that's way much better now. Just finally wondering if caller is such a nice name... not that expressive I guess. I like it also to be active in calling get_caller from the plugin and not having it directly in the kwargs. 

If you have a better name, i'd like to hear it and otherwise let's go for this and adds some documentation about it in the developper section!!

Ashwolaa:
Yeah agree, it is easy to say but not so precise. 
AcquisitionContext / H5AcquisitionContext ?

seb5g:
I think you focused on h5 but some other need may arise at some point unrelated to h5. So I would stick with something not too specific like CallerContext (even acquisition seems too specific), could be used for acuator as well...

I would then include the name of the caller: extension or DAQ_Viewer's title itself. MainCallerContext?? OriginCallerContext??

seb5g:
I think we were almost there with this PR, let's go with CallerInfo dataclass, including its name and maybe it's type (class name)!

Ashwolaa:
Should be good now with CallerInfo now but still some threading issues with pyleco

packages/pymodaq/src/pymodaq/control_modules/viewer_utility_classes.py
```diff
@@ -194,9 +194,26 @@ def close(self):
...
+            Extra keyword arguments forwarded transparently from the caller.
+
+        Notes
+        -----
+        Extensions that drive this plugin (e.g. ``DAQ_Scan``) may set caller context
+        describing the active HDF5 file/node, so a plugin can mirror PyMoDAQ's file
+        layout for its own files. Call ``self.get_caller()`` (inherited from
+        :class:`~pymodaq.control_modules.pluginBase.PluginBase`) at any point in the plugin's
+        lifetime — not just inside ``grab_data`` — to read it. It returns ``None`` outside
+        of an extension-driven run. During a ``DAQ_Scan`` run it is a
```
seb5g: is it not returning the fallback module_and_data saver caller context?

seb5g (changes requested):
Only one question/comment concerning the fallback

Also could you generate some dock on the developper section to let people know of this possibility within a plugin?

packages/pymodaq/src/pymodaq/control_modules/viewer_utility_classes.py
Ashwolaa: you are right, I fixed the docstrings

seb5g:
Actually is it possible/safe to write into the same h5 file as the scan as they live in different threads?

Ashwolaa:
I really don't think so. Maybe on Linux (at least no problem when it is just about appending the file) but I have no clue on how to do that on windows.

seb5g:
then what is the purpose of having the file and its structure within the plugin (that is in a different thread)?

Ashwolaa:
I initially only wanted the file path ! This was to synchronize the saving between my own GUI and pymodaq

seb5g:
You mean to create another file but having the exact same structure?

PR #1097, opened 2026-06-23
color patch

seb5g:
The color called  in the constructor from ROI color is a method and is lacking parenthesis


seb5g:
this is quite strange, in the work I was on, it was lacking this parenthesis and in the tests here it's not...

seb5g:
I messed it up, reverting the change and closing

PR #1099, opened 2026-06-26
Refactoring LECO Communication + scripting polishing

malik-irain:
It will refactor the way LECO is integrated into PyMoDAQ, so that adding a component needs less boilerplate code.

For now it contains some polishing such as renaming RPC to upgrade `preset`/`configuration` to `experiment`/`state`, enable and disable the scripting from a menu. 

Example:
image

When activated, the dashboard is connected to LECO. If an experiment is loaded (before or after activating), each device gets connected through a Signal triggered from the dashboard.

image
When disabled, the same everything is disconnected.


seb5g:
that's quite nice!! 
I have some issues regarding how should I know there is a problem or not with the connection. Maybe a popup when everything is connected successfully and inviting people to start using scripting?

is there still the option to start it automatically at startup by using the preferences?

seb5g (approved):

PR #1100, opened 2026-06-26
Feature: ROI Sync and Brand New ROI management / Manager

seb5g:
This PR build on top of the WidgetSync applied to ROI and Parameter representing their state

I've introduced several objects to help make things clean and bundled together:

In roi_sync module:

* `ROIParameter`: addhoc derived from group parameter exposing children adapted to 1D or 2D ROI
  * Contains the settings
  * exposes methods to produce `ROI `having the same values as the settings (using the `ROIFactory`) 
  * exposes a method (to_options) to return a dataclass: `ROIParameterOptions` to bundle those values shared between ROI and settings
* `ROISync` whose dict values are init from `ROIParameterOptions` and exposes a method to sync with these values a `ROI` and a `ROIParameter`

In roi_viewer_manager:

* `ROIParameterManager`: ParameterManager showing the ROIParameter as children of a top Parameter of type `ROIScalableGroup`
* `ROIViewerManager` extends  `ROIParameterManager` with QObject to trigger signals both from the Parameters or from the ROI
* `ROIMeta` is an basic object bundling ROIParameter, ROISync and ROI and syncing them all. The `ROIViewerManager` is exclusively manipulating those `ROIMeta` 

This PR is therefore a complete update of the existing ROIManager that has been deleted. The future ROIManager will exists on the DashBoard level to allow quick saving and loading of ROIs in several viewers (to be done)

The main advance due to all this is the perfect and smooth synchronization between ROI/ROIParamter values/attributes as well as adding/removing/copying them (from the tree menu or the ROI context menu)

As a side I included new MaterialIcons for the viewers. Those could be modified anytime @runrunm :-)


seb5g:
Forgot to say that this work also triggered the refactorisation on the way we are using ROI with respect to plot axes. The main issue was that 2D arrays are by default displayed as a function of the line/column index. From early on I introduced a ScaledAxis allowing to "see" the real axes on top/right. However this induced two issues:

* the ROI expressed in such a ViewBox displays their position and size in the column/row integer value axes (while it is not for 2D Spread data and 1D data)
* When slicing such a data and plotting the slice, the image would move away from the ROI used for the slicing

I'm now using a QTransform on the ImageDisplayer onjects offsetting and rescaling the image wrt the axes. In this frame, the ROI are now expressed directly as a function of those axes.

To ensure backcompatibility or at least make things work again, one should slicing by value (vsig) when using ROIInfo:

```
dwa.vsig[roi.to_info().to_slices(False)]
```

I developped a 2D instrument plugin in MockExamples (ROIStuff) to test this feature using the ROISelect ROI 

image

Ashwolaa:
There's a lot in this PR in addition to just the ROI. I think it would be best to decompose it in multiple PRs to get a better track of all changes.

seb5g:
I agree but the problem with this is that all the pr are interlinked and need each other . It's a bit the same as what we have when using a single repository for pymodaq. so here I needed a single pr...

seb5g:
I guess the way to tackle it is probably just by usage... @Ashwolaa do you think you could try and use it in your everyday life as an alpha version?

Ashwolaa:
I'll give it a shot but I am not really using any of these tools these times. My current dashboard has only 0D signals and modules

Envoyé à partir de Outlook pour Android<https://aka.ms/AAb9ysg>

seb5g:
ok well I'll discuss of this with @adriente who is doing some dev using the ROI stuff!

PR #1106, opened 2026-06-28
UX: Feature/highlight manager entry

seb5g:
In this PR, I added the highlighting of the entry of the manager that is currently being activated

When nothing is activated, it looks as usual

but it looks like this when BeamSteering experiment is active:

image

and like this again if another experiment is selected but not yet activated:
image

This is valid for all Managers inheriting from ManagerBase !!

I'm not hundred percent satisfied but styling combobox is tricky...

### DIFF
--- a/packages/pymodaq_gui/src/pymodaq_gui/managers/manager_base.py
+++ b/packages/pymodaq_gui/src/pymodaq_gui/managers/manager_base.py
@@ -16,7 +16,7 @@ from pymodaq_utils.enums import StrEnum
 from pymodaq_gui.messenger import dialog
 from pymodaq_gui.qt_utils import center_widget_on_screen_and_show, mkQApp
 from pymodaq_gui.utils.splash import get_pymodaq_pixmap
-from pymodaq_gui.utils.widgets.combo import ComboBox
+from pymodaq_gui.utils.widgets.highlighted_combo import HighlightedComboBox as ComboBox
 
 from pymodaq_gui.utils.widget_sync import WidgetSync, SyncMode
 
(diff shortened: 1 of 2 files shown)


seb5g:
@runrunm is it what you had in mind when you talked about this during the training?

runrunm:
I don't remember talking about highlighting comboboxes. I am not sure it helps making things clearer.

If there is a need to clarify which experiment/state/overshoot is selected from a manager, maybe a good idea would be to start it as soon as it is clicked within the drop-down list. That way we remove 3 identical "start" buttons image and we get rid of the potential issue of not knowing what is selected/activated. Actually this is the default behavior when selecting an experiment from the tool menu. Clicking on an experiment in the drop down list just starts it.

Is there a situation where we would like an experiment/state/overshoot to be selected but not activated ?

seb5g:
> I don't remember talking about highlighting comboboxes. I am not sure it helps making things clearer.
I understood you said it would be good to know if the experiment (or other manager entry) is just selected or activated. Highlighting is doong this job.
> 
> If there is a need to clarify which experiment/state/overshoot is selected from a manager, maybe a good idea would be to start it as soon as it is clicked within the drop-down list. That way we remove 3 identical "start" buttons image and we get rid of the potential issue of not knowing what is selected/activated. Actually this is the default behavior when selecting an experiment from the tool menu. Clicking on an experiment in the drop down list just starts it.

I think you may be right....
> 
> Is there a situation where we would like an experiment/state/overshoot to be selected but not activated ?

This is True for experiment/state/overshoot/roi but is not for extension as we may want to restart an extension already selected. Maybe there could be two way of doing it either direct activation for experiment/state/overshoot/roi or with the activate button for some others. Some class attribute could trigger one or the other way. 

We will need to find a use for my highlighted combo as I spent several hours on it :-(

PR #1109, opened 2026-06-29
Feature/adding viewer config to config gui

Ashwolaa:
# Add viewer UI defaults to config

This PR exposes viewer toolbar/action defaults in the global config file, allowing users to persist their preferred viewer state across sessions.

Changes:
- Added a [viewer] section to config_template.toml covering viewer0D, viewer1D, viewer2D, and viewerND, with one entry per configurable toolbar action and a comment explaining each
- On startup, each viewer now reads its config section and triggers the appropriate actions to match stored defaults (replacing the previous hardcoded defaults)

How to test:
1. Open ~/.pymodaq/config.toml and set e.g. [viewer.viewer2D] histo = true
2. Open a Viewer2D — the histogram panel should be visible by default

### DIFF
--- a/packages/pymodaq_gui/src/pymodaq_gui/plotting/data_viewers/viewer0D.py
+++ b/packages/pymodaq_gui/src/pymodaq_gui/plotting/data_viewers/viewer0D.py
@@ -7,6 +7,7 @@ import sys
 import pyqtgraph
 
 from pymodaq_utils import utils
+from pymodaq_utils.config import GlobalConfig as Config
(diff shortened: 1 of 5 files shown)


seb5g:
nice **but** this is valid for all viewers of the same type. Could you think of a way to configure this independently for each viewer in a dashboard. I'm in to accept this one but maybe the State manager could help in configuring the state of individual viewers?

seb5g (approved):

PR #1110, opened 2026-06-29
Featyre/master slave same thread

seb5g:
In this PR, I modified the thread handling in case of master/slave control modules. 

Modules sharing a controller as Master/Slave were so far split into separated thread which caused a lot of trouble when several calls to the controller could happened at the same time (very bad when communication is serial). (Some controller are thread safe and were handling it better).

In this PR, both the thread and the controller are shared (bundled into a dataclass) and properly terminated/removed when either master or slave is desinit.

The direct outcome is that you can deconnect in the dashboard a slave and reconnect it seemlessly (if you don't change master/slave options)

**NEXT STEP TO COME**: beng able with the help of a Qthread pool within the dashboard to connect/reconnect master then slave by just changing their settings.


seb5g:
@Ashwolaa and/or @bernhardlang could you tell me what you think of this as we discussed this together some time ago?

bernhardlang:
If I get it correctly, any potential slave plugin is moved to the same thread the master and controller are living in already? That together with the thread essentially only relying signals to slots is a good design.

What advantages would it give to use a QThreadPool? If I understand it correctly, the controller thread's lifetime is the lifetime of the instrument initialized by the dashboard, right? It means, the thread lives until the user decides to change and re/load the experiment. That doesn't look like a throw-away thing to me where one could spare resources. It'd rather bloat the code. But I may miss something here.

seb5g:
Sorry my choice of word is inadequate. I meant a pool like just a reference of all the threads and controllers living within the dashboard and any control module r could pick the one it needs to if it is a slave

packages/pymodaq/src/pymodaq/control_modules/utils.py
```diff
@@ -37,6 +37,109 @@
...
+            status = self.ini_hardware(*command.attribute)
+            self.status_sig.emit(ThreadCommand(ThreadStatus.INI_HARDWARE, status))
+        elif command.command == ControlToHardware.CLOSE:
+            self.close_hardware()
+        else:
+            return False
+        return True
+
+
+class QThread(QThread):
```
bernhardlang: Perso, I would avoid re-declaring classes (like class QThread(QThread): ) to keep clear what is external and what is not. But that's a matter of taste and style.

packages/pymodaq/src/pymodaq/control_modules/utils.py
```diff
@@ -551,21 +683,36 @@ def init_hardware(self, do_init=True):
...
+
             self._setup_hardware_thread(hardware)
 
             self.command_hardware[ThreadCommand].connect(hardware.queue_command)
             hardware.status_sig[ThreadCommand].connect(self.thread_status)
             self._update_settings_signal[edict].connect(hardware.update_settings)
             self._connect_hardware_signals(hardware)
 
-            self._hardware_thread.hardware = hardware
+            self.controller_thread.thread.add_hardware(self.title, hardware) # to hold a reference
```
bernhardlang, line 705: Why do you split book keeping of "hardware" (method of QThread) and handling its thread affinity (in self._setup_hardware_thread) into two methods? Shouldn't that better be done in a single place?

packages/pymodaq/src/pymodaq/control_modules/utils.py
bernhardlang, line 723: hardware_thread has changed into controller_thread
Wouldn't it be sensible to change the name of the method accordingly?

packages/pymodaq/src/pymodaq/control_modules/utils.py
```diff
@@ -576,9 +723,9 @@ def _setup_hardware_thread(self, hardware):
         Default: always move and start. Override when the move/start should be
         conditional (e.g. DAQ_Viewer's ``viewer_in_thread`` config option).
         """
-        hardware.moveToThread(self._hardware_thread)
-        self._hardware_thread.finished.connect(hardware.deleteLater)
-        self._hardware_thread.start()
+        hardware.moveToThread(self.controller_thread.thread)
+        self.controller_thread.thread.finished.connect(hardware.deleteLater)
+        self.controller_thread.thread.start()
```
bernhardlang: Does it matter that a slave could accidentally start the controller thread here before the master came to this line? Or is that guaranteed not to happen? (in which case QThread.start() called a second time returns without doing anything)

Ashwolaa:
I have my own [branch ](https://github.com/Ashwolaa/PyMoDAQ/tree/feature/controller-thread-architecture)that I am still working on adressing some of it.
I went with a more complex route but your approach is more direct and I can rebase partly on that.

seb5g:
Ok thks for your comments. I will keep going with that, in particular with a way to fetch the right controller/thread any time for a slave as long that a corresponding master is on. 

This will solve the issue of a need to configure a slave with a State, then init. Indeed at the moment, slave init can only be done at the experiment step, so if some modifications of let's say COM port need to be done we're screwed... Of course the defualt COM port could be set in the preferences but this would require potential plugin modification while with this solution it will be transparent.

As for your points @bernhardlang I have to say those several classes dealing with control modules are quite messy. I think we will soon be in a need of a fresh rewritting. This will be the occasion to adopt (or not) the idées we discussed before with @Ashwolaa. But this will need careful preparations

packages/pymodaq/src/pymodaq/control_modules/utils.py
seb5g: I can rename it differently to make it clearer

packages/pymodaq/src/pymodaq/control_modules/utils.py
seb5g: I didn't actually, I think @Ashwolaa was the one splitting it (but read my general comment below)

packages/pymodaq/src/pymodaq/control_modules/utils.py
seb5g: controller_thread is meant for "Controller and Thread" dataclass but the thread is still the one of the hardware.

seb5g: should probably rename it also ;-)

packages/pymodaq/src/pymodaq/control_modules/utils.py
seb5g: so far it cannot (i hope it can"t ). But this is the dashboard and the experiment manager who're doing the routing first starting the master then the slave. Later I will make sure that a slave won't start if it cannot find a running master

PR #1115, opened 2026-07-02
Feature/rotation action icon

### LINKED ISSUE #1076: [BUG] flipped icon, color not showing
## Summary

The icon of buttons, doesn't behave as expected when using the option ```flip_h```.

## Full description

I am writing a custom app and some of its buttons activate only after data are loaded. The buttons are, on init,  created this way (for example) : 

```python
self.add_action('previous', 'Previous data chunk', 'step',
                        tip='Display data of the previous integrated period of time.',
                        checkable=False,
                        enabled=False,
                        flip_h=True,
                        icon_color='green')
self.add_action('next', 'Next data chunk', 'step',
                        tip='Display data of the next integrated period of time.',
                        checkable=False,
                        enabled=False,
                        icon_color='green')
``` 
Then, the action of loading data unlocks the usage of the ```next``` and ```previous``` buttons. It is done using the following code : 

```python
self.get_action('next').setEnabled(True)
self.get_action('previous').setEnabled(True)
``` 

Before loading data the icons look like that : 

Image

Only the file icon is working (as intended), while the arrows are inactive.

and after loading the data it looks like that : 

Image

Then all the buttons are active.

### Expected Behavior

The 'next' button works fully as intended and the 'previous' button should follow the same behaviour. 

### Steps To Reproduce

_No response_

### Reference(s)

```markdown

```

### Environment

```markdown
- OS: Windows 11
- Pymodaq: 5.2.3
- python env : python3.12, pyqt5
```

### Additional context

_No response_

### LINKED ISSUE #1102: [BUG] Icon Transform loses the colors
### Describe the bug

When using fliping or angle in transform_icon (pymodaq_gui.utils.styling) the color initially set on the icon is lost!

### Expected Behavior

Preserve the colors even if transformed

### Steps To Reproduce

1. create an action with the action manager
2. apply flip or rotation

### Reference(s)

```markdown
transform_icon
```

### Environment

```markdown
- OS: win10
- Pymodaq: 5.3.0dev
- python env: 3.12.*
```

### Additional context

_No response_

seb5g:
Add rotation transform at icon creation and with addaction

### DIFF
--- a/packages/pymodaq_gui/src/pymodaq_gui/examples/material_icon_ex.py
+++ b/packages/pymodaq_gui/src/pymodaq_gui/examples/material_icon_ex.py
@@ -1,5 +1,8 @@
 from pathlib import Path
+from random import choice, choices
 
+import numpy as np
+from pyqtgraph import mkColor
 from qtpy import QtGui, QtWidgets
 
 from pymodaq_gui.resources.material_icons import MaterialIcon
@@ -15,7 +18,7 @@ icons = toml.load(here.parent.joinpath('resources/icons.toml'))['icons']['names'
 icons.sort()
 
 class Icons():
-    def __init__(self, widget: QtWidgets.QWidget):
+    def __init__(self, widget: QtWidgets.QWidget, with_transform=False):
         super().__init__()
         self.widget = widget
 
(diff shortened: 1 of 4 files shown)


seb5g:
fix #1076 
fix #1102

PR #1117, opened 2026-07-04
Feature/ux popup or not

### LINKED ISSUE #1116: [BUG] Docking of settings window
### Describe the bug

There is an option in Preferences/Pymodaq/Viewer/Settings_as_popup so that settings are appearing in an external window. This option does not work properly : it is impossible to get the settings windows to dock in the main window (when the preference is unchecked).

### Expected Behavior

When the setting Preferences/Pymodaq/Viewer/Settings_as_popup is unchecked, the settings widgets should dock somewhere in the main window.

### Steps To Reproduce

Run a daq_viewer instance while the Preferences/Pymodaq/Viewer/Settings_as_popup is unchecked

### Reference(s)

```markdown

```

### Environment

```markdown
- OS: Windows 10
- Pymodaq:5.2.x
- python env: python 3.12, pyqt5
```

### Additional context

I would like to mention that this option appears in "viewer" even though it concerns both actuators and viewers.

seb5g:
Fix #1116 

In this PR The user has the choice to display ControlModules Settings either in a popup window (as is standard in 5.2) or in a dedicated Dock. The idea is to have both controls and viewers on the same window.

The PR implements the same for the extra controls of the actuators

ROIManager will behave similarly

https://github.com/user-attachments/assets/65b166a8-bc14-40ce-8cd9-3012868a95e5


seb5g:
@runrunm I believe this will be much better for UX in particular to have all in the same window but fully modular!

@adriente this should answer your wish!

seb5g:
this has been applied to ROIs as well

PR #1118, opened 2026-07-05
allows two default values to prefill the move_abs spinboxes

### LINKED ISSUE #965: values of binary actuator as settings
The two default values for an actuator of binary type (one/two or 'green'/'red') are determined from PyMoDAQ's general settings. It would be nice to be able to configure them per device and per axis.

Suggested solution: add the two values as generic parameters to the list of params of a binary move so that the they can be defined from a preset.

Application example: shutters driven by PWM servo motors. To drive the shutter to the two positions 'open' and 'close', the two corresponding PWM values have to be set by the move. During a configuration phase, the user will have to play with these parameters until the mechanical position of the shutter correctly covers / frees the beam. And these values will change from device to device.

seb5g:
This PR adds two settings in DAQ_Move main settings to define default values to prefill the move_abs spinboxes (red and green). They are prefilled themselves with config/preferences default values. As they are settings, they can be set/prepared in a StateManager hence easily configuring any actuator.

An other approach (just thinking about it) would be to create special entries for the State manager to allow to set these initial defaults...


seb5g:
fixes #965

packages/pymodaq/src/pymodaq/resources/config_template.toml
```diff
@@ -10,6 +10,9 @@ timeout = 10000  # default duration in ms to wait for data to be acquirred
 siprefix = true  # tell if printing of current value use a SI prefix or not (µ, m, k, M...)
 siprefix_even_without_units = false
 display_units = true # display units in the SpinBoxes
+default_value_red = 0.0
+default_value_green = 10.0
```
bernhardlang: What makes 10 different from 1 ? :-)

bernhardlang (commented):
Looks fine to me. As far as I can see for now, that should cover what is needed with pretty minimal changes. Would you see any advantage of adding special entries to the state manager?

PR #1121, opened 2026-07-13
Feature/setup UI at init

seb5g:
Follow up on #1119 and #1120 to set the DAQ_Move uis according to the capabilities of the selected plugin:

A plugin is only binary: the UI should be Binary, same for Relative only or Simple in the general case. So PyMoDAQ will now check for the existence in a plugin of a class attribute specifying which of these UIs should be used. For the Relative only case, all the absolute actions should not be visible, morevover get_actuator_value and move_abs should not be called. 

This PR implements this: check the selected instrument, infer it's UI and adapt it. 

The changes are not backcompatible in the sense that the UI factory mechanism has been changed. The UI is always the base, and is composed of a custom class (relative, binary, simple) modifying the feel/shape of this base ui. The modifications are therefore marginal but I don't think one should be able to completely modify the UI especially that now we have compact versions. For real custom stuff, just create a CustomApp.

There is an order in the way UI type are applied:

1. check if the actuator define a particular UI to be used
2. if (1) is 'None' (default) and not in DashBoard, check and use the default UI in config
3. if (1) is 'None' (default) and in DashBoard, the value chosen in Experiment Manager is used

DashBoard screenshot:

image

DAQ_Move demo:

https://github.com/user-attachments/assets/7ed06ebc-bdaf-4ca8-b0f0-52c532968b4a


seb5g:
@bernhardlang , @Ashwolaa  and @rgeneaux you may be interested in this PR and it's functionnalities. I think you may be using these, could you tell me if this is still in line in how you want to use these UIs?

please read #1119 and #1120 before to assert all the functionnalities!

seb5g:
@CitizenErase this is also related to a question you asked during the training namely how to set the initial values to the abs spinboxes at startup: use general preferences or use state manager in DashBoard to customize these individually

seb5g:
The documentation on the DAQ_Move has been fully updated to reflect these changes!

bernhardlang (commented):
To me that looks pretty good. Do you want me to look deeper into the new code? Since there are quite a few changes spread out over many files, that would take considerably more time.

The only further thing coming into my mind concerning specialized actuators would be relative ones of the (rather cheap) piezo-friction type where you give a longer or shorter button push and get some feedback on how far the thing actually went. That would be a relative UI getting a feedback from the device on the current position. Are there needs to implement such a thing?

seb5g:
> To me that looks pretty good. Do you want me to look deeper into the new code? Since there are quite a few changes spread out over many files, that would take considerably more time.
> 

No need to go in the code too much, I'll just wait a return from @Ashwolaa to make sure things are ok also for him.

> The only further thing coming into my mind concerning specialized actuators would be relative ones of the (rather cheap) piezo-friction type where you give a longer or shorter button push and get some feedback on how far the thing actually went. That would be a relative UI getting a feedback from the device on the current position. Are there needs to implement such a thing?

Do you mean something that has an encoder but you cannot set the actual absolute position neither the precise relative step? If yes, I guess we'll do the UI when we encounter the need for this...

CitizenErase:
Hi I am trying to use the relative ui and I think there's a bug, the DAQ Move won't load.
Here is the error in the log: 
AttributeError: 'DAQ_Move_UI_Relative' object has no attribute 'move_toolbar'. Did you mean: 'add_toolbar'?

I fixed it by replacing self.move_toolbar by self.toolbar in line 38 of control_modules/daq_move_ui/uis/relative.py. Should I do a PR or can you simply fix it directly in the code?

seb5g:
wait, are you using the 5.2 or the dev branch. I modified the usage in between (hence this PR)!

seb5g:
maybe your fix may target the 5.2? I think I'll do a presentation in the pymodaq days on this (execpt if you want to do another one?)

CitizenErase:
> wait, are you using the 5.2 or the dev branch. I modified the usage in between (hence this PR)

I'm using the 5.2.10 version I'm not using the new implementation

> maybe your fix may target the 5.2? I think I'll do a presentation in the pymodaq days on this (execpt if you want to do another one?)

Yes it's an easy fix for 5.2, I'll let you do the presentation :)

seb5g:
ok then go on for the PR on 5.2.x

PR #1127, opened 2026-07-17
Feature/linewidth and white color background

seb5g:
In this PR, the light styles applies also to pyqtgraph background (default is black). It is now set to white in case light style is set, and the foreground colors (labels, ticks...) are set black.

Also, I introduced a configuration entry in data/plotting namely linewidth to plot data with thicker lines. 

DarkStyle+linewidth=2:

image

LightStyle+linewidth=1:

image


seb5g:
maybe one thing is still off (but unrelated to this PR): the pen styles in the 0D viewer lineout are not dashed like the other lineouts!!!

seb5g:
> maybe one thing is still off (but unrelated to this PR): the pen styles in the 0D viewer lineout are not dashed like the other lineouts!!!

last commit solved this last point!

PR #1130, opened 2026-07-20
Feature/generic settings manager and application to ScanManager

seb5g:
This PR does 2 things:

* A generalization based on the State Manager called SettingsManager (base class to reimplement) allowing to select from a ParameterTree settings and drop them into a table and save them. One can then load them and apply them in the order in the table. This could be used for many configuration tasks, and will be used to refactor the StateManager. In order to develop it properly, I included here a first use of it for the ScanManager:
* A ScanManager allowing to select control modules, pick data, select/configure settings => save them and reload/apply them to the DAQ_Scan extension, see video below

This ScanManager entry will be one of the building blocks of the future sequencer.

https://github.com/user-attachments/assets/9819a0ff-188f-433d-81d5-f3d01a28c661


seb5g:
@Ashwolaa this is the new ScanManager (and more) with no issues as was the case initialy. I'm not restoring all the settings anymore but picking only some and applying them in a given order. 

What do you think?

@quantumm @rgeneaux @bernhardlang any comments?

PR #1134, opened 2026-07-26
fix(dashboard): preserve startup visibility

### LINKED ISSUE #1132: [BUG] CusonApp.show_dashboard(False) doesn't close the dashboard window
### Describe the bug

Calling CusonApp.show_dashboard(False) in a derived CustomExt classes __init__ raises the dashboard window in front of the custom extension window. This also happens when the call is issued in DerivedCustomExt.do_things_after_ui_setup().

### Expected Behavior

The call should close the dashboard window. 

### Environment

```markdown
- OS: Debian Linux
- Pymodaq: 5.2
- python env: 3.13
```

### Additional context

Since the show/hide action icon the the 'dashboard' toolbar works, it looks like the method works only once the corresponding Qt event loop has been started. Is there any other way to close the dashboard during initialization (or not to open it at all)? (related to #1131, not yet working work-around by using a custom extension instead of a custom app)

JosephDemarest:
## Summary

- preserve dashboard visibility selected while an experiment and extension are loading
- remove the redundant final `show()` now that `SharedUI` shows the dashboard during creation
- add a regression test covering a dashboard hidden during startup

## Root cause

`CustomExt.show_dashboard(False)` correctly hid the dashboard during extension setup, but `dashboard.main()` later called `win.show()` unconditionally. That reopened and activated the dashboard in front of the extension. The same call worked after startup because there was no later show to override it.

## Verification

- regression test failed before the fix and passed afterward
- focused dashboard and extension tests: 9 passed
- full `packages/pymodaq/tests` suite with Python 3.13.2 and PyQt6: 288 passed
- CI fatal-error flake8 selection: 0 findings
- Pyright on the new regression test: 0 errors

Closes #1132

The redundant call is also present on `dev`; this fix should be forward-ported there after the stable-branch approach is accepted.

### DIFF
--- a/packages/pymodaq/src/pymodaq/dashboard.py
+++ b/packages/pymodaq/src/pymodaq/dashboard.py
@@ -1440,8 +1440,8 @@ def main():
     else:
         win, dashboard = create_load_dashboard()
 
-    win.show()
-
+    # SharedUI shows the dashboard on creation; preserve visibility changes
+    # made while loading.
     # Run application
     sys.exit(app.exec())
 
--- a/packages/pymodaq/tests/dashboard_test.py
+++ b/packages/pymodaq/tests/dashboard_test.py
@@ -1,6 +1,10 @@
+import sys
+
+import pymodaq.dashboard as dashboard_module
+from pymodaq_gui import qt_utils
 from pymodaq.utils.config import get_set_experiment_path
 from pymodaq_utils.config import GlobalConfig
-from pytest import fixture, mark
+from pytest import fixture, raises
 from pymodaq.dashboard import create_load_dashboard
 
 import qt_themes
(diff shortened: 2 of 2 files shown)


packages/pymodaq/src/pymodaq/dashboard.py
```diff
@@ -1440,8 +1440,8 @@ def main():
     else:
         win, dashboard = create_load_dashboard()
 
-    win.show()
-
+    # SharedUI shows the dashboard on creation; preserve visibility changes
+    # made while loading.
```
seb5g, line 1444: good catch

seb5g (approved):

seb5g:
however this PR is not solving issue #1132 as dashboard.py main method is not called when loading directly extensions!

PR #1135, opened 2026-07-28
Bugfix/shared UI visibility

seb5g:
Add specific keywords to create_dashboard and create_extension method to explicitly show or hide their window at startup.

Applied to all builtin extension main methods


seb5g:
will help with #1132

PR #1143, opened 2026-08-03
Feature/patch/saving mechanisms

seb5g:
As one would often reuse H5Saver objects and related opening/closing files and many related stuff in several extensions, I'm refactoring these in both CustomApp and CustomExt given the application.

This has been done for the new ramp extension pymodaq_plugins_ramping and in this PR for the DAQ_Scan (for the moment)


seb5g:
@Ashwolaa this is mainly the refactorisation of your work on swmr and on some use of properties to use easily h5saver. What do you think?

packages/pymodaq/src/pymodaq/extensions/scan/daq_scan.py
```diff
@@ -151,19 +151,16 @@ def __init__(self, dockarea: gutils.DockArea = None, dashboard: DashBoard = None
...
-        self._h5saver = H5Saver()
-        self._h5saver.settings.child('do_save').hide()
-        self._h5saver.settings.child('custom_name').hide()
-        self._h5saver.new_file_sig.connect(self.create_new_file)
-        self._h5saver.file_changed_sig.connect(self._on_file_changed)
-
+        self.h5saver.settings.child('do_save').hide()
+        self.h5saver.settings.child('custom_name').hide()
+        self.h5saver.new_file_sig.connect(self.create_new_file)
+        self.h5saver.file_changed_sig.connect(self._on_file_changed)
```
Ashwolaa: Connections are already done in CustomApp?

packages/pymodaq_data/src/pymodaq_data/h5modules/backends.py
```diff
@@ -1023,12 +1023,45 @@ def get_children(self, where):
...
+        if depth == 0:
+            return
+        else:
+            for gr in self.walk_groups(where):
+                for child in self.get_children(gr).values():
+                    node_path = child.path[len(where.path):]
+                    if node_path[0] == '/':
+                        node_path = node_path[1:]
+                    if depth is not None and node_path.count('/') == depth:
+                        return
```
Ashwolaa: shouldn't it be a continue instead of a return?

packages/pymodaq_gui/src/pymodaq_gui/utils/custom_app.py
```diff
@@ -295,3 +355,179 @@ def connect_things(self):
...
+        self.statusbar.addPermanentWidget(self._swmr_label)
+
+    @property
+    def h5saver(self) -> H5Saver:
+        if self._h5saver is None:
+            self._h5saver = H5Saver()
+            self._h5saver.settings.child('do_save').hide()
+            self._h5saver.settings.child('custom_name').hide()
+            self._h5saver.settings['base_name'] = self._h5_base_group_name
+            self._h5saver.new_file_sig.connect(self.create_new_file)
```
Ashwolaa: not defined here but in CustomExt

packages/pymodaq_utils/src/pymodaq_utils/math_utils.py
```diff
@@ -165,7 +165,14 @@ def find_index(x, threshold: Union[Number, List[Number]]) -> List[tuple]:
         threshold = [threshold]
     out = []
     for value in threshold:
-        ix = int(np.argmin(np.abs(x - value)))
+        if np.any(np.isnan(x)):
+            try:
+                ix = int(np.nanargmin(np.abs(x - value)))
+            except ValueError:
+                x = np.nan_to_num(x, nan=np.max(x))
+                ix = int(np.argmin(np.abs(x - value)))
```
Ashwolaa: if there is only NaN, np.max will still return NaN?

packages/pymodaq_gui/src/pymodaq_gui/utils/custom_app.py
```diff
@@ -295,3 +355,179 @@ def connect_things(self):
...
+            current_file = self._h5saver.settings['current_h5_file']
+            if current_file and Path(current_file).exists():
+                return self._try_open_existing_file(current_file)
+            else:
+                return FileStatus.NO_FILE
+        return FileStatus.REOPENED
+
+    def close_file(self):
+        self._h5saver.close_file()
+
```
Ashwolaa: Add? Like in show_file_content
         if self._h5saver is not None:

packages/pymodaq_data/src/pymodaq_data/h5modules/backends.py
```diff
@@ -1023,12 +1023,45 @@ def get_children(self, where):
...
+        ------
+        Node
+        """
         where = self.get_node(where)  # return a node object in case where is a string
-        yield where
-        for gr in self.walk_groups(where):
-            for child in self.get_children(gr).values():
-                yield child
+        if only_groups and where.attrs['CLASS'] != 'GROUP':
+            pass
```
Ashwolaa: missing a guard is "CLASS" is not found in attrs. Use get?

packages/pymodaq_gui/src/pymodaq_gui/utils/custom_app.py
```diff
@@ -295,3 +355,179 @@ def connect_things(self):
...
+                    self._h5saver.init_file(update_h5=True)
+                    return FileStatus.NEW
+
+    def load_file(self):
+        self.h5saver.load_file(self.h5saver.h5_file_path)
+        self.update_file_status_led()
+
+    def save_file(self):
+        Path(self.h5saver.settings['base_path']).mkdir(exist_ok=True)
+        filename = select_file(self.h5saver.settings['base_path'], save=True, ext='h5')
```
Ashwolaa: Missing parents=True in args?

seb5g:
I had to solve multiple conflicts and merged dev in this PR...

The only real thing to look at in here are:

* pymodaq_gui.managers.runner_thread_manager
* pymodaq_gui.managers.h5manager
* 
* pymodaq_gui.utils.custom_app
* 
* pymodaq_gui.examples.saving_loading

The first 2 are stand alone objects implementing patterns we have everywhere in pymodaq (extending it to multithread management)
CustomApp is now composed of those two to handle its h5 management and multithreading out of the box

the last one is a simplified but full example of a custom app handling data generation, saving and processing all three in different thread with proper management of the h5 file (writing/reading)

I think it could be made into a tutorial @quantumm  :-)

 @Ashwolaa can you look into the h5file stuff? The manager encompass most of the stuff you implemented recently

 @bernhardlang I'm sure some of this example could be reused for your data processor in your experiment.

Out of this PR we will be able to refactorise quite a lot of code!

seb5g:
all tests pass locally but we still have issues on the pymodaq only test on CI...

seb5g:
In the last commits, I refactored the ModuleSaver related to Extensions. At some point I was transferring the responsability to save data to the DAQ_Viewers (where they were supposed to insert in the h5 their current data). This was completely silly as I wanted a async behaviour but then the time it takes to save data was blocking the daq_viewers....

So now the ExtensionsSaver build the h5structure from the ControlModules they have in the Dashboard (adding their metadata) but the data to save are transferred in a real independant thread adding the data into the h5saver, using inner DetectorExtendedSaver, themselves built on top of a DataToExportExtendedSaver. Even if the data is taking a bit of time to be saved, the UI and the acquisition is not frozen (data are just queued in the Qthread eventloop). Before I merge this huge work, I will adapt the saving to some similar extension, making sure they are still operationals!

PR #1152, opened 2026-08-17
Feature/load extensions with command line arguments

seb5g:
Allow to load directly extensions script :

```bash
python -m path/to/extension -x experiment_name -s state name
```

to load dashboard with preconfigured experiment and state!


bernhardlang (approved):
Looks good and clear to me. Tested and working fine on extension tutorial (which brought this up).

seb5g:
do you think this should be ported to 5.2.x here is only dev?

PR #1156, opened 2026-08-18
Feature/experiment manager state machine

seb5g:
Using StateMachine to remove the need of polling and using ProcessEvents while loading modules in the dashboard.

The implementation use a statemachine and signals to make sure the modules are ready for creation, setting their type, adding them to docks and initializing them!


seb5g:
I hope this will remove some random errors that happened when loading experiments!

packages/pymodaq/src/pymodaq/utils/managers/experiment/experiment_manager.py
```diff
@@ -1,11 +1,27 @@
...
+
+try:
+    state_machine_available = True
+    if qtpy.PYQT6:
+        from PyQt6.QtStateMachine import QStateMachine, QState, QFinalState, QSignalTransition
+    elif qtpy.PYSIDE6:
+        from qtpy.QtStateMachine import QStateMachine, QState, QFinalState, QSignalTransition
+    elif qtpy.PYQT5:
+        from PyQt5.QtCore.QtStateMachine import QStateMachine, QState, QFinalState, QSignalTransition
+except ImportError:
```
Ashwolaa: I think that should be 
        from PyQt5.QtCore import QStateMachine, QState, QFinalState, QSignalTransition

packages/pymodaq/src/pymodaq/dashboard.py
```diff
@@ -369,6 +369,7 @@ def _remove_module_list(self, modules, module_list, compact_manager_attr,
                         setattr(self, compact_manager_attr, None)
                 module.quit_fun()
                 dock = self.dockarea.docks.get(module.title, None)
+                self.docks_viewer.remove(dock)
                 if dock:
```
Ashwolaa: should be after the if condition since dock can be None from above

seb5g:
the loading seems way more robust. But looking at the failing tests, I can see there are still quite a bit of seg faults. I believe, the quit_fun mechanism and propagation is ohayish but because it doesn't wait for modules to be quitted properly it creates all those errors. To go furhter, one will have to include such state machine to all async processes (init as I did here but also quitting).

a dashboard has many "children": control modules and extensions. All should have a Composite state handling their own initialisation and quitting. When the dashboard quits, it would run all thoses quitting states in parrallel and transition to its own FinalState when all parallel children are done!

Quite a piece of work!

seb5g:
@Ashwolaa could you check this PR is ok for merging?

seb5g:
@Ashwolaa I have a feeling that your test on multiple execution of the dashboard experiments with some failling is inducing a lot of problems

Ashwolaa:
Could be, the main idea was to check that multiple execution would not lead to weird behavior (which I think they still do though). Most of them were initially done to be able to relaunch a preset.


I have launched some agents to check your changes. At a first sight, the QStateMachinery looks a bit overkilled here. I like the PluginInfo

seb5g:
Yeah could have done it with a few methods and signals only but the states also helps in understanding the flow!

seb5g:
@Ashwolaa I'm done here with the new implementation. Starting from Claude PR I added some timeout handling

I also cleaned this loader to be agnostic of who is using it. It only need an instance of DashBoad. Like this I could use it to simplify also the add_det_frm_extension and add_move_from_extension

I have still many failures on CI but None locally!

Could you check this code?

seb5g:
I'm merging this as I need it to finish the one instrument / one thread PR especially the possibility to connect to masters after the experiment has been loaded!

PR #1157, opened 2026-08-19
Factorizing/group_parameter

Ashwolaa:
- Pyqtgraph group parameter has been updated to take into account the  AddMenu option making the current version obsolete
- The GroupParameterPatch done in move_utility_class.py was affecting the whole group parameter. Instead, it seems better suited to just patch these exceptions in experiment_manager.py


seb5g (approved):

seb5g:
@Ashwolaa there is an issue in with this PR. 

The thing is that some plugins were written to call explicitly self.settings['multiaxes', 'axis'] even though their "common settings" have been changed by the controller/axis ... layout. The patch was working while trying to access those values, it diverted the reading to the new/right position.

The new patch in ExperimentManager only (actually removed in the new implementation of the manager) is useless as all the Actuators seems  from external use to have the new layout (because we are forcing them) but the actual inner code may call the old style (as below BSPiezoMirror in the mockexamples plugin) and now it's raising an error. I can patch the plugin of course but who knows how many of this old style code still exists!!!!

```
position = self.set_position_with_scaling(position)
axis = self.settings['multiaxes', 'axis']
pos = self.controller.move_at(position, axis)
```

Could you think of a way to make it still work while initializing and using such an old style module (BSPiezoMirror for instance)?

seb5g:
@Ashwolaa we have another hotbug here, the AddMenu has been included in the dev branch of pyqtgraph but not yet released, people not using the dev branch will not be able to use the AddMenu feature (as I did this morning...). The pymodaq group module removal will have to be postponed for after the release of pyqtgraph 0.14.1 ....

seb5g:
I'm restoring the groupe module and overriding group type in #1195

Ashwolaa:
arf... I got biased only working from the master branch of pyqtgraph. I asked them when they plan to get it on pip

PR #1161, opened 2026-08-19
Feature/Using_module_loader_class_in_experiment_manager

Ashwolaa:
Claude's proposition to simplify the loading and execution of the experiment manager without going through QState machinery.


seb5g (approved):
missing 

self.manager.close_subentries_display()

in the _on_load_failed!

PR #1164, opened 2026-08-21
Bugfix/qthread qtimer teardown leak

Ashwolaa:
Here is the tentative to get rid of errors in pytest and making sure that connections using self are properly garbage collected once the instance gets closed.
`DashBoard` instances (and everything reachable from them — modules,
hardware `QThread`s, extensions) were never garbage collected once created,
because several Qt signal connections closed over `self` (bound methods or
`self`-capturing lambdas). PySide/PyQt keep an internal reference to a
connected Python callable that outlives `disconnect()` and can survive even
after the underlying `QObject` is destroyed, so any such connection kept the
whole object graph alive indefinitely — including live hardware `QThread`s.
Left running, this surfaced as a `SIGABRT` at full test-suite exit (Qt
aborting on threads that were never `.wait()`-ed on because nothing had
released them).

This PR centralizes on a single pattern, `weak_slot()`, for every connection
that needs to close over `self`, removes teardown scaffolding that becomes
redundant once that pattern is applied consistently.

### DIFF
--- a/packages/pymodaq_gui/src/pymodaq_gui/managers/parameter_manager.py
+++ b/packages/pymodaq_gui/src/pymodaq_gui/managers/parameter_manager.py
@@ -282,39 +282,6 @@ class ParameterManager:
         )
         self._settings_tree.tree.resizeColumnToContents(0)
 
-    def clear_settings_tree(self):
-        """Disconnect and destroy the settings tree widget, dropping references to it.
-
-        :meth:`set_settings` connects lambdas (closing over ``self``) to the tree's
-        ``itemExpanded``/``itemCollapsed`` signals, and ``self._settings_tree`` is
-        itself an :class:`~pymodaq_gui.managers.action_manager.ActionManager` with its
-        own toolbar actions connected to bound methods of ``self``. As with
-        :meth:`~pymodaq_gui.managers.action_manager.ActionManager.clear_actions`,
-        disconnecting alone is not enough to release those closures: PySide/PyQt keep
-        an internal reference to a connected Python callable that outlives
-        ``disconnect()`` until the underlying C++ widget is actually destroyed.
-        """
-        try:
-            self._settings_tree.tree.itemExpanded.disconnect()
-            self._settings_tree.tree.itemCollapsed.disconnect()
-        except (TypeError, RuntimeError):
-            pass
-        if isinstance(self._settings_tree, ActionManager):
-            self._settings_tree.clear_actions()
-        # deleteLater() the tree widget itself, not just its container: a parent's
-        # deleteLater() only *schedules* its own destruction, and does not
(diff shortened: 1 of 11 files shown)


packages/pymodaq_utils/src/pymodaq_utils/weak.py
```diff
@@ -0,0 +1,128 @@
...
+        # directly: identical connections deliver the full signal argument in
+        # one case and none of it in another, depending only on whether a
+        # pytest-qt `qtbot` fixture is active in the process -- not on
+        # anything specific to the target method). Rather than depend on
+        # guessing that resolution, generate a wrapper with exactly
+        # `max_extra_args` named positional parameters, so its arity is
+        # concrete and unambiguous to any introspector.
+        param_names = [f'_a{i}' for i in range(max_extra_args)]
+        extra_call_args = ''.join(f', {name}' for name in param_names)
+        source = (
```
seb5g, line 105: why not declaring a plain def internal function?

packages/pymodaq_gui/src/pymodaq_gui/managers/parameter_manager.py
```diff
@@ -319,8 +319,15 @@ def set_settings(self, settings: Union[Parameter, List[Dict[str, str]], Path]):
             self._settings, showTop=False,
         )  # load the tree with this parameter object
         self._settings.sigTreeStateChanged.connect(self.parameter_tree_changed)
-        self._settings_tree.tree.itemExpanded.connect(lambda: self._settings_tree.tree.resizeColumnToContents(0))
-        self._settings_tree.tree.itemCollapsed.connect(lambda: self._settings_tree.tree.resizeColumnToContents(0))
+        # Close over the tree widget directly, not `self`: every ParameterManager
+        # subclass (DashBoard, ModulesManager, per-module settings, ...) goes through
+        # here, and a `self`-closing lambda connected to a Qt signal is not reliably
+        # released by disconnect() alone (PySide/PyQt keep an internal reference to a
+        # connected Python callable that outlives disconnect()), which would keep
```
seb5g, line 326: do you have the source (discussion/doc) of this ?

packages/pymodaq_utils/src/pymodaq_utils/weak.py
seb5g, line 1: should be in gui as related to qt stuff!

seb5g (commented):
Could you benchmark the effectiveness of this? 

I commented/asked a few questions in the changed files.

Also, why marking it as draft? I guess there are lots of other place where we have such connection?

Ashwolaa:
Let me hold on a bit on this PR. I am not so pleased about some of these solutions yet. 

I'll go back to review whence I think it is ready.

PR #1166, opened 2026-08-21
Feature/improving control dock

Ashwolaa:
Adding close + detach/attach buttons to widget_with_label 

This allows to close the settings directly from the dock and to detach it 

https://github.com/user-attachments/assets/68ca9448-34d7-4a71-8a5f-e92ebb88f3b9


seb5g:
@Ashwolaa there are some conflicts that probably arose due to recent merged PR . Could you have a look and let me know? Also, I already implemented a close button in one past PR, don't know why I didn't show it to you the other day, but this is it:
image

Did you build your PR on that?

Ashwolaa:
Nope, I did not see it. There were not appearing on the dev when I made it.
Do you prefer to stick to the red button for the close action?

seb5g:
I think it is better indeed to keep the same button for the same kind of action. Your layout on the right is nice. You did it only for the settings, I could not see it for the move controls and the ROIs?

seb5g:
@Ashwolaa could you build on what you did here to extend it to ROI dock and Actuator Controls:
image

Also, the ROI button in a Viewer1D is supposed to open its ROI manager within that side ROI dock, which is not the case in your PR??

Ashwolaa:
This should be now implemented for ROIs and Actuators

seb5g:
perfect! merging it!

PR #1169, opened 2026-08-21
Feature/viewer0 d history upgrade

Ashwolaa:
Viewer0D: fix buffer/xaxis desync on history resize, add reserved "future space" toggle

Description:

- Fix a crash (X and Y arrays must be the same shape) that could occur when changing the history length (Nhistory) mid-acquisition — the internal sample counter could grow unbounded past the buffer's actual trimmed size, and re-raising Nhistory above that stale counter desynced the x-axis from the data arrays.
- Add a new toggle (slideshow icon) + Nfuture % spinbox: when enabled, reserves a configurable fraction of the history span as blank space ahead of the last data point, instead of the plot always autoranging tight to the data — useful when optimizing a signal and you want a bit of headroom on the right.
- Nhistory spinbox now only accepts integers.
- Toolbar cleanup: smaller spinboxes, separators grouping related actions.

https://github.com/user-attachments/assets/57767706-da82-418d-9ff6-10cfd1d66310


seb5g (commented):
I don't get the why of this "future" space. How could it helps with optimizing?

Ashwolaa:
> I don't get the why of this "future" space. How could it helps with optimizing?

When you touch your actuators (optics/actuators), it can be difficult to see when your last points are always on the far border. So you typically make a larger range and try to optimize it before the curve touch the right edges. With this feature, you do not need to do this kind of tweaking.

seb5g:
Why not using the min and max bars that are meant for this?

Issue #1173, opened 2026-08-23
[BUG] MultiAxis settings reading in old plugin code not working


seb5g:
### Describe the bug

PR #1157 introduces a simplification on the Group Parameter redefinition made to route some settings reading to have backcompatibility.

The thing is that some plugins were written to call explicitly self.settings['multiaxes', 'axis'] even though their "common settings" have been changed by the controller/axis ... layout. The patch (before #1157) was working while trying to access those values, it routed the reading to the new/right path.

The new patch in ExperimentManager only (actually removed in the new implementation of the manager, PR #1156 ) is useless as all the Actuators seems, from external use, to have the new layout (because we are forcing them) but the actual inner code may call the old style (as below BSPiezoMirror in the mockexamples plugin) and now it's raising an error. I can patch the plugin of course but who knows how many of this old style code still exists!!!!

```
position = self.set_position_with_scaling(position)
axis = self.settings['multiaxes', 'axis']
pos = self.controller.move_at(position, axis)
```

### Reference(s)

```markdown
pymodaq.utils.config.deep_update
```

### Environment

```markdown
- OS:
- Pymodaq:
- python env:
```

seb5g:
fixed by #1174

PR #1174, opened 2026-08-24
hotfix/legacy_mutiaxes_access

Ashwolaa:
Fixing #1173 
- Using the old patching (with a custom GroupParameter) for DAQ_move with a custom type 
- Adding a deprecation warning for old style plugins


seb5g (approved):
looks good!

PR #1175, opened 2026-08-24
Feature/slave connect

seb5g:
Following PR #1110 and #1156  this new PR is:

* setting the info on the master controller and thread to any slave related to id by the same ID
* pulling from the already initialized master, the one with the same ID and getting its controller and thread

Overall this means you can now init/desinit any instrument. As long as a Master is reinit, any slave sharing the same controller id will be able to get it from it and initialize.


seb5g:
before reviewing, I want still add a colouring scheme for Master/Slave init/desinit state (probably coloring the title) like this:

* green: master is init
* red: all desinit
* purple: slave is init

@Ashwolaa do you have some other suggestions along the line?

also to do: allow to add directly from the DashBoard any given instrument (adding it to the experiment manager (or not)). Could be a menu button attached to the ExperimentManager toolbar and allowing to select any given detector/actuator/axis/slave_of!

seb5g:
@Ashwolaa YOU HAVE TO CHECK THIS BELOW, you'll be pleased!!!!!


https://github.com/user-attachments/assets/9e289bf5-5b8a-49d6-9651-99fd5b81443b

seb5g:
Included in last commits is to regroup in a ModuleCreator the action of creating the Control modules from the definition of PluginInfo (from ExperimentManager or elsewhere like extensions) and adding them into the dashboard. It is also responsible for looking for available masters for a slave to init and finally exposes a menu button where one can select a ControlModule to create/add with easy shortcut to select axis, master/slave and if slave of which already opended master.

A minus action is added in each module toolbar to remove it from the DashBoard!

seb5g:
I checked it is functionnal for extension wanting to create their modules even if a bit more complex compared to before. Done it for PID. 

To do:

* check for DataMixer
* check for BeamShaping

* but there should still be some backcompatibility!

seb5g:
all done for me! This PR is ready for review (big chunk...)

packages/pymodaq/src/pymodaq/utils/managers/state/subentries.py
```diff
@@ -358,10 +377,11 @@ def get_subentry_from_dialog(self) -> StateSubEntry:
     def execute_subentry(self, entry: StateSubEntry,
                          dashboard: 'DashBoard'):
         """ Execute the given subentry """
-        start = time.perf_counter()
+        QtCore.QTimer.singleShot(int(entry.setting.value()),
+                                 self._on_wait_time_done)
```
Ashwolaa, line 252: there is no self here I guess?

packages/pymodaq/src/pymodaq/control_modules/move_utility_classes.py
```diff
@@ -277,6 +277,15 @@ class level attribute. Defines if the plugin controller controls multiple axes.
 
     data_shape = (1,)  # expected shape of the underlying actuator's value (in general a float so shape = (1, ))
 
+    @classmethod
+    def get_class_axis(cls) -> list[str] | None:
+        if cls.stage_names is not None and len(cls.stage_names) != []:
+            #check for old and deprecated plugins
+            axis = cls.stage_names
```
Ashwolaa: Need to add warning here at least. It is ok to keep some deprecated syntaxes but this should not be permanent

packages/pymodaq/src/pymodaq/control_modules/move_utility_classes.py
```diff
@@ -277,6 +277,15 @@ class level attribute. Defines if the plugin controller controls multiple axes.
...
+        if cls._axis_names is not None:
+            axis = cls._axis_names
```
Ashwolaa: shouldn't it be elif then? Will it not overwrite the previous one?

packages/pymodaq/src/pymodaq/control_modules/move_utility_classes.py
```diff
@@ -277,6 +277,15 @@ class level attribute. Defines if the plugin controller controls multiple axes.
...
+        return axis
+
```
Ashwolaa: in some cases axis is never defined, should be None as default?

packages/pymodaq/src/pymodaq/utils/managers/state/state_manager.py
```diff
@@ -136,33 +136,51 @@ def _execute_entry(self, entry_path: Path = None, **kwargs) -> bool:
...
+            self.subentry_handler.executed_signal.connect(self._on_executed)
+            self.subentry_handler.execution_failed.connect(self._on_execution_failed)
+            self.subentry_handler.execute_subentry(entry, dashboard=self.dashboard)
+        except SubEntryError as e:
+            logger.exception(str(e))
+            self.subentries_model.set_status(self._ind_subentry, False)
+            self._advance()
+
+    def _on_execution_failed(self):
+        self.subentries_model.set_status(self._ind_subentry, False)
```
Ashwolaa: should accept an error in the arguments to log the error?

packages/pymodaq/src/pymodaq/utils/managers/modules/module_creator.py
```diff
@@ -0,0 +1,474 @@
...
+    app = mkQApp('Selector')
+
+
+    def print_path(path: tuple[str]):
+        print(path)
+
+
+    button = build_menu_for_module_creation(masters=['act0', 'act1', 'act2'], )
+    button.triggered.connect(print_path)
+    button.show()
```
Ashwolaa: No longer relevant

packages/pymodaq/src/pymodaq/control_modules/utils.py
```diff
@@ -122,12 +124,31 @@ def queue_command(self, command) -> bool:
...
+        self.thread = thread if thread is not None else QThread()
 
         self._hardwares = {}
 
+    def __getattr__(self, item):
+        if hasattr(self.thread, item):
+            return getattr(self.thread, item)
+        else:
+            raise AttributeError(f'{item} is not a known attribute of Qthread')
+
```
Ashwolaa: From Claude
 ```
 If self.thread is not yet set (e.g., during __init__ before self.thread = ...), accessing self.thread in __getattr__ triggers __getattr__ again, causing infinite recursion. In Python, __init__ happens after __new__, and super().__init__() (QObject init) is called before
  self.thread = ..., so any attribute access on self during QObject init that fails normal lookup will recurse. This is fragile. Consider using self.__dict__.get('thread') inside __getattr__ as a guard.
  ```

packages/pymodaq/src/pymodaq/utils/managers/modules/module_creator.py
```diff
@@ -0,0 +1,474 @@
...
+            pass
+
+    def create_actuator(self, name: str, class_name: str, ui_identifier: str) -> DAQ_Move:
+        actuator_class = find_actuator_class_from_name(class_name)
+        forced_ui = actuator_class.ui_type
+        ui_identifier = forced_ui if forced_ui != UiType.NONE else ui_identifier
+
+        if ui_identifier is not None:
+            pass
+        else:
```
Ashwolaa:
```
if ui_identifier is None:
    do_stuff
```

packages/pymodaq/src/pymodaq/utils/managers/state/subentries.py
seb5g: This is a standalone convenience method, so no self needed (as for the doc)

packages/pymodaq/src/pymodaq/control_modules/move_utility_classes.py
seb5g: ok added deprecation message

packages/pymodaq/src/pymodaq/control_modules/move_utility_classes.py
seb5g: yes it will overwrite indeed but if for some reason the developer included both or left old stuff in new implementation, the _axis_names bit has priority

seb5g: added comment in code

packages/pymodaq/src/pymodaq/control_modules/move_utility_classes.py
seb5g: it is defined as None in this base class! however I added that is will be set to [''] in case it is defined as [] (this is what happens in the common_parameter_fun method). This is why I'm going to use this class method in there as well  to avoid multiple handling and defaulting values

seb5g: I renamed it to get_class_axis_names to be more explicit

packages/pymodaq/src/pymodaq/utils/managers/state/state_manager.py
seb5g: indeed, I added it as a warning for the log

packages/pymodaq/src/pymodaq/utils/managers/modules/module_creator.py
seb5g: yes removed this if clause

packages/pymodaq/src/pymodaq/control_modules/utils.py
seb5g: I reworked the Proxy with much care to handle various issues (as the one pointed out by Clause) but also for IDE like pycharm or VS propose the right autocompletion and doc

packages/pymodaq/src/pymodaq/utils/managers/modules/module_creator.py
seb5g: done

seb5g:
implemented or reworked your suggestions!

Ashwolaa:
Looks ok on usage but there are new errors now in the tests that are worth fixing

seb5g:
Yes I'm on it !!!
Do you agree it is ok to not be in separated thread anymore? Usage on the
UI is smooth while scan is running with the help of the new workflow in
DAQScanAcquisition
Sébastien Weber

*Référent Organisation AbsoluRaid*
***@***.*** ***@***.***>*
*https://absoluraid.fr/ <https://absoluraid.fr/>*

PR #1177, opened 2026-08-28
Feature/async move grab

seb5g:
This PR builds on the experience gained on PR #1156 on sequencing action triggered with signals, I rewrote the daq scan acquisition step using only such transitions between clearly defined steps during the scan.

The first work has been to add methods to async move and grab (in modules manager) using only the emission of a generic signal to assert if move or grab is done.

Then I split the previous long and big while loop in the scan in basic steps and connected them together with signals/callback.

Interruption and debugging is much simpler like this. 

This is a first step into being able to include interuptions like pausing but dedicated to complex task like for instance optimizing a signal during the scan!


seb5g:
@bernhardlang this is the direct impact of one of our discussions!

@Ashwolaa you should be familiar with the new pattern and syntax as I reused the "_on_thing_happening" method naming which I find quite explicit!

seb5g:
tests are not passing due to two tests related to timeouts. These tests look at a deleted timeout_flag. I'm not sure how to modify them. @Ashwolaa you created them, do you see somehting simple to be tested here?

seb5g:
for those tests I reintroduced the timeout scan flag to assert the tiemout has been set

Ashwolaa:
They were mostly introduce to test the stop_on_timeout feature and gives an attribute to indicate if the scan got stopped from it.
Not mandatory overall but it is just a bool so I would leave it

packages/pymodaq/src/pymodaq/control_modules/move_utility_classes.py
```diff
@@ -102,28 +101,25 @@ def comon_parameters(epsilon=config('pymodaq', 'actuator', 'epsilon_default'),
...
         the value the move should reach
 
     """
 
-    def __init__(self, move_type, value=0):
-        if move_type not in MOVE_COMMANDS:
-            raise ValueError(f'The allowed move types fro an actuator are {MOVE_COMMANDS}')
+    def __init__(self, move_type: MoveType | str, value=0):
+        if move_type not in MoveType.names():
+            raise ValueError(f'The allowed move types for an actuator are {MoveType.names()}')
```
Ashwolaa: This command gives uppercase instead of lowercase.
MoveType.value()

packages/pymodaq/src/pymodaq/extensions/scan/daq_scan.py
```diff
@@ -1308,144 +1304,205 @@ def queue_command(self, command: utils.ThreadCommand):
...
-                        QThread.msleep(50)
+        self._update_status(message)
+        logger.info(message)
 
-                    #move motors of modules and wait for move completion
-                    positions = self.modules_manager.order_positions(self.modules_manager.move_actuators(positions))
+    def advance(self):
+        try:
+            if self.stop_scan_flag:
+                self.finalize_scan()
```
Ashwolaa: missing a return

packages/pymodaq/src/pymodaq/extensions/scan/daq_scan.py
```diff
@@ -1308,144 +1293,209 @@ def queue_command(self, command: utils.ThreadCommand):
...
     def set_ini_positions(self):
-        """ Set the actuators's positions totheir initial value as defined in the scanner  """
+        """ Set the actuators's positions to their initial value as defined in the scanner  """
+        self.modules_manager.move_actuators_with_callback(
+            self.scanner.positions_at(0),
+            mode=MoveType.ABS,
+            callback=self._on_ini_positions)
+
+    def _on_ini_positions(self, dte: DataToExport):
+        self._update_status("Initial values of actuators reached!")
```
Ashwolaa: Could add a condition if stop_scan is clicked while positions are being init.

        if self.stop_scan_flag:
            self.start_scan_flag = False
            self.stop_scan_flag = False
            self._update_status("Acquisition has been stopped")
            self.status_sig.emit(utils.ThreadCommand("Scan_done"))
            return

packages/pymodaq/src/pymodaq/utils/managers/modules/modules_manager.py
```diff
@@ -111,6 +122,10 @@ def actuator_timeout(self):
     def detector_timeout(self):
         return config('pymodaq', 'viewer', 'timeout')
 
+    def enable_modules(self, enable=True):
+        for module in self.modules_all:
+            module.ui.toolbar.setEnabled(enable)
```
Ashwolaa: add a condition to check that module.ui is not None

packages/pymodaq/src/pymodaq/extensions/scan/daq_scan.py
```diff
@@ -1308,144 +1293,209 @@ def queue_command(self, command: utils.ThreadCommand):
...
-            self.modules_manager.connect_detectors(False)
+            self.status_sig.emit(
+                utils.ThreadCommand("Update_scan_index",
+                                    attribute=[self._ind_scan, self._ind_average]))
+            return self.scanner.positions_at(self._ind_scan)  # get positions
+        except Exception as e:
+            self.scan_step_failed_signal.emit(ScanStepError(f"Error when getting next step position:\n"
+                                                            f"ind_step: {self._ind_scan}:\n"
+                                                            f"ind_average: {self._ind_average}:\n"
+                                                            f"{str(e)}"))
```
Ashwolaa: no need for the exception here I think, you will already collect it in advance and you are then returning None which can break up the move_actuator_callback I thnk

PR #1179, opened 2026-09-01
Feature/sequential parallel state execution

seb5g:
In this PR, I:

* applied the generic Settings Manager to State Manager
* added in the manager toolbar an action to tell wether to aply the settings sequentially (default) or all at the same time (in parallel)

This action's checked state is saved in the state file and therefore is persistent. 

The user should be cautious in using parallel states in particular when dealing with initialization. But for moving things around and saving time it's perfect!

parallel state still waits for all the state to be applied before moving on!


seb5g:
https://github.com/user-attachments/assets/e07a3c4f-bc1b-44ba-bc8c-d082e30d40e2

One small bit that is not that nice in term of UX => if you check/uncheck the sequential/parallel action, you'll have to save again the file (as the file is reloaded before execution...)
I'm not sure what's the best approach:

* save before executing 
* apply what's on screen that may differ from the locally saved file
* stay like this
* 
@Ashwolaa , @bernhardlang @runrunm any thought on that aspect?

Would this behaviour (to be discussed) shoudl be applied to experiment manager, scan manager and here state manager or could it be different on different managers?

seb5g:
I merged the PR but still want your opinion on the raised questions above!

Ashwolaa:
Hum not sure, I would usually make a pop up when executing/quitting checking the current status vs the saved one.
If they differ, ask the user:
The file has been changed: do you want to save it?

seb5g:
Do you agree that when you execute, what is executed should be what is shown on screen (not what is saved on a file somewhere)? Maybe adding a visual indicator that the content of the current file differ from the actual state of the manager (like in a text editor when the file is not saved it appears with a star or in italic...)?

Ashwolaa:
> Do you agree that when you execute, what is executed should be what is shown on screen (not what is saved on a file somewhere)? Maybe adding a visual indicator that the content of the current file differ from the actual state of the manager (like in a text editor when the file is not saved it appears with a star or in italic...)?

Yes, I agree. 
Agree for the visual indicator:  Star / color / font . Not sure what is the best (maybe all of them ^^)

Issue #1022, opened 2026-04-15
PyMoDAQ's color scheme


Ashwolaa:
From the initial #1002 and the following #1016
I was thinking about getting feedback from an action between the user and the GUI. This idea would be to extend the QLED to have a multistate.

In this regard, I think it could be useful if we agree on a color scheme that could be apply throughout pymodaq. This is nothing major but I guess it enters into the current questions that we have.

Some of it could be related to some particular logger state. For example, warning, error, critical could be yellow / orange / red ?

Here is an example for it. In my opinion, the qt_themes are not so good with all of them to properly differentiate yellow from orange. We could maybe ask the maintainer about it?

https://github.com/user-attachments/assets/a5336adc-52ce-4baa-a70a-d351edf5e636

If you want to try it on your computer, the branch is:
https://github.com/Ashwolaa/PyMoDAQ/tree/feature/multistate_LED
The example is \packages\pymodaq_gui\src\pymodaq_gui\examples\status_palette_proposal.py

seb5g:
We don't have to stick to the qtthemes colors if we are not happy with it especially with these leds so you can pick a set of colors specific for the leds that you would like and share it with us

PR #1207, opened 2026-09-20
Feature/multistate led

Ashwolaa:
Following #1022
Here is a PR to expand the current binary LED system to a generic multi-state LED widget (MultistateLED) wherever a control module or extension needs to show more than just  on/off status, and applies it across the codebase's status indicators.
It essentially contains:
- New widget (pymodaq_gui/utils/widgets/multistate_led.py): a painted LED that cycles through an arbitrary list of named (state, color) pairs, with configurable shape (circle/square/triangle/diamond/rectangle) and gradient style, built on a new shared painter_utils module.
  - StatusPalette (pymodaq_gui/utils/status_palette.py): a small set of semantic colors (off, idle, running, critical, …) so state colors stay consistent across the app instead of being hand-picked per widget. 
  - Backward compatibility: the led and action_led parameter types (pymodaq_ptypes) are now backed by MultistateLED internally but keep their existing bool-valued value()/setValue() API.

It is applied to real indicators, replacing single-color QLEDs with multi-state ones:
    - DAQ_Move UI: actuator status LED now shows uninitialized / idle / moving / error (merging what used to be two separate LEDs — init state and move-done — into
      one).
    - DAQ_Logger: logging LED shows idle / running / error.
    - DAQ_Scan: scan LED shows idle / running / complete / error.
    - DAQ_PID: model/runner LEDs show uninitialized / ready / error; run LED shows idle / running / error.
    - Optimizer extension: same idle/running/complete/error pattern as DAQ_Scan.                                                                                    
  - Test coverage added for the new painter utilities and the led parameter type.


seb5g (changes requested):
I like it a lot, thx!

Only thing, I would like you to create a StrEnum for the various possible states and use this enum in hte code to avoid using hardcoded strings!

seb5g (approved):

PR #1210, opened 2026-09-21
I fixed the non passing tests, especially the ioxml serialization

seb5g:
@Ashwolaa  I'm wondering why you used list of tuple for the states?
Also why not using the standard 'limits' keyword that can already be a dict for the list parameter?

I'm merging this as I want it to pass the test but maybe another go at cleaning this should be done. Also the potential 'false' != 'False' was creating a mess, I turned the strings as 'False' and 'True'


Ashwolaa:
I'll take a deeper look, I give it a quick update this morning and did not review it properly, especially the serialization

seb5g:
yeah <I was too fast in merging it also... Still on it fixing saturation stuff. Then you'll be able to rework it a bit!

